PR #22, opened 2012-11-16
Updated French translations

regisd:
Updated translations (but it doesn't cover 100%)


derekgengenbacher-wf:

PR #23, opened 2012-11-24
Google analytics ARELLE-16

regisd:
Hello Herm!

I am pleased to offer you this pull request to fix ARELLE-16.
- The Google analytics integration is done with a plugin.
- It works
- The plugin generates a random client identifier, that it stored in Arelle preferences, so that the "Unique number of users" is meaningful.
- The only impact on the core code is a new extension point in `Cntrl`, called "Cntrl.init"
- The plugin tracks all `Dialog*` classes as "screen view". This is displayed live in Google Analytics "real time analytics"
- The plugins tracks a couple of selected methods as "events". This is displayed in Google analytics with a 24h delay
- I still need to think how to track more methods:
  - I was originally thinking about introspection
  - I have just discovered you have a method called `ModelXbrl.profileStat`. But I have not understood how t works
  - Also I think such a tracking method should be a decorator, rather than modyfing so deeply the profiled method calls
- Is there anything else you'd like to track? In particular, crash reports can be tracked (but I don't see how not to modify the core code for this)
- The Google measurement protocol is just a couple of days old. I had to write my own python library
- This `google_measurement.py` library defines all usefull methods for App tracking. 
- All http calls in `google_measurement.py` are asynchronous


derekgengenbacher-wf:

PR #24, opened 2012-11-24
Gui package

regisd:
I propose to move all modules related with GUI to a dedicated python package.

I think it makes the code easier to browse.
Another benefit is that improper imports and code dependencies are easier to spot


derekgengenbacher-wf:

PR #25, opened 2012-11-24
Qnames with labels 103

regisd:
A plugin that displays the labels for QNames, as requested in https://arelle.atlassian.net/browse/ARELLE-103


derekgengenbacher-wf:

PR #26, opened 2012-12-20
Save DTS should Handle nested directories

regisd:
I thought this was done already, but v0.9 of save DTS doesn't handle nested directories.

This pull reqeust comes with:
- a ZIP that keeps the hierarchy, based on a root directory
- a method to guess the root  directory of a DTS (it is not necessary the directory of the entry point)
- a CLI option to manually specify the root directory


derekgengenbacher-wf:

PR #35, opened 2014-12-02
EBA filing indicators

gmongelli:
Please only apply this pull request after having handled https://github.com/Arelle/Arelle/pull/34 (Keep table changes when navigating from one table to another). This pull request also contains all the changes from that pull request as it cannot work reliably without it.

This pull request adds a new plug-in to Arelle. It improves the EBA/IOPA compliance of the XBRL instances generated by Arelle.

On request, filing-indicator tuples are added to the instances. Any previous filing-indicators are removed and new ones are generated. This change is not 100% compliant with the EBA because the EBA mandates the usage of some filing-indicators depending on the filer type. Those rules have not yet been implemented.

However, the current filing-indicators are usable for some national supervisory authorities for the case where they do not need to transmit the instances to the EBA (currently for investment firms).

### DIFF
--- a/arelle/ModelFormulaObject.py
+++ b/arelle/ModelFormulaObject.py
@@ -1119,7 +1119,7 @@ class ModelConceptName(ModelFilter):
     
     def filter(self, xpCtx, varBinding, facts, cmplmt):
         if not self.qnameExpressionProgs: # optimize if simple
-            qnamedFacts = set.union(*[inst.factsByQname(qn)
+            qnamedFacts = set.union(*[inst.factsByQname(qn, set())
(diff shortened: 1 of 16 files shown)


gmongelli:
This pull-request is still very useful, but most probably it will be the cause of many conflicts (for instance regarding the EBA validation).

gmongelli:
@hefischer :
The work is mostly hard because ModelXbrl has changed a lot. Could you please review these changes? If you find them useful, I shall prepare a pull request that will be easier to merge with the current master branch.

derekgengenbacher-wf:

PR #36, opened 2014-12-02
Drop-down lists for EBA facts

gmongelli:
Please only apply this pull request after having handled #35 (EBA filing indicators). This pull request also contains all the changes from that pull request as it cannot work reliably without it.

This pull request introduces the usage of drop-down lists in Arelle.

The drop-down lists are generated thanks to a specific Eurofiling attribute ('model') that is found in EBA taxonomies.

There is also support for enumerations, but it has not been tested.


gmongelli:
This pull-request is now completely outdated compared to #47 .
A new one must be created.

PR #37, opened 2014-12-02
DECIMAL attribute handling

gmongelli:
Please only apply this pull request after having handled #36 (Drop-down lists for EBA facts). This pull request also contains all the changes from that pull request as it cannot work reliably without it.

If the default decimals value supplied by the Arelle GUI is `INF`, it is used unchanged.
If the default value is not `INF`, the decimals are computed according to the supplied value for the numeric fact:
* 12345 => decimals = 0
* 12120 => decimals = -1
* 12000 => decimals = -3
* 12000.12 => decimals = 2
* 12000. => decimals = 0
* 12000.00 => decimals = 2

Moreover, the EBA filing rules (see section "2.18 — @decimals" in version 2.0.0) require a minimal number of decimals:
* Monetary facts: decimals >= -3
* Percent facts: decimals >= 4
* Integer facts: decimals = 0

The above mentioned rules are enforced in this implementation of the EBA compliance module.

In this implementation percent facts are defined as those facts having the type `{http://www.xbrl.org/dtr/type/numeric}num:percentItemType`.

### DIFF
--- a/arelle/ModelFormulaObject.py
+++ b/arelle/ModelFormulaObject.py
@@ -1119,7 +1119,7 @@ class ModelConceptName(ModelFilter):
     
     def filter(self, xpCtx, varBinding, facts, cmplmt):
         if not self.qnameExpressionProgs: # optimize if simple
-            qnamedFacts = set.union(*[inst.factsByQname(qn)
+            qnamedFacts = set.union(*[inst.factsByQname(qn, set())
                                       for inst in varBinding.instances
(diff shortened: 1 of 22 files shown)


gmongelli:
As for #36, this pull request has been outdated because of pull request #47 . Moreover, there are several supplemental conflicts regarding the EBA validation.

gmongelli:
Anyway, this pull request is also included in #38

PR #39, opened 2014-12-02
Create inexistent facts only

gmongelli:
Please only apply this pull request after having handled #38 (Performance improvement by using database indices). This pull request also contains all the changes from that pull request as it cannot work reliably without it.

For new facts with open dimensions, only really create them if they do not exist yet.
They exist if they are c-equal (context equal) and if they have the same QName.


derekgengenbacher-wf:

PR #40, opened 2014-12-02
Improved display of tables based on exact contexts

gmongelli:
Please only apply this pull request after having handled #35 (EBA filing indicators ). This pull request also contains all the changes from that pull request as it cannot work reliably without it.

An issue has been noticed with Arelle: when there are several open dimensions in the y axis, it might happen that a context is matched if all dimensions needed by a fact are present in it together with other irrelevant but over-specified dimensions.

This improvement aims at verifying that a context _exactly_ contains the needed dimensions and not more than the needed ones.


derekgengenbacher-wf:

PR #41, opened 2014-12-19
Proposal for improving the validation error messages

gmongelli:
The following changes apply to validation rules extracted automatically from formula linkbases.
1. Whenever there is no error message and a validation rule is not satisfied, display as fall-back the default label of the validation rule.
2. When a validation rule triggers an error message, also display the evaluation results. This new behaviour can be switched off in the validation parameters by un-checking the new "Expression in Error" checkbox (on by default).
3. When there is a Python exception or an error, also display the default label (full wordings) of the validation rule that is in error (if available).


derekgengenbacher-wf:

PR #46, opened 2015-02-05
Replace 'Save...' by 'Save' and 'Save As...' in the GUI

gmongelli:
This pull request suggests a new saving scheme.

The old 'Save...' has been split into a 'Save' and a 'Save As...' menu entry. 'Save' does not ask for a file name, it uses the current file name. If the file is new, a file name is asked nevertheless. The 'Save' icon now does the 'Save' operation. The 'Save As...' menu entry systematically asks for a new file name to save the current instance.

The name of the new file name replaces the old name in the current instance.

The file name visible in the window title is also updated.

Moreover, the recent files list is now also updated with the latest saved files.

Finally, before closing an instance, the program checks whether there is something to save. If there is, an alert message is displayed and the closing may be interrupted.

### DIFF
--- a/arelle/CntlrWinMain.py
+++ b/arelle/CntlrWinMain.py
@@ -434,21 +434,41 @@ class CntlrWinMain (Cntlr.Cntlr):
         self.data = {}
         self.parent.title(_("arelle - Unnamed"));
         self.modelManager.load(None);
-        
+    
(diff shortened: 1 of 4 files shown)


gmongelli:
Outdated as it is included in #47.

PR #47, opened 2015-09-24
New table rendering withTKTable because the current one is not effective for large tables

gmongelli:
## Problem Description

For some taxonomy endpoints like COREP 2.4, 'Additional Liquidity Monitoring - COREP, Individual', there are table linkbases that generate large tables with many columns and many rows. These are namely the tables  C 66.00.a and C 70.00.a.

When these tables are rendered, the rendering is abruptly stopped and TKInter does no longer respond to user inputs in the usual way.  
## Solution proposal

Currently the table structure is simulated by using a large number of 'Entry' elements.

In this proposal, the table structure is rendered by a real table widget, namely TkTable as found in the ActiveTcl distribution of Tk/Tcl.

By using TkTable, we get many interesting benefits:
- The rendering is faster (tables C 66.00.a and C 70.00.a are now rendered correctly)
- There is a built-in copy/paste function that works quite well for a whole range of cells, albeit it is still perfectible
- As a by-product, it was easy to change the background of the cells according to the expected content
- The code has become simpler because the table simulation is no longer needed.

The Python glue code has been taken from http://tkinter.unpythonic.net/wiki/TkTableWrapper, added as `arelle/TKTableWrapper.py` in the proposal and modified to correct the blocking errors.

The consequence of all this is that in order to make the table rendering work, one has to make sure that TkTable is in the search path used by TkInter!

Moreover, it has been chosen to store the data in a NumPy array because the default storage proposed by TkTable consumes too much memory.

For the preparation of the setups, there is a new dependence on NumPy.

I have verified that both the TkTable license and the NumPy license are compatible with the Apache license used by Arelle. However, NumPy requires that it should be mentioned in the 'About' box. This has not yet been done in this pull request!

This pull-request also modifies the Save behaviour of Arelle. It includes pull request #46 making it obsolete.
## Caveats

In the master branch of Arelle, it is correctly not possible to create a new XBRL instance from a COREP taxonomy endpoint. That problem has not been solved in the proposed branch and thus the creation of instances could not be tested in this pull request.


hefischer:
merged to Arelle branch TLB_TKTABLE

gmongelli:
Hy,

I just committed an improvement to avoid the nasty error "_tkinter.TclError: cannot set spanning on hidden cell 10,1"

I hope that this will make the solution usable in a larger range of rendering cases.

derekgengenbacher-wf:

PR #49, opened 2015-10-09
Drop-down lists for TkTable version of Arelle

gmongelli:
This is an implementation of table rendering where enumerations, qname values and booleans are presented as drop-down lists.
It has also some small bonuses like the one that makes the rendering faster by checking if the fact list is not yet empty before filtering again.


hefischer:
merged

PR #53, opened 2016-02-03
Library cleanup of XLRD and XLWT

chrislococo-wf:
#### Changes:

- Removed the installed XLRD and XLWT files from Arelle.
- Created an "attic" directory which can be used to store old code components that aren't currently in use but could provide value at a later date.
- Moved CntlrGenVersReports.py to the attic, as it is unused.
- Re-encoded the locale/messages.pot file to utf-8.
- Removed XLRD and XLWT references from the about in each of the three main controller entry points. 

Please review: @hefischer  @andrewperkins-wf


andrewperkins-wf:
+1 - this looks like a good set of changes to me

andrewperkins-wf:
+1

PR #54, opened 2016-02-05
Sort baseSet keys on ELR for consistent locators

jasonleinbach-wf:
When running on python 3.4, it was discovered that the rules within the affected loop could fire on different locators across subsequent runs.  This is due to python 3.4's hash randomization, which causes the keys to be returned in different order across those runs.

The fix is to sort the baseSet keys on ELR.


hermfischer-wf:
Thanks for identifying this problem and a fix.  Other validations and rendering also use the same kind of loop.

Checkin 353e914b775a2f31619994a75f1685d34ae4744d on 2016-02-06 implements an alternative, by replacing modelXbrl.baseSets with an OrderedDict after completion of loading an entry point and after adding a supplemental linkbase.

derekgengenbacher-wf:

PR #66, opened 2016-11-21
Webcache reload param rework

jasonleinbach-wf:


arelle/WebCache.py
```diff
@@ -290,26 +290,31 @@ def getfilename(self, url, base=None, reload=False, checkModifiedTime=False, nor
...
-                        remoteFileTime = lastModifiedTime( self.getheaders(quotedUrl) )
-                        if remoteFileTime and remoteFileTime > os.path.getmtime(filepath):
-                            newerOnWeb = True
-                    except:
-                        pass # for now, forget about authentication here
-                    if not newerOnWeb:
-                        # update ctime by copying file and return old file
-                        self.cachedUrlCheckTimes[url] = timeNowStr
-                        self.cachedUrlCheckTimesModified = True
+            if os.path.exists(filepath):
```
jasonleinbach-wf: If the file path exists, then check reload.

arelle/WebCache.py
```diff
@@ -290,26 +290,31 @@ def getfilename(self, url, base=None, reload=False, checkModifiedTime=False, nor
...
+                if reload:
```
jasonleinbach-wf: if reload is TRUE (as opposed to false in the existing impl), then proceed with the existing impl.

PR #68, opened 2017-05-16
Remove redundant test file

roberth:
testParser2.py loads the same module as testParser.py, but calls a non-existing function.

It seems as though testParser2 has become redundant, because it is mostly equivalent to testParser, but does not work.


roberth:
Is there anything more I can do to help?

roberth:
Never mind.

PR #70, opened 2017-06-30
Compatibility with python 3.6 in One character

CheeseForKneez:
Allows arelle to run in python 3.6
In python3.6, re raises sre_constants.error: bad escape \m

Changes are Easy to test:
import re
p = re.compile(r"&\w+;|[^0-9A-Za-z`~!@#$%&\*\(\)\.\-+ \[\]\{\}\|\\:;\"'<>,_?/=\t\n\r\\m\f]") 
fn = lambda x,txt: [match.group() for match in x.finditer(txt)]
stringL =  ['nnn' ,'_m_nnn', '_\m_nnn']
for s in stringL: print( fn(p,'text &{}; text'.format(s)) )


sphenk:
The better fix is #73.
Character \m is bogus. And 'm' is already in a-z.

PR #73, opened 2017-07-22
Remove \m from docCheckPattern re compilation

BitWorks:
For python >= 3.6.
https://docs.python.org/3/library/re.html
Changed in version 3.6: Unknown escapes consisting of '\\' and an ASCII letter now are errors.


derekgengenbacher-wf:
This change was made in an unknown commit. Thank you for the PR.

PR #74, opened 2017-08-02
Fix packages command line bug

seocahill:
Packages args were never being run by the command line controller.

Added an extra clause to conditional to fix.


derekgengenbacher-wf:

PR #76, opened 2017-08-31
Ignore error codes if "nonStandardErrorCodes" is specified.

pdw-mb:
The iXBRL conformance suite now has a "nonStandardErrorCodes" attribute, which is a flag saying that you can ignore the error codes and just get the pass/fail result right.  This commit adds support for the flag to Arelle.


hefischer:
This pull request seems to overlap 5c12fcc826a2a9e2bb9eb63473a3b030923f00b1 (not having noted this pull at that time)

PR #77, opened 2017-08-31
Xf prefix sl fixes

pdw-mb:
msg and cfi schemalocations were not being included in generated formula.  The xfi prefix was not always being declared.


hefischer:
Merged (5dac0737d0c604110a66954986dd02ee6ebba05b).

Thank you.

PR #78, opened 2017-09-15
Remove \m from regex in ValidateFilingText for Python 3.6

roberth:
Hello Herman Fischer,

While packaging Arelle for the NixOS Linux distribution, another contributor has come across what seem to be small problems preventing Arelle from working in Python 3.6. The fix in this PR seems to touch the actual validation code, but some further research seems to indicate that the behavior does not change. If you agree, you may merge this pull request.

If you're interested, our progress is here https://github.com/NixOS/nixpkgs/pull/29435

Below is the commit message with further info:

This removes what seems to be an escape sequence from a regular expressing that validates filing texts.
In support of this change, I have gathered the following information:
 - An \m escape results in an error in Python 3.6 
 - An \m escape is not in the Python 2 documentation
 - An \m escape does not seem to be suggested by the relevant part of the EFM specification table as mentioned in the comment
 - The backslash before an m in the pattern seems to be ignored by the re (regex) module until recent versions
 - The m character was matched by this regex anyway. It was redundant but has now become an error.


hefischer:
Merged (d570d8e1a2579c2f915ac374fd74b4e78b1f617a)
Thank you.

PR #79, opened 2017-11-08
Fixes the saving and update of facts for table linkbase

RobinetDenisAcsone:
Small patch to correct the issues met by Luigi Carini and Mike van der Ploeg: [https://groups.google.com/forum/#!topic/arelle-users/XDwQwoUaLLY](https://groups.google.com/forum/#!topic/arelle-users/XDwQwoUaLLY)

The problems happened when the user want to save a report and when he wants to update a existing fact.

The first problem (the save) appends because the definition of `saveInstance` allows only keyword
From ModelXbrl.py, line 468: 
`def saveInstance(self, **kwargs):`

The second appends because unitID wasn't settable in `ModelFact ` and in ViewWinRenderedGrid.py the update of the fact want to set it (line 1051)

### DIFF
--- a/arelle/ModelInstanceObject.py
+++ b/arelle/ModelInstanceObject.py
@@ -167,6 +167,11 @@ class ModelFact(ModelObject):
         """(str) -- unitRef attribute"""
         return self.getStripped("unitRef")
     
+    @unitID.setter
+    def unitID(self, value):
+        """(str) -- unitRef attribute"""
+        self.set("unitRef", value)
+    
     @property
     def utrEntries(self):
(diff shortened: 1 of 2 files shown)


RobinetDenisAcsone:
Hello @hefischer,

Can you check this pr to help the people from the arelle group thread https://groups.google.com/forum/#!topic/arelle-users/XDwQwoUaLLY ?

Thanks
Denis

hefischer:
Merged in commit bdb22d577d5b4c59d1253464c3623799a1d4e234

PR #80, opened 2017-12-14
Preserve whitespace when processing iXBRL if there is no format 

pdw-mb:
Whitespace should be preserved, except for tuple children.

All current string transforms have whitespace="collapse", but in the absence of a format, it should be preserved.


hefischer:
Merged 2e9ee1cb1f242a0db904dc7e1e9e0811e15d2c39

PR #82, opened 2018-01-29
Add support for detecting JPEG images with Exif data (APP1 segment).

jeffthieleke-wf:
Issue: EFM XBRL validation fails with a "EFM.5.02.05.graphicFileContent" error if the content references a JPEG image which contains Exif data. 

The Edgar Filer Manual doesn't specify, one way or the other, if Exif data is allowed in JPEG images.  But since it is only metadata added to an otherwise compliant image format and isn't active content, and since all modern browsers and software are able to display and manipulate such images, it would seem reasonable to allow it to pass Arelle's validation.  

Example SEC disseminated JPEG image file with Exif data: https://www.sec.gov/Archives/edgar/data/920760/000162828018000562/lennarlogo.jpg

References:
http://www.exif.org/Exif2-2.PDF (4.7 JPEG Marker Segments Used in Exif)
http://www.ozhiker.com/electronics/pjmt/jpeg_info/app_segments.html


danaoredson-wf:
@hefischer , is there anything preventing this PR from being accepted into Arelle?

PR #86, opened 2018-05-25
add internet retries flag

jonreiter:
add --internetRetries command line argument and make the retry count in webcache adjustable instead of being a constant


derekgengenbacher-wf:

PR #87, opened 2018-06-27
Accumulated fixes from pdw

pdw-mb:
XF fixes for explicit-dimension filter and parameter declaration.

Change to conformance test summary stats reporting (now reports on a single testcase as well as a full test suite).

Support for nonStandardErrorCodes test suite attribute.


hefischer:
Merged.

Regarding missing space in explicit-dim filter, think that same space needed in typed-dim line 260?  Added it.

mastercommit 370c8a011107e139b54b4a18a40cb7f5ccbc0020

PR #88, opened 2018-08-07
Build docker container for Arelle

fhalim:
Added a very basic dockerfile for running Arelle as a server.


derekgengenbacher-wf:

PR #89, opened 2018-09-12
Xf syntax fixes

pdw-mb:


derekgengenbacher-wf:

PR #91, opened 2018-10-24
Fix the warning message for the DEI WKSI / Submission WKSI value mismatch

jeffthieleke-wf:
A customer received the following XBRL Warning while test filing a 10-K today:

```
WRN:  XBRL WARNING MESSAGE
MSG:  Warning: [dq-0540-Entity-Well-Known-Seasoned-Issuer-Value] 
      dei:EntityWellKnownSeasonedIssuer value Yes in the Required Context does 
      not agree with submission voluntary filer flag value False. 
      foo-20180831.xml line 1234 
LOC:  LINE NUMBER: 0
```

I was confused until I looked at the Arelle code - the check appears to be correct but the error message is wrong.


hefischer:
Fixed by check-in's to Arelle - master (plugin/validate/EFM/Filing.py) and EdgarRenderer -  edgr191preview (resources/arelleMessagesText.xml) from SEC.

PR #92, opened 2018-12-13
Comprehensive dependencies

piotrsroka:
When installing Arelle with use of either pip or setup.py two dependencies are missing:
* isodate 
* openpyxl

which forces the user to install them manually.
Also, egg directories are not excluded from the git.

This PR is fixing those two issues.


hefischer:
Merged to master, thank you.

PR #94, opened 2019-02-20
[FIX] Make sure that xAttributes is initialized when validating the defintion node rules

RobinetDenisAcsone:
When validating some taxonomies, xAttributes can be uninitialized, which will cause a false `xbrlte:incompleteAspectRule` error.

You can find a file for testing here:
https://groups.google.com/forum/#!topic/arelle-users/HUkF9MF9yH4


hefischer:
Thank you.  Merged to master.

PR #96, opened 2019-09-25
Expand valid years for 'regYr' and 'finYr' of filename regex

sagesmith-wf:
Changes:
regYr used to only accept years [2010 - 2029] that range has been expanded to [1800 - 2099]
finYr used to only accept years [2010 - 2029] that range has been expanded to [2010 - 2099]

Ultimate Problem:
Companies registered before 2010 throw a warning in arelle, but not on the cipc filing website.
     ex: 'Ceramic Industries Proprietary Limited - 1982-008520-07 - 2018.xhtml' filename generates no warnings from the cipc website but throws an error in arelle


hefischer:
Merged to master

PR #97, opened 2019-10-15
Updated SBRnl plugin

ekaats:
Rewrote some logic in __init__.py as not all validations were properly called.
Updated definitions in sbr-nl-taxonomies.py to reflect currently allowed schemas (possibly still incomplete)
Renamed NTA rules (from x.x.x.xx to x.xx.xx.xx)
Deprecated and/or updated some checks.


derekgengenbacher-wf:

PR #98, opened 2020-01-07
change return value in function  eloValueOfFact from null to None

OlegShteynbuk:
change return value in function  eloValueOfFact from null to None


hefischer:
Good catch!

This will be merged for EDGAR release 20.1.1 (public release in March), so the actual public visibility will be a month before then (sometime in February).

derekgengenbacher-wf:
Already merged

PR #99, opened 2020-02-24
Set User-Agent for web requests

parthjoshi2007:
There needs to be a way to allow the User-Agent header to be sent for web requests made by Arelle. Eg. the Indian MCA does not allow access to the taxonomy without a User-Agent header (http://www.mca.gov.in/XBRL/2016/07/26/Taxonomy/CnI/in-ci-ent-2016-03-31.xsd)


hefischer:
Thank you for the pull request.

Checked in with c411c075bcb91c67c65534d6f0d895c259d3b947

PR #100, opened 2020-02-24
Provide a parsedValue property for ModelFact

parthjoshi2007:
When using Arelle to parse and extract data from an XBRL instance, most of the time what you need is a Python value that you can then manipulate (eg. decimal.Decimal for numbers or True/False for booleans) rather than the string value which is what effectiveValue provides. The parsedValue property adds support for exactly that.

### DIFF
--- a/arelle/ModelInstanceObject.py
+++ b/arelle/ModelInstanceObject.py
@@ -407,7 +407,32 @@ class ModelFact(ModelObject):
             return val
         except Exception as ex:
             return str(ex)  # could be transform value of inline fact
-
+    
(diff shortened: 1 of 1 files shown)


hefischer:
The parsed value is f.xValue.  This property may be unavailable if there was some reason the fact wasn't parsed at all, so most code has the check if getattr(f, "xValid",0) >= VALID (where VALID=4 is imported from arelle.XmlValidate).  See for example ValidateXbrl line 749.

derekgengenbacher-wf:

PR #101, opened 2020-03-27
Fix HTML/XML export of rendered tables

akochurov:
In ViewFileRenderedGrid.py there was one less condition than in
ViewWinRenderedGrid.py which resulted in wrong facts being displayed
in the exported HTML tables.


akochurov:
I discovered that when the tables exported to HTML, the wrong values are displayed.

Example: 

Here's an XBRL file (based on EBA-provided skeleton instance):
[sample_corep_le_con.zip](https://github.com/Arelle/Arelle/files/4394252/sample_corep_le_con.zip)

Here's how Arelle displays it in the GUI:
arelle-1

Here's how the exported HTML looks like:
arelle-2

hefischer:
Thank you.  Change merged with 1403fe43a52d6ded6cd74112e68eeec8918a8f61

PR #105, opened 2020-08-10
make memberType errors easier to diagnose

hamscher:


hamscher:
memberType errors are hard to diagnose because the error message doesn't say exactly where in the JSON file the error occurred.  Also there seems to be no way to provide a line number to the hooks in json.load.  This change creates a more elaborate error message in loadFromOIM on member type errors.

derekgengenbacher-wf:

PR #107, opened 2020-08-11
Fix numeric duplicate fact check to fire for any facts with same context, unit, concept, and decimals

sagesmith-wf:
### Description
The original code wouldn't fire a duplicate fact error unless all facts with the same context, unit, and concept had the same values and decimals. This commit changes the check to group facts with the same context, unit, and concept into groups based on the decimals. If we find multiple facts with the same decimals then we proceed with checking the values. This method of checking allows arelle to correctly identify all duplicate facts and fire errors for them

Specifically looking at the following 3 facts from  [duplicate-fact-error.zip](https://github.com/Arelle/Arelle/files/5058923/duplicate-fact-error.zip):

| Tag                          | Fact      | Period                     | Measure | Decimals          |
|------------------------------|-----------|----------------------------|---------|-------------------|
| us-gaap:RestructuringCharges | 748,000   | 3 months ending 06/30/2020 | USD     | Thousands         |
| us-gaap:RestructuringCharges | 749,000   | 3 months ending 06/30/2020 | USD     | Thousands         |
| us-gaap:RestructuringCharges | 0,700,000 | 3 months ending 06/30/2020 | USD     | Hundred thousands |

The first 2 should have fired a duplicate fact error, but didn't due to the third fact having a different decimals.

### Side Effects
Filings with duplicate facts that should be rejected by the SEC are not being correctly rejected

### Testing Suggestions
Validate using [duplicate-fact-error.zip](https://github.com/Arelle/Arelle/files/5058923/duplicate-fact-error.zip). You should see 2 duplicate fact errors that weren't fired before.

### Please Review
@hefischer

### DIFF
--- a/arelle/plugin/validate/EFM/Filing.py
+++ b/arelle/plugin/validate/EFM/Filing.py
@@ -812,13 +812,23 @@ def validateFiling(val, modelXbrl, isEFM=False, isGFM=False):
                     aspectEqualFacts[(f.qname,f.contextID,f.unitID,
                                       f.xmlLang if f.concept.type.isWgnStringFactType else None)].append(f)
                 for fList in aspectEqualFacts.values():
+                    duplicate_numeric_fact_group = None
+                    first_decimal_equivalent_fact = None
                     f0 = fList[0]
                     if f0.concept.isNumeric:
+                        facts_by_decimals = defaultdict(list)
+                        for f in fList:
+                            facts_by_decimals[inferredDecimals(f)].append(f)
                         if any(f.isNil for f in fList):
                             _inConsistent = not all(f.isNil for f in fList)
-                        elif all(inferredDecimals(f) == inferredDecimals(f0) for f in fList[1:]): # same decimals
-                            v0 = rangeValue(f0.value)
-                            _inConsistent = not all(rangeValue(f.value) == v0 for f in fList[1:])
+                        elif any(len(decimal_equivalent_facts) > 1 for decimal_equivalent_facts in facts_by_decimals.values()): # same decimals
+                            for decimal_equivalent_facts in facts_by_decimals.values():
+                                if len(decimal_equivalent_facts) > 1:
+                                    first_decimal_equivalent_fact = decimal_equivalent_facts[0]
+                                    v0 = rangeValue(first_decimal_equivalent_fact.value)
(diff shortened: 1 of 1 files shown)


sagesmith-wf:
This PR is wrong based on the spec. The spec is really confusing, but all three numbers are part of the same interval.

2.2 Consistent duplicates (numeric facts)
Facts that fulfil the XBRL v2.1 definition of duplicate items, and which additionally:

are numeric facts; and
have reported values and accuracies that are consistent with having been rounded from a single actual value.
Consistency of numeric values should be established by considering the interval represented by each reported fact, and ensuring that there is overlap between all intervals.

For example, given the reported values 2,500 (stated to -2 decimal places) and 2,000 (stated to -3 decimal places), we have the following intervals:

2,500 (-2 d.p.) => [2450, 2550]
2,000 (-3 d.p.) => [1500, 2500]
These two intervals overlap, and so are considered consistent (both facts are consistent with having been rounded from a value in the range [2450,2500]).

Technically, the interval represented by a reported value depends on the rounding method used to obtain it. For example, 2,000 (-3 d.p.) is only a valid rounding of 2,500 under certain rounding methods (e.g. "round half to even" or "bankers rounding"). If the value were rounded using the "round half away from zero" (or "commerical rounding") method, it would be 3,000 (-3 d.p.).

In order to avoid declaring values to be inconsistent in this edge case, the intervals used in establishing consistency should be inclusive (or "closed"), but with the constraint that values reported to the same accuracy must have the same reported value. In other words, 2,000 (-3 d.p.) is inconsistent with 3,000 (-3 d.p.) even though there is overlap between the closed intervals; such values would imply an actual value of 2,500, but rounded using different methods in different places in the report.

The approximation of using closed intervals only affects the outcome of the consistency check in the event that the exact value is also included in the report stated to infinite precision.

Numeric facts that are Complete Duplicates are also considered to be Consistent Duplicates.

hefischer:
The duplicates detection has been changed to match this spec: http://www.xbrl.org/WGN/xbrl-duplicates/WGN-2018-04-19/xbrl-duplicates-WGN-2018-04-19.html#sec-complete-duplicates.  The three candidate duplicates above are all consistent, now with open interval checks, because as a group, 748000 to  thous => [747500 - 748500], 749 => [748500 - 749500] so consistent having been derived both from 748500 and 700000 to hund thou => [650000 - 750000] which includes 748500, so they're all consistent with having been derived from 748500

PR #109, opened 2020-09-24
update ESEF validation guidances 2.1.2, 2.5.3, 3.1.1, 3.4.5

ChristopheDeBernardi:
### Regarding ESEF validation:
updates four messages in regards to the latest update of the reporting manual.

* **G2.1.2**: changed message and severity level (to "error")
The latest update of the manual is stricter on this guidance

* **G2.5.3**: changed message and severity level (to "warning")
The latest update of the manual discourages but allows the use of the "target" attribute (if required by local jurisdictions)

* **G3.1.1**: changed severity level (to "warning") and message code in the case where several linkbases are in the same xml file (not the .xsd)
The message is: "Each linkbase type SHOULD be provided in a separate linkbase file"
The message code has been changed from "extensionTaxonomyWrongFilesStructure" to "linkbasesNotSeparateFiles", for which the manual sets the level to "warning"

* **G3.4.5**: changed severity level (to "error") and message ("SHOULD" to "SHALL").
While the manual does not explicitly set a severity level, it says that "Each taxonomy extension element **shall** be defined with at most one label for any combination (...)"

### DIFF
--- a/arelle/plugin/validate/ESEF/DTS.py
+++ b/arelle/plugin/validate/ESEF/DTS.py
@@ -139,8 +139,9 @@ def checkFilingDTS(val, modelDocument, visited, hrefXlinkRole=None):
                                 modelObject=modelConcept, concept=modelConcept.qname, lc3names=", ".join(lc3names))
                         for (lang,labelrole),labels in langRoleLabels.items():
                             if len(labels) > 1:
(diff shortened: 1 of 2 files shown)


hefischer:
Thank you for identifying these updates, they've been merged with commit df0458eeb4657fce1f8b666f85d3c90e38559a62

PR #110, opened 2020-09-30
Fix displaced cells in HTML export

akochurov:
Skip Y-axis values that are abstract and do not have any children
during the HTML rendering in `ViewFileRenderedGrid`.

Such `<th>` cells were inserted before header cells from the next
row and pushed them to the right, breaking the table structure.
There are no facts reported for such Y-axis values and hence there's
no need to include them in the table at all.

An example of such a table with displaced cells is EBA DPM 2.8 `C 29.00`.


akochurov:
Here's an example:

- corrupted HTML table for C 29.00:
image

- it is displayed correctly in the UI:
image

- here's the sample XBRL file and the generated HTML file:
[xbrl.zip](https://github.com/Arelle/Arelle/files/5304191/xbrl.zip)

derekgengenbacher-wf (approved):

sagesmith-wf (approved):

austinmatherne-wk (approved):

hermfischer-wf (approved):

PR #111, opened 2020-10-19
Add the value that couldn't be atomized to the error message

philippdolder:
The error message misses the information what value could not be atomized.

Note: Unfortunately the contributor agreement linked to from https://arelle.atlassian.net/wiki/spaces/ARELLE/overview returns me a 404


hefischer:
Merged, thanks for identifying this issue.

PR #117, opened 2021-06-23
Allow non-rec units to be loaded

sagesmith-wf:
### Description
Arelle has a convenient util for parsing units, but it doesn't allow you to load all utr units. This adds a simple bool that allows all utr units to be loaded, not just the units with status "REC". The default behavior of only loading units with status "REC"  is unchanged

please review: @hefischer


hefischer:
Merged
[master a47e8cc] Pull request #117

PR #118, opened 2021-07-01
Fix '[Errno 22] Invalid argument' in validate/EFM-htm

ChristopheDeBernardi:
This issue was originally reported on [Google Groups](https://groups.google.com/g/arelle-users/c/752ylmruI3k)

### Problem

Arelle raises an unhandled exception while trying to load a report from the SEC archives with the validate/EFM disclosure system check.

image

### Cause

In validate/EFM-htm/__init__.py, isLoadableHtml(), calling `io.open(filepath, 'rt', encoding='utf-8')` with `filepath` being a web url.

Moreover, the designated .htm file is inside a .zip archive.

### Implemented fix

1. Checking if the `filepath` value is a web URL with `isHttpUrl(filepath)`  
if true, getting the local file path with `modelXbrl.modelManager.cntlr.webCache.getfilename()` (same as in ModelDocument.load())
2. Checking if `filepath` resolves to an actual file using `Path(filepath).is_file()`  
If not, using pathlib and zipfile modules to look for a zip archive in the path, then using `ZipFile.open()` instead of `io.open()` to read the .htm file inside the archive

### DIFF
--- a/arelle/plugin/validate/EFM-htm/__init__.py
+++ b/arelle/plugin/validate/EFM-htm/__init__.py
@@ -6,7 +6,8 @@ Created on Dec 12, 2013
 
 
 '''
-import os, io
+import os, io, zipfile
+from pathlib import Path
 from lxml.etree import HTMLParser, parse, DTD, _ElementTree, _Comment, _ProcessingInstruction
 try:
     from regex import compile as re_compile, match as re_match, DOTALL as re_DOTALL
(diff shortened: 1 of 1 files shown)


hefischer:
Fixed with commit add7f00 using an approach with filesource .file, to accommodate zips and also possibility of URI remappings, alternative archives (SEC EIS or gzip), etc.

PR #124, opened 2021-09-14
Adding translation for ar-EG (Arabic Egypt)

selgamal:
## About
This PR adds Arabic Language to Arelle and proposes fixes relating to presentation of right to left languages as mentioned in the commit messages below. 

### Translation
Translations used for XBRL terminology listed [here](https://selgamal.github.io/xbrl-glossary-collection/xbrl-glossary-collection.html) for reference. This translation follows naming conventions commonly used in Egypt, generally it is similar to other Arabic speaking regions, but there might be slight differences.

### Right to left proposed changes (included in this PR):
1.  Set `html` attribute `dir` to `rtl` when a right to left language is detected for `html` views.
2. Add an option to GUI and CmdLine to disable labels right to left processing in [~arelle/Locale.py:256](https://github.com/Arelle/Arelle/blob/745f8940e0c1e374d35179b12c609de1b325cca6/arelle/Locale.py#L256), the reason being in brief; rtl is supported by tkinter on windows, and by `html`, so actually rtlString ends up reversing what was actually rendered correctly. It is only needed for GUI on `linux` because tkinter on linux does not support rtl, and in the case of Arabic language, it needs [`arabic_reshaper`](https://pypi.org/project/arabic-reshaper/) also to render properly, because Arabic letters change shape according to position in the word and it is not enough to just reverse the order, please see [here](https://stackoverflow.com/questions/4150053/python-tkinter-using-tkinter-for-rtl-right-to-left-languages-like-arabic-hebr) and [here](https://wiki.tcl-lang.org/page/bidi+rendering) for more on this. (Tested on windows 10 and Debian 10)  

The proposed fix does not change Arelle current behavior, just gives the user the option to disable rtlString processing.

**Arabic labels renders correctly in html view when rtlString is disabled:**
html-correct

**Arabic labels renders in reverse in html view when rtlString is not disabled**
html-wrong

**Arabic labels renders correctly in windows GUI when rtlString is disabled**
gui-correct

**Arabic labels renders in reverse in windows GUI when rtlString is not disabled**
gui-wrong


hefischer (commented):
Merged

hefischer (commented):
merged

hefischer (commented):
The u string prefix should not be needed by python 3, is not compatible with early versions of 3 for which Arelle is still in use.  This is not merged.

selgamal:
> The u string prefix should not be needed by python 3, is not compatible with early versions of 3 for which Arelle is still in use. This is not merged.

The u string is removed, the last commit today reverts the previous u string commit, translation still works fine.

The reason I added the u string was because with out it, the .po file generated (by poedit) adds an escape backslash to special charterers, for example `"arelle\u00ae"` becomes `"arelle\\u00ae"` (I think that's why it wasn't matching the translation). Now that the .po is generated and .mo compiled it will still match even without the u string.

hefischer:
Merged with changes 
[master 77d0a5c] pull #124 add ar-EG language support
If you can test please contact support@arelle.org, we can do a zoom session to see how it looks.
Thank you

selgamal:
> Merged with changes [master [77d0a5c](https://github.com/Arelle/Arelle/commit/77d0a5c8dfea2f4b9830619af5a99b34c75f7636)] pull #124 add ar-EG language support If you can test please contact [support@arelle.org](mailto:support@arelle.org), we can do a zoom session to see how it looks. Thank you

Email sent.

PR #129, opened 2022-01-14
Fix insert of resource value in xbrlDB plugin for SQL database

kris2kris:
Insert value in resource table doesn't work for some kind of data when using textValue property because value is empty (for example with ruleNode of EBA taxonomy)
Semantic database other than SQL already used stringValue property for resource value


derekgengenbacher-wf (approved):

sagesmith-wf (approved):

austinmatherne-wk (approved):

hermfischer-wf (approved):

derekgengenbacher-wf:

PR #130, opened 2022-01-27
Add ROS 2022 to allowable taxonomies for ROS validations

derekgengenbacher-wf:
#### Description:
Added hrefs for ROS 2022 as allowable entry points for ROS validations

**review**:
@hefischer

### DIFF
--- a/arelle/plugin/validate/ROS/__init__.py
+++ b/arelle/plugin/validate/ROS/__init__.py
@@ -21,11 +21,14 @@ from arelle.ValidateXbrlCalcs import inferredDecimals, rangeValue
 from arelle.XbrlConst import qnXbrliXbrl, xhtml
 from arelle.XmlValidate import VALID
 
-taxonomyReferences = {
-    "https://xbrl.frc.org.uk/ireland/FRS-101/2019-01-01/ie-FRS-101-2019-01-01.xsd": "FRS 101 Irish Extension",
-    "https://xbrl.frc.org.uk/ireland/FRS-102/2019-01-01/ie-FRS-102-2019-01-01.xsd": "FRS 102 Irish Extension",
-    "https://xbrl.frc.org.uk/ireland/IFRS/2019-01-01/ie-IFRS-2019-01-01.xsd": "EU IFRS Irish Extension"
-    }
+taxonomyReferences = [
+    "https://xbrl.frc.org.uk/ireland/FRS-101/2019-01-01/ie-FRS-101-2019-01-01.xsd",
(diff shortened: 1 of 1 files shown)


sagesmith-wf (approved):
Multiple keys pointing to the same value seems a bit odd, wondering if it would make more sense to add the year to the RHS or to do a `https://xbrl.frc.org.uk.ireland/FRS-101 in` check for example instead. 🤷

austinmatherne-wk (approved):

hermfischer-wf:
Agree with Sage's comment on the year in RHS (but need to see how that looks where it is actually used)

derekgengenbacher-wf:
Realistically that does not need to be a dict. It can be a list of allowable taxonomy references. Ill update the commit to reflect that.

arelle/plugin/validate/ROS/__init__.py
```diff
@@ -21,11 +21,14 @@
...
-    "https://xbrl.frc.org.uk/ireland/IFRS/2019-01-01/ie-IFRS-2019-01-01.xsd": "EU IFRS Irish Extension"
-    }
+taxonomyReferences = [
+    "https://xbrl.frc.org.uk/ireland/FRS-101/2019-01-01/ie-FRS-101-2019-01-01.xsd",
+    "https://xbrl.frc.org.uk/ireland/FRS-102/2019-01-01/ie-FRS-102-2019-01-01.xsd",
+    "https://xbrl.frc.org.uk/ireland/IFRS/2019-01-01/ie-IFRS-2019-01-01.xsd",
+    "https://xbrl.frc.org.uk/ireland/FRS-101/2022-01-01/ie-FRS-101-2022-01-01.xsd",
+    "https://xbrl.frc.org.uk/ireland/FRS-102/2022-01-01/ie-FRS-102-2022-01-01.xsd",
+    "https://xbrl.frc.org.uk/ireland/IFRS/2022-01-01/ie-IFRS-2022-01-01.xsd",
+]
```
austinmatherne-wk, line 31: nit: I would keep these sorted

austinmatherne-wk (approved):

sagesmith-wf (approved):

arelle/plugin/validate/ROS/__init__.py
sagesmith-wf: I can go either way on that. I'm imagining a world where the list is long and its actually easier to have it ordered this way unless the datastructure is changed

sagesmith-wf (approved):

austinmatherne-wk (approved):

hermfischer-wf:
approve

derekgengenbacher-wf:
Closing because commit was made here: [master 7aafba2] #130 Add ROS 2022 to allowable taxonomies for ROS

PR #134, opened 2022-01-30
Fix escaping of formula messages

paulwarren-wk:
This PR fixes the escaping of taxonomy-supplied formula messages (and potentially user-supplied fact data).

Without this PR, any custom validation messages from a formula linkbase are used directly as the message that gets added to the log.  Python's `logging` applies `%` substitutions to these messages, to include any provided fields, and/or various standard fields.

For example, if you include `%-D` in a formula validation message, you will get the following output from Arelle:
```
Message log error: unsupported format character 'D' (0x44) at index 100
```
If the formula message includes the current fact value in the message, then this can be triggered by end-user supplied data. 

I originally noticed this because the error was causing an uncaught exception in Arelle, when validating a particular ESEF report (the ESEF duplicate facts formula rule includes the fact value) although I can't now reproduce the exception.  This bug does allow leaking of internal information, as any of the [standard LogRecord attributes](https://docs.python.org/3/library/logging.html#logrecord-attributes) can be inserted into messages.

### DIFF
--- a/arelle/FormulaEvaluator.py
+++ b/arelle/FormulaEvaluator.py
@@ -86,7 +86,8 @@ def evaluate(xpCtx, varSet, variablesInScope=False, uncoveredAspectFacts=None):
                 xpCtx.modelXbrl.log(
                     "INFO" if result else {"OK":"INFO", "WARNING":"WARNING", "ERROR":"ERROR"}[unsatSeverity],
                     "message:" + (varSet.id or varSet.xlinkLabel or  _("unlabeled variableSet")),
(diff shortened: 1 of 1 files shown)


hermfischer:
+1 HF

sagesmith-wf (approved):

derekgengenbacher-wf (approved):

derekgengenbacher-wf:
@paulwarren-wk Can you update the PR description with possible ways that this PR could be tested, please?

derekgengenbacher-wf:
QA +10 Testing locally succeeded. Ran several tests against formulas with modified labels. No regressions were found.

PR #135, opened 2022-01-31
Initial implementation of Calculations 1.1

hermfischer:
Initial implementation of Calculations 1.1 spec. 

**Replaced by [#680 - rebased REC implementation](https://github.com/Arelle/Arelle/pull/680)**


arelle/ValidateXbrlCalcs.py
```diff
@@ -117,54 +145,86 @@ def validate(self):
...
-                                                    if (itemConcept is not None and 
-                                                        (itemConcept, ancestor, contextHash, unit) not in self.itemFacts):
-                                                        unreportedContribingItemQnames.append(str(itemConcept.qname))
-                                                self.modelXbrl.log('INCONSISTENCY', "xbrl.5.2.5.2:calcInconsistency",
-                                                    _("Calculation inconsistent from %(concept)s in link role %(linkrole)s reported sum %(reportedSum)s computed sum %(computedSum)s context %(contextID)s unit %(unitID)s unreportedContributingItems %(unreportedContributors)s"),
-                                                    modelObject=wrappedSummationAndItems(fact, roundedSum, _boundSummationItems),
+                                        if not fact.isNil:
+                                            if fact in self.duplicatedFacts:
+                                                dupBindingKeys.add(sumBindKey)
+                                            elif (xbrl21 and sumBindKey in boundSums and sumBindKey not in dupBindingKeys
```
brettkail-wk, line 188: This is fine, but it seems like it would be clearer to hoist the `if xbrl21` to the `for fact in sumFacts` so that `if xbrl21` and `if calc11` are at the same level.  (I think that's equivalent since `dupBindingKeys` is only used by this xbrl21 loop anyway?)

arelle/ValidateXbrlCalcs.py
```diff
@@ -278,6 +338,43 @@ def bindFacts(self, facts, ancestors):
...
+                if isnan(_v):
+                    _inConsistent = True
+                    break
+                if _d in decVals:
+                    _inConsistent |= _v != decVals[_d]
+                else:
+                    decVals[_d] = _v
+                a, b = rangeValue(_v, _d)
+                if a > aMax: aMax = a
+                if b < bMin: bMin = b
```
brettkail-wk, line 365: This is fine, but it suggest moving these lines into the `else`?  Either this is the first time we've seen this value at this decimal (and we need to expand the range), or we're just updating `_inConsistent`.

brettkail-wk (approved):

arelle/ValidateXbrlCalcs.py
```diff
@@ -278,6 +338,43 @@ def bindFacts(self, facts, ancestors):
...
+                _v = f.xValue
+                if isnan(_v):
+                    _inConsistent = True
+                    break
+                if _d in decVals:
+                    _inConsistent |= _v != decVals[_d]
+                else:
+                    decVals[_d] = _v
+                a, b = rangeValue(_v, _d)
+                if a > aMax: aMax = a
```
sagesmith-wf, line 364: Nit suggest: `aMax = max(aMax, a)`, same for `bMin = min(bMin, b)`

sagesmith-wf (dismissed):

scripts/runXBRL21Tests.sh
```diff
@@ -0,0 +1,15 @@
...
+
+# Run XBRL 2.1 Conformance Suite tests
+
+TESTCASESROOT=/users/hermf/Documents/mvsl/projects/XBRL.org/conformance-svn/trunk
+rm -f /users/hermf/temp/XBRL21-conf-*
+OUTPUTLOGFILE=/users/hermf/temp/XBRL21-conf-log.txt
+OUTPUTERRFILE=/users/hermf/temp/XBRL21-conf-err.txt
+OUTPUTCSVFILE=/users/hermf/temp/XBRL21-conf-report.xlsx
+TESTCASESINDEXFILE=$TESTCASESROOT/xbrl.xml
+ARELLEDIR=/users/hermf/Documents/mvsl/projects/arelle/arelleproject/src
```
sagesmith-wf, line 11: Nit: seems like your computer specific for a bunch of these values

arelle/ValidateXbrlCalcs.py
```diff
@@ -468,6 +565,12 @@ def roundValue(value, precision=None, decimals=None, scale=None):
     return vRounded
 
 def rangeValue(value, decimals=None):
+    if isinstance(value, list):
+        if len(value) == 1 and isinstance(value[0], ModelFact):
+            return rangeValue(value[0].xValue, inferredDecimals(value[0]))
+        # return intersection of fact value ranges
+    if isinstance(value, ModelFact):
```
paulwarren-wk, line 572: This can cause an exception.

`rangeValue` is used by ESEF validation (and possibly other places).  If it is called before the `validate` function above, `ModelFact` doesn't get initialised and so is `None`, which causes `isinstance` to throw an exception.

I assume that `ModelFact` isn't imported directly due to a circular dependency, but I've not chased it down.

paulwarren-wk: This appears to be the only use of `ModelFact` in this file, so moving the import into this function would be one fix.

I done think there's any benefit to using the global + conditional import compared to just putting the import inside a function: Python won't import the same file twice.

derekgengenbacher-wf (dismissed):

arelle/ValidateXbrlCalcs.py
hermfischer-wf: somewhere at sometime it was in the list of bad and inefficient python practices to import in functions

scripts/runXBRL21Tests.sh
hermfischer-wf: All the test scripts under "scripts" have a similar setup, there should be a way to improve  them for CI as well as stand alone operation.

arelle/ValidateXbrlCalcs.py
hermfischer-wf: Prefer inline code here since it's an inner loop the function overhead would be greater than inline (I think, unless python optimizes).  This could all be redone in Cython anyway when time permits.

arelle/ValidateXbrlCalcs.py
hermfischer-wf: The summation facts if xbrl21 and calc11 are at different levels because for xbrl21 the sum facts are processed one by one (with different effects when dups are allowed) whereas with calc11 the sum facts are processed as an intersection of consistent intervals at the top level.

arelle/ValidateXbrlCalcs.py
brettkail-wk: I understand, so in the case where we have multiple values at the same decimals, either the secondary values equal the first (so `aMax`/`bMin` won't change) or `_inConsistent will be set, so we could just write:
```
                if _d in decVals:
                    _inConsistent |= _v != decVals[_d]
                else:
                    decVals[_d] = _v
                    a, b = rangeValue(_v, _d)
                    if a > aMax: aMax = a
                    if b < bMin: bMin = b
```

arelle/ValidateXbrlCalcs.py
brettkail-wk: I meant that [this whole loop](https://github.com/Arelle/Arelle/blob/af5f948a53baac66132754e254fad24e7aeeda76/arelle/ValidateXbrlCalcs.py#L184) seems to do nothing unless `xbrl21` is true, so it would be clearer to hoist the `xbrl21` so that the two checks are at the same level:
```
                                if factKey in self.sumFacts:
                                    sumFacts = self.sumFacts[factKey]
                                    if xbrl21:
                                        for fact in sumFacts:
                                            ...
                                    if calc11:
                                        ...
```

arelle/ValidateXbrlCalcs.py
sagesmith-wf: This is simple enough that it doesn't matter that much to me whether you change it, but in general I don't think function overhead is very expensive for most languages, or at least to the point of not calling a function to save on execution time that seems like a micro optimization to me.

arelle/ValidateXbrlCalcs.py
hermfischer: Ah, gotcha, done.

arelle/ValidateXbrlCalcs.py
hermfischer: done

sagesmith-wf (approved):

aviary3-wk:
## Security Insights

## Action Items
---

sagesmith-wf (commented):
Could we add some tests? This is pretty complicated

arelle/ValidateXbrlCalcs.py
```diff
@@ -24,15 +48,29 @@
...
     
 class ValidateXbrlCalcs:
-    def __init__(self, modelXbrl, inferDecimals=False, deDuplicate=False):
+    def __init__(self, modelXbrl, validateCalcs):
         self.modelXbrl = modelXbrl
-        self.inferDecimals = inferDecimals
-        self.deDuplicate = deDuplicate
+        self.inferDecimals = validateCalcs == ValidateCalcsMode.XBRL_v2_1
+        self.deDuplicate = validateCalcs == ValidateCalcsMode.XBRL_v2_1_DEDUPLICATE
+        self.xbrl21 = ValidateCalcsMode.XBRL_v2_1_INFER_PRECISION <= validateCalcs <= ValidateCalcsMode.XBRL_v2_1_DEDUPLICATE
```
brettkail-wk, line 70: I would find `in` to be clearer than relying on those constants being adjacent.

arelle/ValidateXbrlCalcs.py
```diff
@@ -49,61 +88,89 @@ def __init__(self, modelXbrl, inferDecimals=False, deDuplicate=False):
...
+        calc11 = self.calc11 # round or truncate
+        calc11t = self.calc11t # truncate
+        sumConceptItemRels = defaultdict(dict) # for calc11 dup sum-item detection
+          
+        if xbrl21:
+            if not self.modelXbrl.contexts and not self.modelXbrl.facts:
+                return # skip if no contexts or facts (note that in calc11 mode the dup relationships test is nonetheless required)
+            if not self.inferDecimals: # infering precision is now contrary to XBRL REC section 5.2.5.2
+                modelXbrl.info("xbrl.5.2.5.2:inferringPrecision","Validating calculations inferring precision.")
+        if calc11:
```
brettkail-wk, line 102: Is there a reason not to use `elif` here (or below)?

Should there be a final `else: return` to handle `NONE`?

hermfischer: elif and in suggestions done

A final return to handle NONE would defeat checking for essence-alias or  requires-element, both of which are independent of summation calc mode.

hermfischer:
Replaced by #680 rebased REC implementation

PR #136, opened 2022-01-31
Fix erroneous catalog.xml validation error in package validation

ChristopheDeBernardi:
#### Reason for Change

When opening / validating a report package that does not contain a single top-level directory (meaning the "META-INF" directory is at the package root) and the "META-INF" directory contains a "catalog.xml" file, Arelle signals (wrongly) an issue with said catalog:

image

(Note the inconsistency between the error code `invalidMetaDataFile` and the actual file catalog.xml)

This problem happened with an actual ESEF report package validated by a national authority and can be reproduced (for instance) with an altered version of the  G2-6-1_1/TC2_invalid.zip test case from the ESEF conformance suite, by removing the top-level directory:

image

In `FileSource.taxonomyPackageMetadataFiles`:

In such cases, because we only check for the "taxonomyPackage.xml" file in subdirectories (marked by the leading '/' in the code below), the actual target file is not found and the `taxonomyPackageMetadataFiles` property instead returns the location of the "catalog.xml" file. Hence the error during package validation.

image

(Side note: I'm not sure which parts of this constitutes the "reason for change" and which belong in the "description" section. I personally would add a "suggested change / fix" section)

#### Description

Changes:

1. Fixing the problem

After the lookup for the "taxonomyPackage.xml" file in subdirectories but before looking for a file in `TAXONOMY_PACKAGE_FILE_NAMES`, looking specifically for the metadata file in a META-INF at the archive root.

So even if the file is not where it is supposed to be, it's better to still find it rather than validating the catalog in its place.

By looking for the META-INF directory at the archive root outside of the `for` loop checking subdirectories, we ensure that the correct one will be used if there are more than one "taxonomyPackage.xml" file (unlikely but who knows).

image

2. Handling a subsequent exception

Loading the test report package after this change has been applied causes the following exception:

image

Because while writing error messages we use `filesource.url` which in this case is a list of a single element: the xhtml file.

In `PackageManager.validateTaxonomyPackage` I replaced all `filesource.url` in logging with `filesource.baseurl` as was already the case for the `invalidDirectoryStructure` message. That way the referenced file is the zip archive rather than the xhtml file, which makes more sense anyway.

#### Steps to Test

Loading the test report package described above gives the following result:

image

* No error message about the "catalog.xml" file
* No unhandled exception raised
* **Note** that when using the ESEF validation plugin as is the case in the screenshot, package validation messages appear twice.

**review**:
@hermfischer-wf

### DIFF
--- a/arelle/FileSource.py
+++ b/arelle/FileSource.py
@@ -342,6 +342,9 @@ class FileSource:
         for f in (self.dir or []):
             if f.endswith("/META-INF/taxonomyPackage.xml"): # must be in a sub directory in the zip
                 return [f]  # standard package
+        # If not found in subdirectory, start by checking if META-INF is at archive top-level (and contains target file)
+        if "META-INF/taxonomyPackage.xml" in self.dir:
+            return ["META-INF/taxonomyPackage.xml"]
         return [f for f in (self.dir or []) if os.path.split(f)[-1] in TAXONOMY_PACKAGE_FILE_NAMES]
     
     def isInArchive(self,filepath, checkExistence=False):
--- a/arelle/PackageManager.py
+++ b/arelle/PackageManager.py
@@ -383,7 +383,7 @@ def validateTaxonomyPackage(cntlr, filesource, packageFiles=[], errors=[]):
         if filesource.isZipBackslashed:
             cntlr.addToLog(_("Taxonomy package directory uses '\\' as fire separator"),
                            messageCode="tpe:invalidArchiveFormat",
-                           file=os.path.basename(filesource.url),
+                           file=os.path.basename(filesource.baseurl),
                            level=logging.ERROR)
             errors.append("tpe:invalidArchiveFormat")
             return False
@@ -396,7 +396,7 @@ def validateTaxonomyPackage(cntlr, filesource, packageFiles=[], errors=[]):
                            messageArgs={"count": len(topLevelFiles),
                                         "topLevelFiles": ', '.join(sorted(topLevelFiles))},
                            messageCode="tpe:invalidDirectoryStructure",
-                           file=os.path.basename(filesource.url),
+                           file=os.path.basename(filesource.baseurl),
                            level=logging.ERROR)
             errors.append("tpe:invalidDirectoryStructure")
         if len(topLevelDirectories) != 1:
@@ -404,7 +404,7 @@ def validateTaxonomyPackage(cntlr, filesource, packageFiles=[], errors=[]):
                            messageArgs={"count": len(topLevelDirectories),
                                         "topLevelDirectories": ', '.join(sorted(topLevelDirectories))},
                            messageCode="tpe:invalidDirectoryStructure",
-                           file=os.path.basename(filesource.url),
+                           file=os.path.basename(filesource.baseurl),
                            level=logging.ERROR)
             if not topLevelFiles:
                 errors.append("tpe:invalidDirectoryStructure")
@@ -421,7 +421,7 @@ def validateTaxonomyPackage(cntlr, filesource, packageFiles=[], errors=[]):
         else:
             cntlr.addToLog(_("Taxonomy package does not contain a metadata file */META-INF/taxonomyPackage.xml"),
                            messageCode="tpe:metadataFileNotFound",
-                           file=os.path.basename(filesource.url),
+                           file=os.path.basename(filesource.baseurl),
                            level=logging.ERROR)
             errors.append("tpe:metadataFileNotFound")
         return len(errors) == numErrorsOnEntry


arelle/FileSource.py
```diff
@@ -342,6 +342,9 @@ def taxonomyPackageMetadataFiles(self):
         for f in (self.dir or []):
```
sagesmith-wf, line 342: Arbitrary line to allow reply for discussion:

> Side note: I'm not sure which parts of this constitutes the "reason for change" and which belong in the "description" section. I personally would add a "suggested change / fix" section

I think you nailed what we had in mind for the PR template at least, but open to suggestions to make it clearer. Our idea with the template is to get a description of what problem you are trying to solve (Reason for change), and then a second description of what you actually did to fix it (Description). I think "suggested change" makes a little more sense in the context of an issue but not as much sense when you have an open PR that's fixing the problem. Fix is similar I think, it implies that an issue/bug is being fixed but wouldn't make as much sense in the context of a new feature or improvement. Anyways I don't feel super strongly ultimately whatever is going to be easiest

sagesmith-wf (approved):
Nice job with the PR description, very clear thank you!

ChristopheDeBernardi:
The same happens if the taxonomyPackage.xml file is not present.  
This can be observed by loading the test case "004-missing-taxonomy-package-file" from the "taxonomy_package_1.0" conformance suite:

image

image

I think the "catalog.xml" should be excluded from possible values returned by `taxonomyPackageMetadataFiles()` since it's never the value that method is expected to provide.

derekgengenbacher-wf:
We appreciate the PRs that you have been opening @ChristopheDeBernardi. We are working towards getting these PRs tested and merged. We appreciate your patience as we work towards accepting commits from the world-wide Arelle community.

arelle/FileSource.py
```diff
@@ -342,6 +342,9 @@ def taxonomyPackageMetadataFiles(self):
...
             if f.endswith("/META-INF/taxonomyPackage.xml"): # must be in a sub directory in the zip
                 return [f]  # standard package
+        # If not found in subdirectory, start by checking if META-INF is at archive top-level (and contains target file)
+        if "META-INF/taxonomyPackage.xml" in self.dir:
```
austinmatherne-wk, line 346: Does removing the leading `/` from the existing `endswith(...)` check have the same effect?

```python
if f.endswith("META-INF/taxonomyPackage.xml")
    return [f]
```

arelle/FileSource.py
austinmatherne-wk: Agreed that suggestion makes more sense for issues. Perhaps the intent for PRs would be clearer if we titled the sections `Reason for change` and `Description of change`?

arelle/FileSource.py
austinmatherne-wk: Also, it's worth mentioning that the taxonomy package spec requires a single top-level directory containing the `META-INF` sub directory.

[Taxonomy Package](https://www.xbrl.org/Specification/taxonomy-package/REC-2016-04-19/taxonomy-package-REC-2016-04-19.html#taxonomy-package)
> A [Taxonomy Package](https://www.xbrl.org/Specification/taxonomy-package/REC-2016-04-19/taxonomy-package-REC-2016-04-19.html#term-taxonomy-package) MUST contain a single top-level directory, with all other files being contained within that directory or descendant subdirectories (tpe:invalidDirectoryStructure). It is recommended, but not required, that the name of the top-level directory be the same as the name of the Taxonomy Package with the .zip file extension removed.
> 
> The top-level directory MUST contain a sub directory named META-INF (tpe:metadataDirectoryNotFound), which in turn MUST contain a taxonomyPackage.xml file (tpe:metadataFileNotFound). The META-INF directory MAY contain a catalog.xml file (see[ Section 3.3](https://www.xbrl.org/Specification/taxonomy-package/REC-2016-04-19/taxonomy-package-REC-2016-04-19.html#sec-remappings)). The META-INF directory MAY also contain other files, but no syntax or semantics are prescribed for such files by this specification.

derekgengenbacher-wf: @ChristopheDeBernardi Can you address @austinmatherne-wk questions, please?

ChristopheDeBernardi: [Sorry I did previously answer but as part as a "review" that I left pending, I'm not really used to writing PR.]

Removing the leading `/` would have the same end result, locating the `taxonomyPackage.xml` file (assuming there is only one of them).  
However preserving the existing loop before searching in other locations allows to:

* Explicitly look only in the expected location at first, then falling back to a "best effort" search (in the unlikely case where there would be more than one `taxonomyPackage.xml` file);
* log a warning or do another specific processing if the file exists but is misplaced.

Regarding the specification, it also says:

> A [Conformant Processor](https://www.xbrl.org/Specification/taxonomy-package/REC-2016-04-19/taxonomy-package-REC-2016-04-19.html#term-conformant-processor) MUST refuse to open or otherwise use a [Taxonomy Package](https://www.xbrl.org/Specification/taxonomy-package/REC-2016-04-19/taxonomy-package-REC-2016-04-19.html#term-taxonomy-package) that violates any of the "MUST" and "MUST NOT" statements in this document defined for Taxonomy Packages (...)

So, indeed, I wonder if looking for the file in alternate places is a good option. My main concern in this PR is that the method ends up returning the `catalog.xml` file if the `taxonomyPackage.xml` is not found.

On the other hand, I suspect that in the case of ESEF reports for instance some authorities would prefer a software to accept opening a package with a misplaced metadata file and see a warning rather than blocking the whole validation process because of that kind of irregularity.  
But that consideration goes beyond what I intended to do with this PR.

PR #138, opened 2022-02-03
Skip XMLSchema.xsd during particle validation

derekgengenbacher-wf:
#### Description:
Exclude validating XMLSchema.xsd during particle validation because it can cause issues when filing docs include extension concepts.

#### Reason:
The FRC 2022 taxonomy makes `http://www.w3.org/2001/XMLSchema.xsd` a discoverable part of the taxonomy. This causes issues in the Arelle xml validator. Because it is implied that everything derives from that schema there is no need to validate it.

#### Testing:
Use the following doc to validate. Validate it with a disclosure system of ESEF.Verify that the `XMLSchema:UnknownElements` error does not get fired with this fix. [UKSEF-2022.zip](https://github.com/Arelle/Arelle/files/7996958/UKSEF-2022.zip)

**review**:
@hermfischer-wf

### DIFF
--- a/arelle/XmlValidateParticles.py
+++ b/arelle/XmlValidateParticles.py
@@ -15,7 +15,7 @@ from arelle.XmlValidate import validate
 
 def validateElementSequence(modelXbrl, compositor, children, ixFacts, iNextChild=0):
     if compositor.modelDocument.targetNamespace == xsd:
-        return None
+        return (iNextChild, True, None, None)
     particles = compositor.dereference().particles        
     iStartingChild = iNextChild
     errDesc = None
--- a/arelle/XmlValidateParticles.py
+++ b/arelle/XmlValidateParticles.py
@@ -9,9 +9,13 @@ from arelle.ModelDtsObject import (ModelConcept, ModelType, ModelGroupDefinition
                                    ModelAll, ModelChoice, ModelSequence, 
                                    ModelAny, anonymousTypeSuffix)
 from arelle.ModelObject import ModelObject, ModelAttribute
+from arelle.XbrlConst import xsd
 from arelle.XmlValidate import validate
 
+
 def validateElementSequence(modelXbrl, compositor, children, ixFacts, iNextChild=0):
+    if compositor.modelDocument.targetNamespace == xsd:
+        return None
     particles = compositor.dereference().particles        
     iStartingChild = iNextChild
     errDesc = None


sagesmith-wf (approved):

arelle/XmlValidateParticles.py
```diff
@@ -11,7 +11,13 @@
 from arelle.ModelObject import ModelObject, ModelAttribute
 from arelle.XmlValidate import validate
 
+excluded_schema = [
+    'XMLSchema.xsd'
+]
+
 def validateElementSequence(modelXbrl, compositor, children, ixFacts, iNextChild=0):
+    if compositor.modelDocument.basename in excluded_schema:
```
brettkail-wk, line 19: This `in` feels too lose, and checking the XSD at all seems a bit too loose.  Can we instead check the namespace for equality?  I would probably reframe it as a check against "builtin" rather than "excluded".

austinmatherne-wk (dismissed):

sagesmith-wf (approved):

arelle/XmlValidateParticles.py
```diff
@@ -11,7 +11,12 @@
 from arelle.ModelObject import ModelObject, ModelAttribute
 from arelle.XmlValidate import validate
 
+BUILTIN_XML_SCHEMA = 'http://www.w3.org/2001/XMLSchema.xsd'
+
+
 def validateElementSequence(modelXbrl, compositor, children, ixFacts, iNextChild=0):
+    if compositor.modelDocument.uri == BUILTIN_XML_SCHEMA:
```
brettkail-wk, line 18: This is still checking the document URL rather than the XSD target namespace.  i.e., it seems like this should be checking `targetNamespace` against `XbrlConst.xsd`.

hermfischer:
+1

sagesmith-wf (approved):

hermfischer-wf:
can we cherry-pick or merge this into the esef-authority-validations branch as well?

austinmatherne-wk (dismissed):

arelle/XmlValidateParticles.py
```diff
@@ -9,9 +9,13 @@
                                    ModelAll, ModelChoice, ModelSequence, 
                                    ModelAny, anonymousTypeSuffix)
 from arelle.ModelObject import ModelObject, ModelAttribute
+from arelle.XbrlConst import xsd
 from arelle.XmlValidate import validate
 
+
 def validateElementSequence(modelXbrl, compositor, children, ixFacts, iNextChild=0):
+    if compositor.modelDocument.targetNamespace == xsd:
+        return (iNextChild, True, None, None)
```
brettkail-wk, line 18: There's no function doc, so I'm not sure exactly what we should be returning, but I wonder if plain/non-tuple `None` would be more accurate since we aren't actually validating anything.  It looks like `XmlValidate` skips if the result is `None`.

brettkail-wk (approved):

brettkail-wk (approved):

sagesmith-wf (approved):

austinmatherne-wk (dismissed):

hermfischer-wf:
+1

arelle/XmlValidateParticles.py
```diff
@@ -9,9 +9,13 @@
                                    ModelAll, ModelChoice, ModelSequence, 
                                    ModelAny, anonymousTypeSuffix)
 from arelle.ModelObject import ModelObject, ModelAttribute
+from arelle.XbrlConst import xsd
 from arelle.XmlValidate import validate
 
+
 def validateElementSequence(modelXbrl, compositor, children, ixFacts, iNextChild=0):
+    if compositor.modelDocument.targetNamespace == xsd:
+        return None
```
sagesmith-wf: We can't return None here because of [this](https://github.com/Arelle/Arelle/blob/ee333d266da4b793e94e4b1e47219f23d38647c5/arelle/XmlValidateParticles.py#L52)

arelle/XmlValidateParticles.py
derekgengenbacher-wf: There is a recursive call [here](https://github.com/Arelle/Arelle/blob/master/arelle/XmlValidateParticles.py#L52) that prohibits returning anything other than a tuple.

sagesmith-wf (changes requested):
QA Failed

sagesmith-wf (approved):

austinmatherne-wk (approved):

arelle/XmlValidateParticles.py
sagesmith-wf: I have tested the new fix and it works as expected

hermfischer-wf:
+1

sagesmith-wf (approved):
QA +10

- Tested locally with attached zipfile

PR #143, opened 2022-02-16
Fix AttributeError due to ModelXbrl passed instead of ValidateXbrl

ChristopheDeBernardi:
#### Reason for Change

Unhandled `AttributeError` exception raised while validating an ESEF report:

image

image

#### Description

In the `validateXbrlFinally.checkLabels()` the method `isExtension(val, modelObject)` is called.  
That method expects a `ValidateXbrl` object but when called inside the `checkLabels()` method it receives instead a `ModelXbrl` instance:

image

#### Steps to reproduce the problem

1. Checking that the disclosure system checks are enabled and that the selected disclosure system is "ESEF"
2. Loading an ESEF report (I used the conformance suite test case "G2-1-2/TC1_valid" but I guess any of these would do)
3. Running the validation of the iXBRL instance

OR in command line, running something like: `-f "[...]/tests/inline_xbrl/G2-1-2/TC1_valid.zip" --plugin "validate/ESEF" --validate --disclosureSystem "esef"`

4. Checking that an `AttributeError` exception is raised.  
See pictures above or in cmd line:

image

#### Steps to Test

1. Going through the "Steps to reproduce the problem" 1 to 3
2. Checking that the exception is not raised:

image

image

#### PS

That was quite a long time spent to describe such a small change.

**review**:
@hermfischer-wf

### DIFF
--- a/arelle/plugin/validate/ESEF/__init__.py
+++ b/arelle/plugin/validate/ESEF/__init__.py
@@ -928,7 +928,7 @@ def validateXbrlFinally(val, *args, **kwargs):
         
         def checkLabels(parent, relSet, labelrole, visited):
             if not parent.label(labelrole,lang=reportXmlLang,fallbackToQname=False):
-                if (not labelrole or labelrole == standardLabel) and isExtension(modelXbrl, parent):
+                if (not labelrole or labelrole == standardLabel) and isExtension(val, parent):
                     missingConceptLabels[labelrole].add(parent)
             visited.add(parent)
             conceptRels = defaultdict(list) # counts for concepts without preferred label role


arelle/plugin/validate/ESEF/__init__.py
```diff
@@ -928,7 +928,7 @@ def checkFootnote(elt, text):
         
         def checkLabels(parent, relSet, labelrole, visited):
             if not parent.label(labelrole,lang=reportXmlLang,fallbackToQname=False):
-                if (not labelrole or labelrole == standardLabel) and isExtension(modelXbrl, parent):
+                if (not labelrole or labelrole == standardLabel) and isExtension(val, parent):
```
sagesmith-wf, line 931: val is passed in all other places 👍

sagesmith-wf (approved):

austinmatherne-wk (approved):

derekgengenbacher-wf (approved):

derekgengenbacher-wf:

PR #145, opened 2022-02-16
Fix ESEF 3-4-3_2 test case

ChristopheDeBernardi:
#### Reason for Change

Arelle gives a false positive for ESEF test case 3-4-3_2/TC1_valid:

image

#### Description

There is a whitespace in the hard-coded value of the default dimension linkrole:

image

This PR has two commits:

1. Removing the incriminated whitespace
2. I wondered if the URI was in the XbrlConst.py file; seems it isn't but I found it in the Const.py file of the plugin, so I thought it best to use that instead of a hard-coded value.

#### Steps to reproduce the problem

1. Checking that the disclosure system checks are enabled and that the selected disclosure system is "ESEF"
2. Loading the ESEF test suite G3-4-3_2/index.xml (or directly the G3-4-3_2/TC1_valid.zip test case)
3. Running the validation

#### Steps to Test

1. Going through the "Steps to reproduce the problem" 1 to 3
2. Checking that the error message is not present for the valid test case

**CAUTION:** when validating the test suite, the valid test case may still fail due to an exception, see pull request 143

#### Side note

I wonder about the relevance of this recently added check anyway. Is it not a duplicate of the `3.4.3.extensionTaxonomyDimensionNotAssignedDefaultMemberInDedicatedPlaceholder` (in DTS.py) ?

image

The message is very similar for the two of them:

image

Plus, the `dimensionDefaultLinkrole` is not in either the Reporting Manual or the Conformance Suite documentation.

image

**review**:
@hermfischer-wf

### DIFF
--- a/arelle/plugin/validate/ESEF/Dimensions.py
+++ b/arelle/plugin/validate/ESEF/Dimensions.py
@@ -165,7 +165,7 @@ def checkFilingDimensions(val):
                             val.modelXbrl.error("ESEF.3.4.3.extensionTaxonomyOverridesDefaultMembers",
                                 _("The extension taxonomy MUST not modify (prohibit and/or override) default members assigned to dimensions by the ESEF taxonomy."),
                                 modelObject=linkChild)
-                    if modelLink.role != "http://www.esma.europa.eu/xbrl/role/cor/ifrs- dim_role-990000":
+                    if modelLink.role != "http://www.esma.europa.eu/xbrl/role/cor/ifrs-dim_role-990000":
                         val.modelXbrl.error("ESEF.3.4.3.dimensionDefaultLinkrole",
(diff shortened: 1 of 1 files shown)


arelle/plugin/validate/ESEF/Dimensions.py
```diff
@@ -165,7 +165,7 @@ def addDomMbrs(sourceDomMbr, ELR, membersSet):
                             val.modelXbrl.error("ESEF.3.4.3.extensionTaxonomyOverridesDefaultMembers",
                                 _("The extension taxonomy MUST not modify (prohibit and/or override) default members assigned to dimensions by the ESEF taxonomy."),
                                 modelObject=linkChild)
-                    if modelLink.role != "http://www.esma.europa.eu/xbrl/role/cor/ifrs- dim_role-990000":
+                    if modelLink.role not in DefaultDimensionLinkroles:
```
sagesmith-wf, line 168: [link](https://github.com/Arelle/Arelle/blob/master/arelle/plugin/validate/ESEF/Const.py#L32) for others ease of review

austinmatherne-wk (approved):

derekgengenbacher-wf (approved):

PR #148, opened 2022-02-18
Fix TypeError: '<' not supported between instances of 'InvalidValue' and 'int'

derekgengenbacher-wf:
#### Description:
Arelle throws a `'<' not supported between instances of 'InvalidValue' and 'int'` error when a monetary fact is tagged with something like `[XX]` when validating with the EFM disclosure. This PR fixes this issue.

#### Testing:
Validate with [type_error_validation.zip](https://github.com/Arelle/Arelle/files/8100007/type_error_validation.zip) and verify that an with `'<' not supported between instances of 'InvalidValue' and 'int'` is not generated.

**review**:
@hermfischer-wf


sagesmith-wf (approved):

austinmatherne-wk (approved):

hermfischer-wf:
+1

Issue #139, opened 2022-02-07
internetConnectivity flag is ignored in 2022-01-31 version


kris2kris:
In version 2021-10-17, I'm using flag internetConnectivity with value offline and I don't have any errors.
With version 2022-01-31, I'm getting the following error when I validate a report 
`[webCache:retrievalError] [Errno -2] Name or service not known retrieving http://www.w3.org/2001/03/xml.xsd`

derekgengenbacher-wf:
@kris2kris Do you have a local cache that your webCache is connected to so that it can find `http://www.w3.org/2001/03/xml.xsd`? I am interested to see if that file exists locally for you when trying to validate?

kris2kris:
I don't think I have it, I create a custom folder for Arelle cache and I only load the following packages 
- eurofiling.info
- xbrl.org
- EBA taxonomies

derekgengenbacher-wf:
It really depends upon the taxonomy that you are trying to validate. Arelle itself does not require specific files. It only references the files as defines by the DTS. Have you started validating a different version of the taxonomies that you usually run with? If so you should be able to do a quick search in the taxonomy files and find where that file association is coming from.

kris2kris:
I run exactly the same taxonomies with the same cache in command line.
Old version of Arelle (2021-10-17) do not produce any error when I am offline, new version of Arelle (2022-01-31) produce an error if I don't put w3.org xsd in my cache

selgamal:
Not directly related to this issue, but thought to report it here, the --internetRecheck option in the command line doesn't seem to have an effect.

derekgengenbacher-wf:
Thanks for pointing that out @selgamal. We will take a look!

derekgengenbacher-wf:
@kris2kris Tracking this down, the 2021-10-17 version had a bug that did not respect the offline mode. So it still in some cases went out and downloaded things. [This](https://github.com/Arelle/Arelle/commit/6ee152be7efe6fb41902f27f625ca713b2051a75) is the fix for that bug. It was released as part of the 2022-01-31 build. Which is why you see a difference.

derekgengenbacher-wf:
@selgamal The internetRecheck thing is broken and we are looking into fixing it.

hermfischer-wf:
See #152 for fix of internetRecheck.

selgamal:
> See #152 for fix of internetRecheck.

Working for me.
[...]

PR #152, opened 2022-02-21
Integrate command line --internetRecheck with WebCache recheck feature

hermfischer-wf:
Reason for Change
The "--internetRecheck" CmdLine option wasn't integrated with the corresponding WebCache feature

Description
Configure option to use feature (specifies how often a request for a web-accessed file is to be checked to see if the web's file has a newer timestamp than the locally-cached file.).

Steps to Test
With debugger one can trace operation at WebCache#360 to verify that when prior check time plus recheck interval passes the http headers for last update time are checked in case http file is newer than cached file's timestamp.

With a web url under tester's control (such as an http-accessed instance doc file), setting to a short recheck interval (quarter-hourly) and touching the web url should cause the file-on-disk in cache to be re-fetched

review:
@derekgengenbacher-wf @sagesmith-wf
cc: @hermfischer-wf

### DIFF
--- a/arelle/CntlrCmdLine.py
+++ b/arelle/CntlrCmdLine.py
@@ -305,8 +305,8 @@ def parseAndRun(args):
     parser.add_option("--internetTimeout", type="int", dest="internetTimeout", 
                       help=_("Specify internet connection timeout in seconds (0 means unlimited)."))
     parser.add_option("--internettimeout", type="int", action="store", dest="internetTimeout", help=SUPPRESS_HELP)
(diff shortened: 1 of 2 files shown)


austinmatherne-wk (approved):

derekgengenbacher-wf:
QA +10 Testing locally succeeded. Verified this using @selgamal suggestions [here](https://github.com/Arelle/Arelle/issues/139#issuecomment-1047420213). Also verified that the headers were being appropriately updated on requests as mentioned above.

PR #154, opened 2022-02-23
Add improved documentation for the README, INSTALLATION and CONTRIBUTION

derekgengenbacher-wf:
#### Description:
The documentation in Arelle needed to be refreshed and updated. As part of this update we are adding clearer instructions on how to contribute to the Arelle project.

#### Testing:
This is a documentation change only and thus does not require testing.

**review**:
@hermfischer-wf

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,4 +1,4 @@
-  <div align="center">
+<div align="center">
   <img src="http://arelle.org/arelle/wp-content/themes/platform/images/logo-platform.png">
 </div>
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,107 @@
+  <div align="center">
+  <img src="http://arelle.org/arelle/wp-content/themes/platform/images/logo-platform.png">
+</div>
+
+# Contributing guidelines
+
+## Reporting issues
+
+Please report issues to the [issue tracker](https://github.com/arelle/arelle/issues).
+
+* Check that the issue has not already been reported.
+* Check that the issue has not already been fixed in the latest code.
+* Be clear and precise (do not prose, but name functions and commands exactly).
+* Include the version of Arelle.
+
+
+## Pull Request Checklist
+
+Before sending your pull requests, make sure you do the following:
+
+*   Read the [contributing guidelines](CONTRIBUTING.md).
+*   Ensure you have signed the [Contributor License Agreement (CLA)](#conttributor-license-agreements).
+*   Check if your changes are consistent with the [guidelines](#general-guidelines-and-philosophy-for-contribution).
+*   Changes are consistent with the [Coding Style](#python-coding-style).
+
+
+## How to become a contributor and submit your own code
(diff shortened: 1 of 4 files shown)


CONTRIBUTING.md
```diff
@@ -0,0 +1,107 @@
+  <div align="center">
```
austinmatherne-wk, line 1: nit: unexpected indention

austinmatherne-wk (approved):

hermfischer-wf (approved):

austinmatherne-wk (approved):

hermfischer-wf (changes requested):
### section: Contributor License Agreements

Regarding: "If you work for a company that wants to allow you to contribute your work..."

Wording on the website: "If you are an individual, not under employee or works for hire agreement (e.g. student), please use the [Individual Contributor](https://arelle.org/arelle/wp-content/uploads/2010/11/ContributorLicenseForIndividuals.txt) agreement, otherwise the [Corporate Contributor ](https://arelle.org/arelle/wp-content/uploads/2010/11/ContributorLicenseForEmployees.txt)agreement. The reason why you need to use the corporate agreement if employed is because employment agreements and employment laws give intellectual property rights to the employer, so it is necessary for the employer to be part of that agreement."

The important issue is that the individual contributor agreement can't be used by somebody under a usual works-for-hire (employee) agreement because those give the employer ownership of all IP, and we don't know the precise wording of each individual's agreement (they usually are non-disclosable)  

Suggest some change is needed here for this point.

### section: General guidelines and philosophy for contribution

before make, I suggest "please", seems petty but...

"When you contribute a new feature or plugin to Arelle" is used in 3 bullets. Suggest putting the three subordinate phrases as sub-bullets under this phrase?

### section: License

I think the CLA licenses Workiva to be the licensor of pulled code.  There shouldn't be newer or overriding license terms from contributions to be cast upon the whole project, scary.

However *plugins* may be independently authored and maintained (EFM, EdgarRenderer, Xule) and we should request that the __pluginInfo__ contain properly displayable license author and copyright.  If a contributor withholds transferring ownership to Workiva (SEC, Xbrl.us) this needs to be discoverable.  If a plugin depends on features copyright by others those copyrights may need to be in __pluginInfo__.

*No viral license*:  No GNU-like licenses can be accepted under the CLA but the contribution may incorporate GNU-like licensed library modules (see plugin internet/proxyNTLM/__init__.py).  In this case the GNU-licensed code cannot be included in core-Arelle, so that an "end user action" is required to cause it's activation.

### section: Installation

I think Windows has to be 10 or 11, macOS (lower case m, no space between mac and OS) needs to be 11 or 12, and only unixes and linuxes compatible with recent Ubuntu or RedHat are supported.

### section: Features

Drop Global Filer Manual? (I don't think anybody uses it nor do I know if it still works). 

table linkbase (Eurofiling version) - we dropped the pre-REC Eurofiling version support

Quickbooks integration hasn't been tested in a long time but it still sounds like good eye-candy.

PR #155, opened 2022-02-25
TR5 implementation

hermfischer:
Reason for Change
* [Transformation Registry 5](https://www.xbrl.org/Specification/inlineXBRL-transformationRegistry/REC-2022-02-16/inlineXBRL-transformationRegistry-REC-2022-02-16.html) (TR5) achieved REC status Feb 16th
* ESEF Guidance 2.2.3 allows filers to use TR5 (as it is now REC)

Description
* TR5 implementation in FunctionIxt.py
* Allowed ESEF TR's.

Steps to Test
* Transforms:
  * GUI: use plugin transforms/tester, consider patterns in [XII tests.xml file](https://gitlab.xbrl.org/base-spec/trr/blob/5.0/conformance-tests/tests.xml).
  * CmdLine: run XII test script for TR5 (scripts/runTR5Tests.sh) which is XII's tests.xml XSLT'ed into test case format by test suite xslt.  For regression, scripts/runTR[2-4]Tests.sh.
* ESEF filing:
  * Substitute TR5 in an ESEF test suite variation or sample filing.

review:
@derekgengenbacher-wf @sagesmith-wf @hermfischer-wf


arelle/FunctionIxt.py
```diff
@@ -1108,7 +1149,6 @@ def numunitdecimalin(arg):
     # same as v2: 'zerodash': zerodash,
 })
     # transformation registry v-4 functions
-tr4Functions = tr3Functions.copy() # tr4 starts with tr3 and adds more functions
```
derekgengenbacher-wf, line 1111: Why this change? Does TR4 not inherit things from TR3 any more?

scripts/runTR2Tests.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
+# Run XBRL Formula Conformance Suite tests
```
derekgengenbacher-wf, line 2: Arbitrary line: NICE!! I was looking for these scripts!

arelle/FunctionIxt.py
hermfischer-wf: Never did, that was a goof.

brettkail-wk (approved):

derekgengenbacher-wf (approved):

austinmatherne-wk (approved):

sagesmith-wf (approved):

sagesmith-wf (approved):
QA +10

- Followed QA testing steps

PR #158, opened 2022-02-26
Allow creation of FileSource from ZIP files without extension

ChristopheDeBernardi:
### Reason for Change

Trying to open a ZIP file without a file extension in Arelle is currently not possible:

image

### Justification for Change

1. Some XBRL actors (such as national authorities) choose to rename report packages when they file them into a documents management system; some also choose to strip the file extension, making subsequent exploitation of zip packages in Arelle impossible.
2. Some XBRL files (taxonomy and report packages) are required to conform to the ZIP format; however said format does not prescribe a specific extension, nor does the Taxonomy Package 1.0 specification (at least not explicitly, to my understanding):  

image
image

3. Relying solely on the file extension is not the most robust or secure approach anyway.
4. The [Public Working Draft for Report Package 1.0](https://www.xbrl.org/Specification/report-package/PWD-2020-12-09/report-package-PWD-2020-12-09.html) explores the possibility of using custom file extensions for report packages in the future.
5. Even if refusing to open a file without an extension is the desired behavior, the error message is not exactly helpful to understand the problem.

### Description

This pull request adds a `_determineType()` method to the `FileSource` class to encapsulate the file type detection.  
That method currently alters only the detection of ZIP files, using `zipfile.is_zipfile()`, but checks for other file formats could be added if needed.

It the file is not a ZIP, its type is determined based on its extension as before.

To change as little code as possible, the method returns a value in a form expected in the `FileSource.type` field, meaning the standard file extension: `.zip` for zip files, or whatever the actual file extension is for others.

### Side effects

This change affects (at least) the test results for the ESEF conformance suite 2021, `RTS_Annex_III_Par_3_G3-1-3` (about report packages formats), making the following cases fail:

1. `TC3_invalid.rar`

It looks like the test file is just a ZIP file with a `.rar` extension, not an actual RAR file?

image

2. `TC5_invalid.jar`

An update of the ESEF validation plugin might be needed if the `.zip` extension is actually mandatory for ESEF, which does not seem to be the case according to the reporting manual:

image

### Steps to Test

Try to load a (valid) zip file without its extension (for instance a test case from the ESEF conformance suite)...

image

... and it should work fine:

image

---

I also wrote a unit test to check the `FileSource.isZip` (and other) values (in a limited way, my main concern was the ZIP files):

~~~python
import json

from pathlib import Path
from unittest import TestCase

from arelle.FileSource import FileSource


class FileSourceTypeTest(TestCase):

    samplesDir: Path = [Wherever you stored the test files]
    testData = {}

    @classmethod
    def setUpClass(cls) -> None:
        with (cls.samplesDir / 'index.json').open('r') as index:
            cls.testData = json.load(index)

    def test(self):
        for testCase in self.testData:
            with self.subTest(f'FileSource type detection - {testCase}: fail'):
                sampleFile = self.testData[testCase]
                url = self._getURL(testCase)
                fileSource = FileSource(url)
                assertions = sampleFile['assertions']
                for attribute in assertions:
                    with self.subTest(f'FileSource type detection - {testCase} - {attribute}: fail'):
                        if not hasattr(fileSource, attribute):
                            self.fail(f'FileSource object has no attribute {attribute}')
                        self.assertEqual(getattr(fileSource, attribute), assertions[attribute], msg=f'{testCase} - {attribute}: fail ({getattr(fileSource, attribute)})')
                        print(f'FileSource type detection - {testCase} - {attribute}: pass ({getattr(fileSource, attribute)})')

    def _getURL(self, testCase: str) -> Path:
        url = self.testData[testCase]['filename']
        if not testCase.lower().startswith('rss'):
            url = self.samplesDir / url
        return url
~~~

With the `index.json` file looking like this (actual test sample files sold separately):

~~~json
{
    "ZIP": {
        "filename": "zip.zip",
        "assertions": {
            "isRss": false,
            "isTarGz": false,
            "isZip": true
        }
    },
    "ZIP without extension": {
        "filename": "zip",
        "assertions": {
            "isRss": false,
            "isTarGz": false,
            "isZip": true
        }
    },
    "TarGz": {
        "filename": "targz.tar.gz",
        "assertions": {
            "isRss": false,
            "isTarGz": true,
            "isZip": false
        }
    },
    "TarGz without extension": {
        "filename": "targz",
        "assertions": {
            "isRss": false,
            "isTarGz": false,
            "isZip": false
        }
    },
    "RSS": {
        "filename": "http://www.sec.gov/Archives/edgar/usgaap.rss.xml",
        "assertions": {
            "isRss": true,
            "isTarGz": false,
            "isZip": false
        }
    },
    "7zip": {
        "filename": "7z.7z",
        "assertions": {
            "isRss": false,
            "isTarGz": false,
            "isZip": false
        }
    },
    "7zip without extension": {
        "filename": "7z",
        "assertions": {
            "isRss": false,
            "isTarGz": false,
            "isZip": false
        }
    }
}
~~~

image

**review**:
@hermfischer-wf

### DIFF
--- a/arelle/FileSource.py
+++ b/arelle/FileSource.py
@@ -107,10 +107,8 @@ class FileSource:
         self.url = str(url)  # allow either string or FileNamedStringIO
         self.baseIsHttp = isHttpUrl(self.url)
         self.cntlr = cntlr
-        self.type = self.url.lower()[-7:]
+        self.type = self._determineType()
         self.isTarGz = self.type == ".tar.gz"
-        if not self.isTarGz:
-            self.type = self.type[3:]
         self.isZip = self.type == ".zip"
         self.isZipBackslashed = False # windows style backslashed paths
         self.isEis = self.type == ".eis"
@@ -141,7 +139,31 @@ class FileSource:
                     if self.cntlr:
                         self.cntlr.addToLog(_("[{0}] {1}").format(type(err).__name__, err))
                     pass
-            
+
+    def _determineType(self):
+        """
+        Attempt to detect the format of the file located at self.url.
+
+        Try to use methods dedicated to some formats such as `zipfile.is_zipfile`.
+        If that fails, use the historical algorithm using the file extension.
+
+        :return: a string corresponding to the usual file extension for the file format:
+                 ".zip" for a ZIP archive, ".tar.gz" for a tgz archive, etc.
+        """
+        filetype = None
+        # Extension-independent file type detection
+        try:
+            if zipfile.is_zipfile(self.url):
+                filetype = '.zip'
+        except Exception:  # if the value of self.url is incompatible with the methods used
+            pass
+        # If no advanced methods gave a result, fall back to file extension
+        if filetype is None:
+            filetype = self.url.lower()[-7:]
+            if not filetype == ".tar.gz":
+                filetype = filetype[3:]
+        return filetype
+
     def logError(self, err):
         if self.cntlr:
             self.cntlr.addToLog(_("[{0}] {1}").format(type(err).__name__, err))


arelle/FileSource.py
```diff
@@ -141,7 +139,31 @@ def __init__(self, url, cntlr=None, checkIfXmlIsEis=False):
...
+    def _determineType(self):
+        """
+        Attempt to detect the format of the file located at self.url.
+
+        Try to use methods dedicated to some formats such as `zipfile.is_zipfile`.
+        If that fails, use the historical algorithm using the file extension.
+
+        :return: a string corresponding to the usual file extension for the file format:
+                 ".zip" for a ZIP archive, ".tar.gz" for a tgz archive, etc.
+        """
```
derekgengenbacher-wf, line 152: Thank you for the docstring!!!

arelle/FileSource.py
```diff
@@ -141,7 +139,31 @@ def __init__(self, url, cntlr=None, checkIfXmlIsEis=False):
...
+        filetype = None
+        # Extension-independent file type detection
+        try:
+            if zipfile.is_zipfile(self.url):
+                filetype = '.zip'
+        except Exception:  # if the value of self.url is incompatible with the methods used
```
derekgengenbacher-wf, line 158: What would an incompatible url look like? Can we catch that prior to this so that we dont require a try/except? If we cant catch a bad url, then can we at least be more specific to what we are catching here?

brettkail-wk: It seems like ideally, this checking would somehow be deferred until `.open` when `basefile` is known.

ChristopheDeBernardi: > What would an incompatible url look like?

I don't know, that why I added the try/except, but I was thinking of web URLs or network location.  
I didn't test those, that was a little lazy of me.

It turn out that `http://...`  URLs or something like `\\localserver\share\...` do not raise an exception but return `False` even if the target is a ZIP file. So not the best solution although the method as a whole would still give the right result based on the file extension.

> It seems like ideally, this checking would somehow be deferred until `.open` when `basefile` is known.

Maybe but the `isZip` attribute and others are initialized and used in the `__init__`, and it's technically possible to instantiate a `FileSource` to exploit such attribute before or without actually opening it.  
Maybe checking again the type in `open()` with `basefile` instead of `url`?

brettkail-wk: > Maybe checking again the type in open() with basefile instead of url?

It seems wrong to speculatively do I/O in the constructor for a path that we don't know is valid.  I'm not so familiar with all the ways FileSource is used, but does it work to leave the logic in the constructor as-is (unknown type and pessimistic isZip=False) and update those vars later in `open`?  That seems to avoid the need for try/except that Derek mentions, seems to fix the .rar issue from the PR description (regardless of testsuite being more strict than the manual...), and seems to fix the motivating use-case of a file without an extension.

hermfischer:
Can we use libmagic to determine file type (not only here but also in EFM and ESEF to match image types to their declared extension or mime)?  (It has MIT license so ok)

Probably SEC EIS files still detect by their extension.

7z and rar by py7xr but must be plugin due to GPL license (other py 7z with non-GPL license)

brettkail-wk:
@hermfischer For zip files in particular, `zipfile.is_zipfile` seems correct: if the libmagic (or whatever) says something is a zip file but `zipfile` can't open it, then things will be broken anyway, so we might as well ask the Python library we're using.

> also in EFM and ESEF to match image types to their declared extension

It seems somewhat odd for validation behavior to depend on the presence/contents of /etc/magic (etc.) on the OS of the machine running Arelle.  It seems like matching on specific magic bytes is preferable.

arelle/FileSource.py
```diff
@@ -107,10 +107,8 @@ def __init__(self, url, cntlr=None, checkIfXmlIsEis=False):
         self.url = str(url)  # allow either string or FileNamedStringIO
         self.baseIsHttp = isHttpUrl(self.url)
         self.cntlr = cntlr
-        self.type = self.url.lower()[-7:]
+        self.type = self._determineType()
         self.isTarGz = self.type == ".tar.gz"
```
austinmatherne-wk, line 111: Can we use [`tarfile.is_tarfile()`](https://docs.python.org/3/library/tarfile.html#tarfile.is_tarfile)?

ChristopheDeBernardi: I added a commit to use that method.

ChristopheDeBernardi:
> Can we use libmagic to determine file type (not only here but also in EFM and ESEF to match image types to their declared extension or mime)?

To deal with file type detection I use [python-magic](https://pypi.org/project/python-magic/).  
It exists a variant [python-magic-bin](https://pypi.org/project/python-magic-bin/) which comes with binary packages as to avoid relying on system dependencies.

However I came across two issues regarding xlsx files (in the context of the "Load from Excel" plugin for instance):

1. Such files are actually ZIP archives, therefore they would not behave as expected if file detection relied only on `zipfile.is_zipfile()`.  
Using python-magic, I could solve that problem with something like this:

~~~python
if zipfile.is_zipfile(self.url):
    mimetype = magic.from_file(self.url, mime=True)
    if mimetype == 'application/zip':
        filetype = '.zip'
~~~

image

2. xlsx files generated by `openpyxl` (in the "Save loadable Excel" plugin) are not correctly identified, either by python-magic or by the `file` Linux command:

image

image

PR #165, opened 2022-03-09
Modernize ESEF conformance suite scripts and add integration tests

derekgengenbacher-wf:
#### Reason for change
In an attempt to modernize Arelle and provide easier to use testing capabilities, I have modernized the ESEF conformance suite scripts so that anyone with a python environment can easily run, verify and test ESEF code. 

Additionally I have setup the foundation of integration testing that can be used to run the ESEF conformance suite for a potential CI feature in the future.

#### Steps to Test
- Run `python -m scripts.conformance_suites.run_esef_conformance_suite` from the main Arelle directory to run the standard script and verify successful completion. It should produce a log file and an xlsx of the results.
- Run `pytest --disable-warnings tests` from the main Arelle directory to run the integration tests and verify successful completion. These tests will not pass at the moment. More work will be done to make sure that Arelle conforms 100% to the 2021 conformance suite.

**review**:
@Arelle/arelle


tests/integration_tests/validation/validation_util.py
```diff
@@ -0,0 +1,37 @@
...
+    Produces a list of Pytest Params that can be fed into a parameterized pytest function
+
+    :param args: The args to be parsed by arelle in order to correctly produce the desired result set
+    :type args: list of strings
+    :return: A list of PyTest Params that can be used to run a parameterized pytest function
+    :rtype: list of ::class:: `~pytest.param`
+    """
+    cntlr = parseAndRun(args)
+    results = []
+    model_document = cntlr.modelManager.modelXbrl.modelDocument
```
sagesmith-wf, line 19: Can we use a `with` block with this, not entirely sure?

tests/integration_tests/validation/validation_util.py
```diff
@@ -0,0 +1,37 @@
...
+
+from arelle.CntlrCmdLine import parseAndRun
+from arelle import ModelDocument
+
+
+def get_test_data(args):
+    """
+    Produces a list of Pytest Params that can be fed into a parameterized pytest function
+
+    :param args: The args to be parsed by arelle in order to correctly produce the desired result set
```
sagesmith-wf, line 12: Not sure if this is conventional within the larger python community. I did some looking and I am not sure they have a standard so I would be inclined to do whatever the ide auto generates

arelle/Validate.py
```diff
@@ -566,6 +566,7 @@ def determineTestStatus(self, modelTestcaseVariation, errors):
         elif isinstance(expected,(QName,_STR_BASE,dict,list)): # string or assertion id counts dict
```
sagesmith-wf, line 566: Random Line:

Better check that the tests and data don't get packaged with the releases

tests/integration_tests/validation/validation_util.py
```diff
@@ -0,0 +1,37 @@
...
+    :return: A list of PyTest Params that can be used to run a parameterized pytest function
+    :rtype: list of ::class:: `~pytest.param`
+    """
+    cntlr = parseAndRun(args)
+    results = []
+    model_document = cntlr.modelManager.modelXbrl.modelDocument
+    if model_document is not None:
+        if model_document.type == ModelDocument.Type.TESTCASESINDEX:
+            for tc in sorted(model_document.referencesDocument.keys(), key=lambda doc: doc.uri):
+                uri_dir_parts = os.path.dirname(tc.uri).split('/')
```
sagesmith-wf, line 23: Wondering if we can clean this path resolution up a bit. I think you are using strings here instead of internal python file utilities

sagesmith-wf (approved):

austinmatherne-wk (approved):

arelle/Validate.py
derekgengenbacher-wf: They most certainly do. Will have to fix that once we get to the building point.

arelle/plugin/validate/ESEF/__init__.py
```diff
@@ -1105,7 +1105,7 @@ def validateFormulaFinished(val, *args, **kwargs): # runs *after* formula (which
 
     modelXbrl = val.modelXbrl
     # reset environment formula run IDs
-    modelXbrl.modelManager.formulaOptions.runIDs = val.priorFormulaOptionsRunIDs
+    modelXbrl.modelManager.formulaOptions.runIDs = getattr(val, 'priorFormulaOptionsRunIDs', None)
```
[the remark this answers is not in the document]
derekgengenbacher-wf: In the conformance suite case

tests/integration_tests/validation/validation_util.py
derekgengenbacher-wf: I thought about this and checked to see if Cntlr has the necessary functions to use in a with block. It does not. This can be a future update.

tests/integration_tests/validation/validation_util.py
derekgengenbacher-wf: What do you have in mind here?

hermfischer-wf:
The Run `python -m scripts.conformance_suites.run_esef_conformance_suite` may need to be more like this:
`PYTHONPATH=$PWD full-path-to-python3.9 -m scripts.conformance_suites.run_esef_conformance_suite`
for anybody with lots of projects, or multiple pythons (like anybody with Mac M1's)

hermfischer-wf:
The produced log files don't end in `.log` but in `-log.txt` (update "Steps to test").

hermfischer-wf (approved):

hermfischer-wf:
Please add this code change so that formula doesn't run on the testcase modelXbrl object and so that the validateFormulaFinished callback doesn't reset formulaOptions.runIDs when they were not saved:
CntlrCmdLine.py#1011:
```
                    if (modelXbrl.modelDocument.type not in ModelDocument.Type.TESTCASETYPES and
                        options.formulaAction in ("validate", "run") and  # do nothing here if "none"
                        not isAlreadyValidated):  # formulas can't run if streaming has validated the instance 
```
validate/ESEF/_ _init_ _.py#1107-1108:
```
    if hasattr(val, 'priorFormulaOptionsRunIDs'): # reset environment formula run IDs if they were saved
        modelXbrl.modelManager.formulaOptions.runIDs = val.priorFormulaOptionsRunIDs
```

austinmatherne-wk (approved):

hermfischer-wf (approved):

austinmatherne-wk:
# Verified

* verified tests locally

PR #170, opened 2022-03-25
validate/FERC add inconsistent duplicate check

hermfischer-wf:
#### Reason for change
FERC lacked python-implemented inconsistent duplicates detection, in a form suitable to be run before processing workflow commits to starting xule processing.

#### Description of change
Adapt inconsistent dup detection coding from validate/HMRC.  Didn't use validate/EFM because EFM relies on units and contexts being unique whereas HMRC allows multiple equivalent units and contexts.  (For a rainy day: refactor all the plugins which implement quite similar versions of this: validate/EFM, validate/HMRC, validate/ROS and loadFromOIM.)

[Method of inconsistent dup detection is specified here](http://www.xbrl.org/WGN/xbrl-duplicates/WGN-2018-04-19/xbrl-duplicates-WGN-2018-04-19.html)

#### Steps to Test
Create test cases.  Check these sort of issues:
1) numeric inconsistencies and consistencies within interval bounds
2) string inconsistencies which honor language codes (e.g. lang codes are checked to each other)
3) other data types

**review**:
@Arelle/arelle

### DIFF
--- a/arelle/plugin/validate/FERC/__init__.py
+++ b/arelle/plugin/validate/FERC/__init__.py
@@ -11,10 +11,13 @@ Filer Guidelines:
 (c) Copyright 2021 Mark V Systems Limited, All rights reserved.
 '''
 import os, re
+from collections import defaultdict
+from math import isnan
 from arelle import ModelDocument, ValidateFilingText, XmlUtil
 from arelle.ModelInstanceObject import ModelFact, ModelInlineFact, ModelInlineFootnote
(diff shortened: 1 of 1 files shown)


austinmatherne-wk (approved):

derekgengenbacher-wf:
Based on a discussion with the Arelle team we have decided not to accept this PR at this time. This PR will duplicate the duplicate fact errors for users who are validating with the FERC/XULE combo.

PR #178, opened 2022-04-19
lxml schema validation resolve when in archive

hermfischer-wf:
#### Reason for change
Conf suite in an archive may have schema files needed by test case files within the archive, add capability for lxml to resolve from achive contents.

#### Description of change
Add modelXbrl as optional parameter to lxmlResolvingParser constructor, when present resolver checks whether requested url is present within the archive (via its fileSource).

#### Steps to Test
Zipped ESEF test suite fail to load a schema file before this change and sails smoothly with the change.

**review**:
@Arelle/arelle

### DIFF
--- a/arelle/XmlValidate.py
+++ b/arelle/XmlValidate.py
@@ -599,16 +599,24 @@ def validateAnyWildcard(qnElt, qnAttr, attributeWildcards):
     return False
 
 class lxmlSchemaResolver(etree.Resolver):
-    def __init__(self, cntlr):
+    def __init__(self, cntlr, modelXbrl=None):
         super(lxmlSchemaResolver, self).__init__()
         self.cntlr = cntlr
+        self.modelXbrl = modelXbrl
     def resolve(self, url, id, context): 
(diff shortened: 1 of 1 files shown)


derekgengenbacher-wf (approved):

sagesmith-wf (approved):

derekgengenbacher-wf:
Tested against the EFM Conformance Suite, but still found that loading xsds was not correct. 

```
[IOerror] dei-2021.xsd: file error: [Errno 2] Archive arelle/tests/resources/conformance_suites/efm_conformance_suite_2022.zip/conf/testcases.xml: 'xbrl.sec.gov/dei/2021/dei-2021.xsd' - i00200gd-20081231_pre.xml
```

This error was generated by the following test `525-ix-syntax/efm/00-filing/00-filing-efm-testcase.xml`

derekgengenbacher-wf:
QA +10 Testing locally succeeded. Tested locally running with the EFM conformance suite. All files correctly load from the zip as expected after this update.

Issue #176, opened 2022-04-13
Formula messages using ESEF Taxonomy 2021


sachakaelin:
While making myself familiar with the ESEF Taxonomy 2021 I noticed a difference in how the ESEF validation in Arelle (2022-04-01)  handels formulae messages in comparison with the ESEF Taxonomy 2020. 

When validating test RTS_Annex_II_Par_2\TC2_invalid.zip (conformance suite 2020) using "ESMA RTS on ESEF" the validation messages show:
```
[ESEF.RTS.Annex.II.Par.2.missingMandatoryMarkups] Mandatory elements to be marked up are missing: LegalFormOfEntity, NameOfReportingEntityOrOtherMeansOfIdentification. - abc-2019-12-31.xhtml 
[info:profileActivity] validating ESMA RTS on ESEF filing rules 0.033 secs
 - abc-2019-12-31.xhtml 
[message:man_LegalFormOfEntity] missingMandatoryMarkups: According to the Regulatory Technical Standards on European Single Electronic Format, the element "Legal form of entity" is part of the mandatory list of tags that must be applied if corresponding information is present in a report. Please make sure that you applied this tag in your report. - http://www.esma.europa.eu/taxonomy/2020-03-16/esef_cor-for.xml 20523
[message:man_NameOfReportingEntityOrOtherMeansOfIdentification] missingMandatoryMarkups: According to the Regulatory Technical Standards on European Single Electronic Format, the element "Name of reporting entity or other means of identification" is part of the mandatory list of tags that must be applied if corresponding information is present in a report. Please make sure that you applied this tag in your report. - http://www.esma.europa.eu/taxonomy/2020-03-16/esef_cor-for.xml 20491
[ESEF.2.7.1.targetXBRLDocumentWithFormulaErrors] Target XBRL document MUST be valid against the assertions specified in ESEF taxonomy, 1 with errors. - abc-2019-12-31.xhtml 
[ESEF.2.7.1.targetXBRLDocumentWithFormulaWarnings] Target XBRL document SHOULD be valid against the assertions specified in ESEF taxonomy, 1 with warnings. - abc-2019-12-31.xhtml 
```

In comparison, performing the same test using ESEF conformance Suite 2021 results in:
```
[ESEF.RTS.Annex.II.Par.2.missingMandatoryMarkups] Mandatory elements to be marked up are missing: LegalFormOfEntity, NameOfReportingEntityOrOtherMeansOfIdentification. - abc-2019-12-31.xhtml 
[...]

PR #184, opened 2022-04-29
Fix regression in validate.determineTestStatus

derekgengenbacher-wf:
#### Reason for change
A bug was introduced in this [commit](https://github.com/Arelle/Arelle/commit/f3072b8ef519118c73e7b7e2a11b6c2e71657218). This code is wrong because it does not take into account all of the different structures that can potentially be run through this function. Most notably it does not handle dicts being passed in from formula. This code wrong altogether and was thought to handle a scenario that does not happen.

#### Description of change
Revert the commit mentioned above.

#### Steps to Test
- Run the ESEF conformance suite. All but one test should pass. 
- Run the formula conformance suite. All but one test should pass.

#176 

**review**:
@Arelle/arelle

### DIFF
--- a/arelle/Validate.py
+++ b/arelle/Validate.py
@@ -566,7 +566,6 @@ class Validate:
         elif isinstance(expected,(QName,_STR_BASE,dict,list)): # string or assertion id counts dict
             status = "fail"
             _passCount = 0
-            matched_expected = []
             if isinstance(expected, list):
                 _expectedList = expected.copy() # need shallow copy
             else:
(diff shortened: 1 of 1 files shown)


stevenbronson-wk (approved):

austinmatherne-wk (approved):

sagesmith-wf (approved):

hermfischer-wf (approved):

stevenbronson-wk:
Ran the ESEF and formula conformance suites and one test failed in each.

Issue #189, opened 2022-05-12
Arelle installation via pypi


http203:
Hello, I noticed that some old Arelle versions are available on **pypi** (https://pypi.org/project/arelle).

Is there any plan to publish Arelle updates on that service?

Thank you!

sagesmith-wf:
@http203, short answer, yes :). We are working on a CI story for this project that you will hopefully see soon that will include a pypi package. Its complicated by the large conformance suits and the multiple OS distributions but will hopefully be available for you and others soon

selgamal:
> @http203, short answer, yes :). We are working on a CI story for this project that you will hopefully see soon that will include a pypi package. Its complicated by the large conformance suits and the multiple OS distributions but will hopefully be available for you and others soon

This is great news!

I have a suggestion, how about a sort of an "interface" that enables access to the installed arelle application from python.

I am implementing something similar for my personal use, I wrote a script that figures out the locations of arelle folders (config, resources...) and adds the necessary entries to python path, when I import that script I have access to import arelle objects in my code from the arelle application already installed on my machine or from source if I point the script to my local project folder (of course all dependencies need to be  installed in my environment for this to work).

The benefit is that I am running from the same arelle version whether I am importing in python or running the app.
 
I think this will be a useful feature to add if it makes sense.

sagesmith-wf:
@selgamal I'm not really sure of the use case for what you are describing. I'm assuming its maybe getting around making the dev environment work properly on your host machine? If that's the use case, we hope to make documentation that helps make that easy for users to avoid the want to use the suggested script. We are hoping to not be a redistribution for TKinter/Python etc. so would appreciate some feedback (perhaps on another issue) on the parts of the dev setup that make it desirable to directly edit the installed application

selgamal:
@sagesmith-wf  yes you are correct I didn't think that through.
Anyway what I understood from your first comment is that you are planning to release a pypi package, if this is the case I will hold this thought until the package is available and make a demo explaining what I meant.

PR #192, opened 2022-05-17
Document dev environment setup

sagesmith-wf:
#### Reason for change
Per [conversation](https://github.com/Arelle/Arelle/issues/189#issuecomment-1125217420) with @selgamal, the python setup was undocumented

#### Description of change
Documents how to setup arelle for development

#### Steps to Test
Follow the instructions

**review**:
@Arelle/arelle
@selgamal

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -41,6 +41,15 @@ return it. Once we receive it, we'll be able to accept your pull requests.
 
 ***NOTE***: Only original source code from you and other people that have signed the CLA can be accepted into the main repository.
 
+### Setting up an environment
+1. Install [pyenv](https://github.com/pyenv/pyenv#installation)
+2. Install a [supported version of Python](https://github.com/Arelle/Arelle/blob/master/INSTALLATION.md). For example, `pyenv install 3.9.9`
+3. Create a virtual env using the minimum python version. For example, `PYENV_VERSION=3.9.9 pyenv exec python -m venv venv`
+4. Activate your environment `source venv/bin/activate`
+5. Install dependencies `pip install -r requirements-dev.txt`
+6. Verify you can run the app
+   1. GUI `python arelleGUI.pyw`
+   2. CLI `python arelleCmdLine.py`
 
 ### Contributing code
 
--- a/.gitignore
+++ b/.gitignore
@@ -14,3 +14,4 @@ Arelle.egg-info
 .eggs
 *.iml
 .idea
+/venv/


austinmatherne-wk (approved):

CONTRIBUTING.md
```diff
@@ -41,6 +41,16 @@ return it. Once we receive it, we'll be able to accept your pull requests.
 
 ***NOTE***: Only original source code from you and other people that have signed the CLA can be accepted into the main repository.
 
+### Setting up an environment
+1. Install [pyenv](https://github.com/pyenv/pyenv#installation)
+2. Install python 3.9 `pyenv install 3.9`
```
brettkail-wk, line 46: This doesn't seem to work as-is:
```
$ pyenv install 3.9
python-build: definition not found: 3.9

The following versions contain `3.9' in the name:
...
```

CONTRIBUTING.md
```diff
@@ -41,6 +41,16 @@ return it. Once we receive it, we'll be able to accept your pull requests.
...
+3. Locate the python installation ex: `~/.pyenv/versions/3.9.9/bin/python3.9`
+4. Create a virtual environment `<PATH_TO_PYTHON> -m venv venv`
```
brettkail-wk, line 48: Perhaps steps 3/4 could be improved with `pyenv global 3.9.9` (or whatever version from previous step), and then just plain `python -mvenv venv` rather than needing to use the `~/.pyenv` path.

(I'm not so familiar with `pyenv` and how it interacts with `venv`, but perhaps `pyenv local 3.9.9` in the Arelle checkout is cleaner than `pyenv global`.  If so, we could perhaps add `.pyenv-version` to `.gitignore` so the marker doesn't accidentally get checked in.)

stevenbronson-wk: Could also use `PYENV_VERSION=3.9.9 pyenv exec python -m venv venv`.

brettkail-wk: Yes, I think that's a great suggestion, thanks.

CONTRIBUTING.md
```diff
@@ -41,6 +41,14 @@ return it. Once we receive it, we'll be able to accept your pull requests.
 
 ***NOTE***: Only original source code from you and other people that have signed the CLA can be accepted into the main repository.
 
+### Setting up an environment
+1. Install [pyenv](https://github.com/pyenv/pyenv#installation)
+2. Create a 3.9.9 virtual env `PYENV_VERSION=3.9.9 pyenv exec python -m venv venv`
```
brettkail-wk, line 46: This reads as if it is mandating 3.9.9, but I think it's just an example.  Do we somewhere specify the minimum Python version?  Perhaps "Create a virtual environment.  For example, `...`"?

You removed the `pyenv install`, which is strange since you left the "Install pyenv" step.  Perhaps "Install the minimum Python version.  For example, `pyenv install 3.9.9`"?

I believe this will create a `venv` directory.  Should we add `venv/` to `.gitignore`?

brettkail-wk (approved):

derekgengenbacher-wf (approved):

brettkail-wk (dismissed):

stevenbronson-wk (approved):

austinmatherne-wk (approved):

PR #198, opened 2022-05-20
Run tests on release

derekgengenbacher-wf:
#### Reason for change
Tests currently are setup to run on PR creation. We need to have them also run on release creation.

#### Description of change
Run tests on release creation

#### Steps to Test
CI

**review**:
@Arelle/arelle

### DIFF
--- a/.github/workflows/conformance-suites.yml
+++ b/.github/workflows/conformance-suites.yml
@@ -1,5 +1,7 @@
 name: Run Xbrl Conformance Suites
 on: pull_request_target
+  release:
+    types: [created]
 
 jobs:
   run-conformance-suite:
--- a/.github/workflows/conformance-suites.yml
+++ b/.github/workflows/conformance-suites.yml
@@ -1,5 +1,6 @@
 name: Run Xbrl Conformance Suites
-on: pull_request_target
+on:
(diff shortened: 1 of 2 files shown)


aaroncameron-wk (approved):

.github/workflows/conformance-suites.yml
```diff
@@ -1,5 +1,7 @@
 name: Run Xbrl Conformance Suites
 on: pull_request_target
+  release:
```
stevenbronson-wk, line 3: I think this is a syntax error.

aaroncameron-wk (approved):

.github/workflows/conformance-suites.yml
```diff
@@ -1,5 +1,8 @@
 name: Run Xbrl Conformance Suites
-on: pull_request_target
+on:
+  pull_request: pull_request_target
```
stevenbronson-wk, line 3: I don't think this means "pull_request_target".  Maybe
```
on:
  pull_request_target:
  release:
    types: [created]
```

austinmatherne-wk: It would be best to test this against a fork before merging

Issue #203, opened 2022-06-09
rdflib required in version 5.0.0


mo-klein:
During the setup, I have discovered that the library _rdflib_ is required to be installed in the version 5.0.0 because of some namespace changes in the later versions.

Cheers.

austinmatherne-wk:
Thanks for reporting Moritz. We'll get this addressed soon.

PR #204, opened 2022-06-10
Move cx freeze to dev requirements

austinmatherne-wk:
#### Reason for change
As mentioned in #203 the XbrlSemanticRdfDB plugin doesn't work with newer versions of rdflib (and will cause cx_Freeze builds to fail). Additionally, the cx_Freeze plugin shouldn't be required for users who simply wish to run the service directly.

#### Description of change
* Makes cx_Freeze a dev dependency
* Restricts rdflib to compatible versions

#### Steps to Test
* CI/Conformance suites

**review**:
@Arelle/arelle

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -1,3 +1,4 @@
+cx_Freeze==6.11.0
 mock==4.0.3
 pytest==7.1.1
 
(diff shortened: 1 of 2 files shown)


derekgengenbacher-wf (approved):

stevenbronson-wk (approved):

derekgengenbacher-wf:
QA +10 CI passes

PR #207, opened 2022-06-14
Discover integration tests.

stevenbronson-wk:
#### Reason for change
Integration tests introduced in PRs don't run until merged.

#### Steps to Test
You could try adding a PR on a fork after incorporating this change.

**review**:
@Arelle/arelle

### DIFF
--- a/.github/workflows/conformance-suites.yml
+++ b/.github/workflows/conformance-suites.yml
@@ -2,18 +2,29 @@ name: Run Xbrl Conformance Suites
 on: pull_request_target
 
 jobs:
+  find-tests:
+    runs-on: ubuntu-latest
+    outputs:
+      matrix: ${{ steps.build-test-matrix.outputs.matrix }}
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          persist-credentials: false
+          ref: ${{ github.event.pull_request.head.sha }}
+          repository: ${{ github.event.pull_request.head.repo.full_name }}
+      - id: build-test-matrix
+        run: |
+          printf '::set-output name=matrix::'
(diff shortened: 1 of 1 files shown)


.github/workflows/conformance-suites.yml
```diff
@@ -2,18 +2,29 @@ name: Run Xbrl Conformance Suites
...
+          find tests/integration_tests -name 'test_*.py' -not -name test_efm_conformance_suite.py | jq -Rsc 'split("\n") | map(select(length > 0))'
+
   run-conformance-suite:
+    needs: find-tests
     permissions: read-all
     runs-on: ubuntu-latest
     continue-on-error: true
     strategy:
       matrix:
-        test-path:
```
sagesmith-wf, line 11: This doesn't really seem like an objective improvement, would suggest not doing this, but will leave it up to the team. My concerns are when we want to change, you now have to know `find`, and `jq`. I will leave up to the team though

aaroncameron-wk: It is more complicated, but the value it adds (running your PR's new tests before merging) is worth it in my opinion.

aaroncameron-wk (approved):

.github/workflows/conformance-suites.yml
austinmatherne-wk: I'm personally fine with both `find` and `jq`. They're both pretty standard utils

austinmatherne-wk (approved):

aaroncameron-wk:
QA +10 
Confirmed that tests are retrieved and run from head branch: https://github.com/aaroncameron-wk/arelle-public/pull/20

PR #224, opened 2022-07-11
Pin dependencies in requirements*.txt

ggravlingen:
#### Reason for change
Supply-chain attacks as an attack vector is a growing concern and unfortunately libraries hosted [on Pypi is no exception](https://www.bitdefender.com/blog/hotforsecurity/supply-chain-attack-detected-in-pypi-library/). One way of mitigating this risk is to pin dependencies.

Further to potential security risks, not pinning versions can create unintended bugs if a dependencies introduces a breaking change.

#### Description of change
This pull request pins dependencies to the version installed when running `pip install -r requirements-dev.txt` without having the depencenies pinned.

#### Steps to Test

**review**:
@Arelle/arelle

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,6 +1,6 @@
 # core modules
 lxml>=4.6.3
-openpyxl>=2.3.3
+OpenPyXL>=2.3.3
 regex==2022.7.9
 isodate==0.6.1
 aniso8601==9.0.1


requirements.txt
```diff
@@ -1,26 +1,26 @@
...
 openPyXL>=2.3.3
-regex
-isodate
-aniso8601
-cherrypy
-cheroot
+regex==2022.7.9
+isodate==0.6.1
+aniso8601==9.0.1
+CherryPy==18.7.0
```
sagesmith-wf, line 7: Nit: Here and several other places, is the change in capitalization on purpose?

requirements-dev.txt
```diff
@@ -1,7 +1,7 @@
 # The Ubuntu GUI exits immediately without error in 6.11.0.
```
sagesmith-wf, line 1: Arbitrary line to allow dependabot discussion

sagesmith-wf: > As a followup, I suggest me adding a Github workflow that uses [Dependabot](https://github.com/dependabot) to automatically bump dependencies as they become available on Pypi.

> Edit: I see you have used dependabot in the past. What was the reason for removing it?

I believe we removed it because we didn't have an appropriate level of automated test coverage to verify dependency bumps.

requirements-dev.txt
```diff
@@ -1,7 +1,7 @@
...
-cx_Freeze==6.10
-cx_Oracle
-sphinx
+cx_freeze==6.10
+cx-oracle==8.3.0
```
ggravlingen, line 3: This looks wrong. An underscore has been replaced by a dash. Will check why.

requirements-dev.txt
ggravlingen: That makes sense, I'll refrain from adding that workflow.

ggravlingen:
> @ggravlingen I see this is your first time contributing, thank you for your contribution! Appreciate the dependabot discussion as well!

Thanks for a great library! I've played around a bit with it and have been able to extract ESEF data in a meaningful way.

I have signed and mailed the CLA by the way.

requirements-dev.txt
austinmatherne-wk: [Package names are case-insensitive](https://peps.python.org/pep-0426/#name). I personally prefer the canonical names

requirements-dev.txt
austinmatherne-wk: With the addition of the conformance suites we can probably begin using dependabot along with a generous use of the ignore list to filter out anything we think likely requires manual testing (especially anything GUI related). 

As an aside, Arelle is a little bit of a strange beast in that it's both an application which should have pinned dependencies (this PR is a good first step) and is used as a library which should also separately have dependency version ranges (we'll need to set this up before publishing the official pypi package).

Ideally we'd use pipenv or poetry to lock transitive dependencies for the application/local dev side of things. @ggravlingen as an outside contributor would the use of pipenv or poetry introduce any potential friction for you using or contributing to the project?

ggravlingen: My current use case is to run `pip install git+https://github.com/Arelle/Arelle.git#0.9.91` along with pandas and lxml. I'm then able to use the library to parse XBRL-files in the ESEF format.

I have never used pipenv or poetry but just having had a quick look, I don't believe using those tools would affect how I'm using this library.

requirements-dev.txt
ggravlingen: I have no strong personal opinion on this, please advise if you want me to change.

austinmatherne-wk: I'd appreciate if you restore the names (and fix the existing `CherryPy` and `PyMySQL` names), but that won't hold up merging this PR.

There appears to have been an issue with the support@arelle.org email address that I believe you sent your signed CLA to (sorry for the inconvenience). Can you send a copy to Austin.Matherne@Workiva.com so we can get this merged?

ggravlingen: No problem! I had a look at the pypi page of each dependency and used the name state there. For `OpenPyXL` I was uncertain so I had a look at the lib's documentation page and it looks like they are using the `OpenPyXL` name there.

austinmatherne-wk (approved):

sagesmith-wf (approved):

PR #228, opened 2022-07-14
Fix bug in PackageManager.save() [incorrect solution to problem]

ggravlingen:
#### Reason for change
I was running through a couple of hundred reports in the ESEF-format when I encountered this error:
```
    cntlr.close()
  File "/opt/.venv/lib/python3.10/site-packages/arelle/Cntlr.py", line 496, in close
    PackageManager.save(self)
  File "/opt/.venv/lib/python3.10/site-packages/arelle/PackageManager.py", line 443, in save
    with io.open(packagesJsonFile, 'wt', encoding='utf-8') as f:
TypeError: expected str, bytes or os.PathLike object, not NoneType
```

#### Description of change
The change proposed here asserts that `packagesJsonFile` has been changed from its default value of `None` before using the variable to save to disk.

My IDE has also removed random spaces throughout the code.

#### Steps to Test
Not sure how to test this as I don't know in what cases the `packagesJsonFile` variable is not changed from `None`.

**review**:
@Arelle/arelle

### DIFF
--- a/arelle/PackageManager.py
+++ b/arelle/PackageManager.py
@@ -20,8 +20,8 @@ ArchiveFileIOError = None
 try:
     from collections import OrderedDict
 except ImportError:
-    OrderedDict = dict # python 3.0 lacks OrderedDict, json file will be in weird order 
-    
+    OrderedDict = dict # python 3.0 lacks OrderedDict, json file will be in weird order
+
 TP_XSD = "http://www.xbrl.org/2016/taxonomy-package.xsd"
 CAT_XSD = "http://www.xbrl.org/2016/taxonomy-package-catalog.xsd"
 
@@ -40,7 +40,7 @@ def baseForElement(element):
                 base = baseAttr + base
         baseElt = baseElt.getparent()
     return base
-   
+
 def xmlLang(element):
     return (element.xpath('@xml:lang') + element.xpath('ancestor::*/@xml:lang') + [''])[0]
 
@@ -57,7 +57,7 @@ def parsePackage(cntlr, filesource, metadataFile, fileBase, errors=[]):
         from arelle.FileSource import ArchiveFileIOError
 
     unNamedCounter = 1
-    
+
     txmyPkgNSes = ("http://www.corefiling.com/xbrl/taxonomypackage/v1",
                    "http://xbrl.org/PWD/2014-01-15/taxonomy-package",
                    "http://xbrl.org/PWD/2015-01-14/taxonomy-package",
@@ -65,7 +65,7 @@ def parsePackage(cntlr, filesource, metadataFile, fileBase, errors=[]):
                    "http://xbrl.org/2016/taxonomy-package",
                    "http://xbrl.org/WGWD/YYYY-MM-DD/taxonomy-package")
     catalogNSes = ("urn:oasis:names:tc:entity:xmlns:xml:catalog",)
-    
+
     pkg = {}
 
     currentLang = Locale.getLanguageCode()
@@ -84,11 +84,11 @@ def parsePackage(cntlr, filesource, metadataFile, fileBase, errors=[]):
                        level=logging.ERROR)
         errors.append("tpe:invalidMetaDataFile")
         return pkg
-    
+
     root = tree.getroot()
     ns = root.tag.partition("}")[0][1:]
     nsPrefix = "{{{}}}".format(ns)
-    
+
     if ns in  txmyPkgNSes:  # package file
         for eltName in ("identifier", "version", "license", "publisher", "publisherURL", "publisherCountry", "publicationDate"):
             pkg[eltName] = ''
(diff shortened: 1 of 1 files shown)


arelle/PackageManager.py
```diff
@@ -20,8 +20,8 @@
 try:
     from collections import OrderedDict
 except ImportError:
-    OrderedDict = dict # python 3.0 lacks OrderedDict, json file will be in weird order 
-    
+    OrderedDict = dict # python 3.0 lacks OrderedDict, json file will be in weird order
+
```
austinmatherne-wk, line 24: Thanks for the formatting cleanup. For future reference it's easier to review if you make formatting changes as a separate initial commit.

arelle/PackageManager.py
```diff
@@ -304,46 +304,46 @@ def init(cntlr, loadPackagesConfig=True):
...
                                                          'remappings': '14',
                                                          }.get(k[0],k[0])))
                        for _packageInfo in packagesConfig['packages']]),
          ('remappings',OrderedDict(sorted(packagesConfig['remappings'].items())))))
-    
+
 def save(cntlr):
     global packagesConfigChanged
-    if packagesConfigChanged and cntlr.hasFileSystem:
+    if packagesConfigChanged and cntlr.hasFileSystem and packagesJsonFile is not None:
```
austinmatherne-wk, line 335: This seems wrong. The package config has changed so I'd expect this to create the packages json file if it doesn't already exist, not silently skip the save operation.

arelle/PackageManager.py
ggravlingen: Noted, will do!

arelle/PackageManager.py
ggravlingen: Hm, what about this instead? We're just reusing the definition of `packagesJsonFile` from the init function.

```
def save(cntlr):
    global packagesConfigChanged
    if packagesConfigChanged and cntlr.hasFileSystem:
        if packagesJsonFile is None:
            packagesJsonFile = packagesJsonFile = cntlr.userAppDir + os.sep + "taxonomyPackages.json"

        with io.open(packagesJsonFile, 'wt', encoding='utf-8') as f:
            jsonStr = _STR_UNICODE(json.dumps(orderedPackagesConfig(), ensure_ascii=False, indent=2)) # might not be unicode in 2.7
            f.write(jsonStr)
        packagesConfigChanged = False
```

austinmatherne-wk: I'd move the json file line into a function that's also called from init so as not to duplicate the file name definition.

Thinking about this further, you ran into this from the CLI not the GUI, correct? The CLI shouldn't persist packages, so there must be a bug elsewhere that incorrectly caused the save method to be called when it shouldn't have. Do you have a larger stacktrace?

ggravlingen: Yes, this could potentially be an implementation error on my side. I've implemented Arelle in a local library in this way and the stacktrace is raised on the `model_xbrl.close()` line. Didn't save the rest of it unfortunately but the lines before are just from my local code.

```
try:
    model_manager: ModelManager = ModelManager.initialize(cntlr)
    model_xbrl: ModelXbrl = model_manager.load(
        url_filing, taxonomyPackages=url_taxonomy
    )

    model_roles = extract_model_roles(model_xbrl=model_xbrl)

    model_xbrl.close()
except Exception as exc:
    cntlr.addToLog(
        f"Error {entry.name} due to {exc}",
        level=logging.WARNING,
    )
```

austinmatherne-wk: Can you try putting the `PackageManager.save` call [in the `Cntlr`](https://github.com/Arelle/Arelle/blob/da93c2046fdfd64aa8c8747b70aa1533cc5d97ad/arelle/Cntlr.py#L392) behind a has gui check?
```Python
if self.hasGui:
    PackageManager.save(self)
```

ggravlingen: That did the trick - I'm not getting an error after doing that. Perhaps it's cleaner for the commit history to close this PR and open a new one with this change?

brettkail-wk: Rebasing and force-pushing also seems fine and would keep the conversations connected.  (Of course, if a new PR is opened that links to that one, the tracking is still there.)

ggravlingen: To be honest, I wouldn't know how to properly do such a rebase so I went for the safe option and created a new PR.

sagesmith-wf: few options, easiest thing is probably to pull latest master and then blow away your commits with `git reset upstream/master`, `upstream` being the name of your fork which will undo all your commits and then you can just recommit and `git push -f`. Other option is to `git rebase upstream/master` and then `git log` find the commit hash of the commit before your first commit and then `git rebase -i <COMMIT HASH>` and follow the instructions. All that to say that your other branch looks good :)

ggravlingen: @sagesmith-wf thanks, I've saved your comment here for future reference!

PR #231, opened 2022-07-18
Restore old linux build scripts

austinmatherne-wk:
#### Reason for change
We need to retain the original Arelle build scripts until we've completely switched over to the CI builds.

#### Description of change
Restores the linux build scripts removed in the CI Linux builds PR

#### Steps to Test
* Generate CI linux builds and make sure the tar file still contains an Arelle build

**review**:
@Arelle/arelle


sagesmith-wf (approved):

derekgengenbacher-wf (approved):

derekgengenbacher-wf:
QA +10 Testing locally succeeded. Verified that there is an arelle build in the tar file locally.

PR #232, opened 2022-07-18
OIM: check DTS for context element

### LINKED ISSUE #221: How to specify OIM-CSV Dimensions  has-hypercube with @xbrldt:contextElement="scenario"
Question

loadFromOIM.py creates a dimension under &lt;segment> by default.
Is there any parameter to specify has-hypercube with @xbrldt:contextElement="scenario"?

XBRL Japan, Nobu

austinmatherne-wk:
#### Reason for change
`segment` is hard coded in the loadFromOIM plugin. The [OIM xbrl-xml spec](https://www.xbrl.org/Specification/xbrl-xml/REC-2021-10-13/xbrl-xml-REC-2021-10-13.html#sec-dimensions) describes how the proper context element from the DTS should be determined. This PR does not address taxonomies which use both.

> When serialising to xBRL-XML, all taxonomy-defined dimensions for all facts in a report are serialised to the same container element (either <xbrli:segment> or <xbrli:scenario> ). If the report's DTS contains hypercubes for a single container, then that container is used for all taxonomy-defined dimensions. If the DTS contains hypercubes for both containers, the container should be chosen so as to achieve dimensional validity for all facts, if possible. If this is not possible, the container should be chosen arbitrarily, and appropriate XBRL Dimensions [[DIMENSIONS]](https://www.xbrl.org/Specification/xbrl-xml/REC-2021-10-13/xbrl-xml-REC-2021-10-13.html#DIMENSIONS) errors raised. If the report's DTS does not contain any hypercubes, or if dimensional validity can be achieved using either container, <xbrli:scenario> should be used for all dimensions.

#### Description of change
fixes #221 

#### Steps to Test
Load [an OIM report](https://github.com/Arelle/Arelle/files/9135370/oim-scenario.zip) with a taxonomy that has scenario has-hypercube relationships.

`python -m arelleCmdLine --plugins loadFromOIM|saveLoadableOIM -f oim-example-metadata.json --saveOIMinstance oim-example-metadata.xml`

**review**:
@Arelle/arelle

### DIFF
--- a/arelle/plugin/loadFromOIM.py
+++ b/arelle/plugin/loadFromOIM.py
@@ -118,6 +118,7 @@ NSReservedAliasURIPrefixes = { # for starts-with checking
     }
 NSReservedURIAlias = {}
 
+OIMDefaultContextElement = "scenario"
 OIMReservedAliasURIs = {
     # "namespaces": NSReservedAliasURIs,  -- generated at load time
     "linkTypes": reservedLinkTypesAndGroups, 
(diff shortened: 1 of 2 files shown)


derekgengenbacher-wf (dismissed):

arelle/plugin/loadFromOIM.py
```diff
@@ -118,6 +118,8 @@
     }
 NSReservedURIAlias = {}
 
+# https://www.xbrl.org/Specification/xbrl-xml/REC-2021-10-13/xbrl-xml-REC-2021-10-13.html#sec-dimensions
+OIMDefaultContextElement = "scenario"
```
brettkail-wk, line 122: To be sure, the algorithm doesn't strictly follow the algorithm from the spec in cast the taxonomy intends for segment to be used (and the report would be valid if segment were used), but it also contains unused scenario elements.  That's presumably burdensome to implement, so perhaps just add a small comment for now?

arelle/plugin/loadFromOIM.py
```diff
@@ -2106,6 +2108,12 @@ def checkDim(tblTmpl, dimName, dimValue, *pathSegs):
         
         numFactCreationXbrlErrors = 0
         
+        taxonomyContextRefs = {
+            modelRelationship.contextElement
+            for hasHypercubeRelationship in (XbrlConst.all, XbrlConst.notAll)
```
sagesmith-wf, line 2113: This seems kind of unfortunately slow in some cases. It's probably a lot more difficult of a fix but it seems like it'd maybe be more cost effective to do in the DTS parsing code itself. However, this does fix the bug so I won't push very hard on this. It'd be nice to see a test if possible

sagesmith-wf (commented):
Can we add a test please?

arelle/plugin/loadFromOIM.py
austinmatherne-wk: Correct. I wrote a comment. Let me know if it makes sense.

We briefly discussed this as a team and decided the complexity wasn't worth the effort right now given that the case is uncommon.

derekgengenbacher-wf (approved):

aaroncameron-wk (approved):

PR #269, opened 2022-08-06
Add type hint esef dimensions

ggravlingen:
#### Reason for change
This continues the work of adding type hints.

#### Description of change
- General cleanup of whitespace and unused imports.
- Adding arelle.plugin.validate.ESEF.Dimensions to the strict typing list.
- Adding missing type hints in arelle.plugin.validate.ESEF.Dimensions and to methods and properties that being used in the Dimensions-file.

Needs rebasing when #267 is merged. I'll keep this as a draft until then.

#### Steps to Test

**review**:
@Arelle/arelle


aviary3-wk:
## Security Insights

## Action Items
---

arelle/plugin/validate/ESEF/Dimensions.py
```diff
@@ -18,28 +19,38 @@
...
     elrPrimaryItems = defaultdict(set)
-    hcPrimaryItems = set()
-    hcMembers = set()
-    
-    def addDomMbrs(sourceDomMbr, ELR, membersSet):
+    hcPrimaryItems: set[ModelConcept] = set()
+    hcMembers: set[Any] = set()
+
+    def addDomMbrs(sourceDomMbr: ModelConcept, ELR: str, membersSet: set[ModelConcept]) -> None:
+        print(f">>>>>>>>>>>>>>> membersSet {type(membersSet)}")
```
ggravlingen, line 37:
```suggestion
```

sagesmith-wf (dismissed):

austinmatherne-wk (approved):

sagesmith-wf (dismissed):

arelle/ModelXbrl.py
```diff
@@ -10,7 +10,7 @@
 from decimal import Decimal
 from arelle import UrlUtil, XmlUtil, ModelValue, XbrlConst, XmlValidate
 from arelle.FileSource import FileNamedStringIO
-from arelle.ModelDtsObject import ModelRelationship
+from arelle.ModelRelationshipSet import ModelRelationshipSet as ModelRelationshipSetClass
```
austinmatherne-wk, line 13: This seems problematic. The naming conflict that this is trying to avoid is for a dynamic import from the same module (I'm not sure why it's dynamic). If that import does in fact need to be dynamic then I'd expect this to break. Otherwise the dynamic import should be dropped.

arelle/plugin/validate/ESEF/Dimensions.py
```diff
@@ -18,28 +19,37 @@
...
             membersSet.add(sourceDomMbr)
             for domMbrRel in val.modelXbrl.relationshipSet(XbrlConst.domainMember, ELR).fromModelObject(sourceDomMbr):
                 #if domMbrRel.isUsable:
                 addDomMbrs(domMbrRel.toModelObject, domMbrRel.consecutiveLinkrole, membersSet)
-            
+
     for hasHypercubeArcrole in (XbrlConst.all, XbrlConst.notAll):
         hasHypercubeRelationships = val.modelXbrl.relationshipSet(hasHypercubeArcrole).fromModelObjects()
+
+        # fromModelObjects() has an optional return type, we need to handle that here
```
austinmatherne-wk, line 46: Nice

arelle/ValidateXbrl.py
```diff
@@ -53,14 +53,14 @@ class ValidateXbrl:
...
-    domainMembers: list
+    domainMembers: set = set()  # concepts which are dimension domain members
     extensionImportedUrls: set
     hasExtensionCal: bool
     hasExtensionDef: bool
     hasExtensionLbl: bool
     hasExtensionPre: bool
     hasExtensionSchema: bool
-    primaryItems: list
+    primaryItems: set = set()  # concepts which are line items (should not also be dimension members
```
austinmatherne-wk, line 63: Why do these need to be initialized?

arelle/ModelXbrl.py
ggravlingen: Oh, didn't intend to cause breaking changes. I'll refactor.

ggravlingen: Would this work instead? It'd be good to have a defined return-type on `relationshipSet`.

arelle/ValidateXbrl.py
ggravlingen: Didn't realise it could be done from `Dimensions.py`. I'll revert.

arelle/ValidateXbrl.py
```diff
@@ -53,14 +53,14 @@ class ValidateXbrl:
 
     authority: str
     authParam: dict[str, Any]
-    domainMembers: list
+    domainMembers: set = set()  # concepts which are dimension domain members
```
austinmatherne-wk, line 56: My previous comment applies to this, as well. Initializing here turns this into a class field that's no longer scoped to a instance.
```python
    instance_1 = ValidateXbrl(Mock())
    instance_1.domainMembers.add('domain_member_1')

    instance_2 = ValidateXbrl(Mock())
    instance_2.domainMembers.add('domain_member_2')

    print(instance_1.domainMembers)
    # {'domain_member_1', 'domain_member_2'}
    print(instance_2.domainMembers)
    # {'domain_member_1', 'domain_member_2'}
```

arelle/ModelXbrl.py
```diff
@@ -10,7 +10,8 @@
 from decimal import Decimal
 from arelle import UrlUtil, XmlUtil, ModelValue, XbrlConst, XmlValidate
 from arelle.FileSource import FileNamedStringIO
-from arelle.ModelDtsObject import ModelRelationship
+from arelle.ModelRelationshipSet import ModelRelationshipSet
+from arelle.ModelRelationshipSet import ModelRelationshipSet as ModelRelationshipSetReturn
```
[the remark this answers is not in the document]
ggravlingen: If I use `ModelRelationshipSet` from `from arelle.ModelRelationshipSet import ModelRelationshipSet` as a return type:
image

If I use `ModelRelationshipSetReturn` from `from arelle.ModelRelationshipSet import ModelRelationshipSet as ModelRelationshipSetReturn` as a return type:
image

I'm trying to achieve the latter, ie proper type hints for `fromModelObjects`.

arelle/ValidateXbrl.py
ggravlingen: Thanks for the clarification, I didn't know this.

arelle/ModelXbrl.py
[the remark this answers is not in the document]
austinmatherne-wk: I'm assuming that's because `ModelRelationshipSet` is set to None further below?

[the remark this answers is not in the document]
ggravlingen: Yes, commenting out `ModelRelationshipSet = None # dynamic import` gives us a correct return type.

What about creating a `TYPE_MODEL_RELATIONSHIP_SET = ModelRelationshipSet` after all the imports and then using the constant as a return type? It yields the desired result.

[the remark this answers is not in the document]
austinmatherne-wk: The variable name isn't what's important here. The dynamic import is (presumably) trying to avoid a circular dependency so loading the module at import time will cause problems regardless of what it's called.

That being said, there's no harm in importing a module multiple times (it'll simple return the existing imported module), so the existing `ModelRelationshipSet = None` pattern is unnecessary.

Use [typing.TYPE_CHECKING](https://docs.python.org/3/library/typing.html#typing.TYPE_CHECKING) in the header:
```python
if TYPE_CHECKING:
    from arelle import ModelRelationshipSet
```

and [remove the `ModelRelationshipSet = None` assignment](https://github.com/Arelle/Arelle/blob/451b9484d7453f71bb6b20935aa3365125bc1248/arelle/ModelXbrl.py#L23) along with unconditionally importing `ModelRelationshipSet` [from within `relationshipSet(...)`](https://github.com/Arelle/Arelle/blob/451b9484d7453f71bb6b20935aa3365125bc1248/arelle/ModelXbrl.py#L397-L399)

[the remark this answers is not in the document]
ggravlingen: Thanks! I had to do the import for `TYPE_CHECKING` like so: `from arelle.ModelRelationshipSet import ModelRelationshipSet`, otherwise type hints didn't work.

arelle/ModelRelationshipSet.py
```diff
@@ -204,50 +206,50 @@ def clear(self):
...
         if self.modelRelationshipsTo is None:
             self.modelRelationshipsTo = defaultdict(list)
             for modelRel in self.modelRelationships:
                 toModelObject = modelRel.toModelObject
                 if toModelObject is not None:   # none if concepts failed to load
                     self.modelRelationshipsTo[toModelObject].append(modelRel)
-                
-    def fromModelObjects(self):
+
+    def fromModelObjects(self) -> defaultdict[Any, list] | None:
```
austinmatherne-wk, line 235: IMO the fact that `self.modelRelationshipsFrom` is a `defaultdict` is really more of an internal class implementation detail that shouldn't leak out of the class through the type system (return type would just be dict). Also, this function will never return None.

arelle/ModelXbrl.py
```diff
@@ -395,8 +399,7 @@ def relationshipSet(self, arcrole, linkrole=None, linkqname=None, arcqname=None,
         :returns: [ModelRelationship] -- Ordered list of effective relationship objects per parameters
         """
         global ModelRelationshipSet
```
austinmatherne-wk, line 401: remove: `global ModelRelationshipSet` is no longer needed

arelle/ModelXbrl.py
```diff
@@ -20,7 +21,10 @@
 from arelle.ValidateXbrlDimensions import isFactDimensionallyValid
 
 
-ModelRelationshipSet = None # dynamic import
+
+if TYPE_CHECKING:
+    from arelle.ModelRelationshipSet import ModelRelationshipSet
```
austinmatherne-wk, line 26: These imports are on separate paths so it works, but it's a bit awkward that `ModelRelationshipSet` refers to two separate things (the class here and the module below). I'd actually use your `ModelRelationshipSetClass` name from before here

```suggestion
    from arelle.ModelRelationshipSet import ModelRelationshipSet as ModelRelationshipSetClass
```

nit: also too many blank lines above

arelle/plugin/validate/ESEF/Dimensions.py
```diff
@@ -18,28 +19,40 @@
...
-    val.domainMembers = set() # concepts which are dimension domain members
+    val.domainMembers = set()  # concepts which are dimension domain members
+
     elrPrimaryItems = defaultdict(set)
-    hcPrimaryItems = set()
-    hcMembers = set()
-    
-    def addDomMbrs(sourceDomMbr, ELR, membersSet):
+    hcPrimaryItems: set[ModelConcept] = set()
+    hcMembers: set[Any] = set()
```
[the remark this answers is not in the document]
ggravlingen: Not in the world of mypy if I'm not misunderstanding you here. Setting `hcMembers = set()` yields this:
```
(.venv) vscode ➜ /workspaces/Arelle (add-type-hint-esef-dimensions ✗) $ mypy arelle --namespace-packages
arelle/plugin/validate/ESEF/Dimensions.py:38: error: Need type annotation for "hcMembers" (hint: "hcMembers: Set[<type>] = ...")  [var-annotated]
Found 1 error in 1 file (checked 263 source files)
```

It probably should be something other than `Any` but at the moment, I don't know exactly what. My thinking is that it could be adressed further down the line.

arelle/ModelRelationshipSet.py
ggravlingen: Oh, VS suggested it would return `None` or `defaultdict` and I believe the reason for that is it's being defined as `None` here:
https://github.com/Arelle/Arelle/blob/dbcf5fd3e89edc45c99481dcb4438b2158e27ee8/arelle/ModelRelationshipSet.py#L182

And `defaultdict` here:
https://github.com/Arelle/Arelle/blob/dbcf5fd3e89edc45c99481dcb4438b2158e27ee8/arelle/ModelRelationshipSet.py#L219

image

Do we want to remove the definition on line 182? Mypy doesn't complain if I remove `None` as a return type in `fromModelObjects` though. I don't feel strongly about this through, I'm happy to remove `None` as the return type.

austinmatherne-wk: `self.modelRelationshipsFrom` can be `None`, but never when it's the return value of this function. The call to `loadModelRelationshipsFrom` always sets it to a dict before this returns.

arelle/plugin/validate/ESEF/Dimensions.py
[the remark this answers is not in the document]
austinmatherne-wk: Gotcha, it wants an explicit `Any`. Makes sense.

arelle/ModelRelationshipSet.py
ggravlingen: I see what you mean now, thanks!

austinmatherne-wk (dismissed):

austinmatherne-wk (approved):

ggravlingen:
To cleanup the commit history, I followed the advice from a previous PR and rebased. I hope I didn't mess anything up now.

arelle/plugin/validate/ESEF/Dimensions.py
```diff
@@ -18,28 +19,40 @@
...
                 #if domMbrRel.isUsable:
                 addDomMbrs(domMbrRel.toModelObject, domMbrRel.consecutiveLinkrole, membersSet)
-            
+
     for hasHypercubeArcrole in (XbrlConst.all, XbrlConst.notAll):
         hasHypercubeRelationships = val.modelXbrl.relationshipSet(hasHypercubeArcrole).fromModelObjects()
+
+        # fromModelObjects() has an optional return type, we need to handle that here
+        if hasHypercubeRelationships is None:
+            continue
```
ggravlingen, line 51: I'll revert this in a follow-up, we agreed that `hasHypercubeRelationships` cannot be `None`.

Issue #309, opened 2022-09-01
Remove Python 2 Support


austinmatherne-wk:
Arelle contains legacy conditional code for running under both Python 2 and 3. Since Python 2 is no longer supported the Python 2 code paths can be removed.

Files with Python 2 conditional code to remove:
* [apidocs/conf.py](https://github.com/Arelle/Arelle/blob/d9fe63173b84df74d3ed15a0afe37214df85ca82/apidocs/conf.py)
* [arelle/Cntlr.py](https://github.com/Arelle/Arelle/blob/d9fe63173b84df74d3ed15a0afe37214df85ca82/arelle/Cntlr.py)
* [arelle/PackageManager.py](https://github.com/Arelle/Arelle/blob/d9fe63173b84df74d3ed15a0afe37214df85ca82/arelle/PackageManager.py)
* [arelle/PythonUtil.py](https://github.com/Arelle/Arelle/blob/d9fe63173b84df74d3ed15a0afe37214df85ca82/arelle/PythonUtil.py)
* [arelle/RenderingResolver.py](https://github.com/Arelle/Arelle/blob/d9fe63173b84df74d3ed15a0afe37214df85ca82/arelle/RenderingResolver.py)
* [arelle/UrlUtil.py](https://github.com/Arelle/Arelle/blob/d9fe63173b84df74d3ed15a0afe37214df85ca82/arelle/UrlUtil.py)
* [arelle/ViewFile.py](https://github.com/Arelle/Arelle/blob/d9fe63173b84df74d3ed15a0afe37214df85ca82/arelle/ViewFile.py)
* [arelle/WebCache.py](https://github.com/Arelle/Arelle/blob/d9fe63173b84df74d3ed15a0afe37214df85ca82/arelle/WebCache.py)
* [arelle/XPathParser.py](https://github.com/Arelle/Arelle/blob/d9fe63173b84df74d3ed15a0afe37214df85ca82/arelle/XPathParser.py)
* [arelle/plugin/loadFromOIM-2018.py](https://github.com/Arelle/Arelle/blob/d9fe63173b84df74d3ed15a0afe37214df85ca82/arelle/plugin/loadFromOIM-2018.py)
* [arelle/plugin/saveLoadableOIM.py](https://github.com/Arelle/Arelle/blob/d9fe63173b84df74d3ed15a0afe37214df85ca82/arelle/plugin/saveLoadableOIM.py)
[...]

PR #276, opened 2022-08-12
Finalise adding type hints for arelle.plugin.validate.esef

ggravlingen:
#### Reason for change
This enforces strict typing for the entire `arelle.plugin.validate.esef` package.

#### Description of change
This PR:
- Cleans up white space in a number of files
- Removes unused imports
- Adds type hints in `arelle.plugin.validate.ESEF.__init__` and files used herein.
- Moves type hints for `_` (gettext, I believe) to a common file.

#### Steps to Test

**review**:
@Arelle/arelle

PR #313, opened 2022-09-03
add type hints Cntlr.py and related files round01

selgamal:
#### Reason for change
Type hints for arelle/Cntlr.py and related files

#### Description of change
At this point still having 37 errors when checking locally, and would like to have some feedback. On group of error I don't know how to deal with is that relating to attributes added to logger and logRecord classes.

Also proposed to change the TypeGetText to return Any, its not guranteed to return text as far as I know.

Current list of errors:
```bash
arelle/Cntlr.py:254: error: Cannot find implementation or library stub for module named "gtk"  [import]
arelle/Cntlr.py:254: note: See https://mypy.readthedocs.io/en/stable/running_mypy.html#missing-imports
arelle/Cntlr.py:315: error: Value of type Module is not indexable  [index]
arelle/Cntlr.py:318: error: Name "_STR_UNICODE" is not defined  [name-defined]
arelle/Cntlr.py:360: error: "Logger" has no attribute "logRefObjectProperties"  [attr-defined]
arelle/Cntlr.py:363: error: "Logger" has no attribute "logRefObjectProperties"  [attr-defined]
arelle/Cntlr.py:378: error: Module has no attribute "_levelNames"; maybe "_levelToName"?  [attr-defined]
arelle/Cntlr.py:381: error: Module has no attribute "loggingLevelNums"  [attr-defined]
arelle/Cntlr.py:382: error: Name "_INT_TYPES" is not defined  [name-defined]
arelle/Cntlr.py:384: error: "Logger" has no attribute "messageCodeFilter"  [attr-defined]
arelle/Cntlr.py:385: error: "Logger" has no attribute "messageLevelFilter"  [attr-defined]
arelle/Cntlr.py:386: error: "Handler" has no attribute "logTextMaxLength"  [attr-defined]
arelle/Cntlr.py:390: error: "Logger" has no attribute "messageLevelFilter"  [attr-defined]
arelle/Cntlr.py:394: error: "Logger" has no attribute "messageCodeFilter"  [attr-defined]
arelle/Cntlr.py:427: error: Name "_STR_BASE" is not defined  [name-defined]
arelle/Cntlr.py:429: error: Name "_STR_BASE" is not defined  [name-defined]
arelle/Cntlr.py:430: error: Module has no attribute "_checkLevel"  [attr-defined]
arelle/Cntlr.py:474: error: Name "_STR_UNICODE" is not defined  [name-defined]
arelle/Cntlr.py:581: error: Returning Any from function declared to return "Optional[str]"  [no-any-return]
arelle/Cntlr.py:594: error: Cannot find implementation or library stub for module named "win32process"  [import]
arelle/Cntlr.py:595: error: "None" has no attribute "GetProcessMemoryInfo"  [attr-defined]
arelle/Cntlr.py:595: error: "None" has no attribute "GetCurrentProcess"  [attr-defined]
arelle/Cntlr.py:603: error: Name "osPrcs" already defined (possibly by an import)  [no-redef]
arelle/Cntlr.py:604: error: Returning Any from function declared to return "Optional[float]"  [no-any-return]
arelle/Cntlr.py:604: error: "None" has no attribute "getrusage"  [attr-defined]
arelle/Cntlr.py:604: error: "None" has no attribute "RUSAGE_SELF"  [attr-defined]
arelle/Cntlr.py:626: error: "LogRecord" has no attribute "refs"  [attr-defined]
arelle/Cntlr.py:635: error: "LogRecord" has no attribute "messageCode"; maybe "message"?  [attr-defined]
arelle/Cntlr.py:641: error: "LogRecord" has no attribute "file"  [attr-defined]
arelle/Cntlr.py:668: error: Incompatible types in assignment (expression has type "bytes", variable has type "str")  [assignment]
arelle/Cntlr.py:712: error: Item "Tuple[object, ...]" of "Union[Tuple[object, ...], Mapping[str, object]]" has no attribute "items"  [union-attr]
arelle/Cntlr.py:723: error: "LogRecord" has no attribute "refs"  [attr-defined]
arelle/Cntlr.py:726: error: "LogRecord" has no attribute "messageCode"; maybe "message"?  [attr-defined]
arelle/Cntlr.py:734: error: Item "Tuple[object, ...]" of "Union[Tuple[object, ...], Mapping[str, object]]" has no attribute "items"  [union-attr]
arelle/Cntlr.py:735: error: Incompatible types in assignment (expression has type "Union[Any, object]", target has type "str")  [assignment]
arelle/Cntlr.py:736: error: "LogRecord" has no attribute "messageCode"; maybe "message"?  [attr-defined]
arelle/Cntlr.py:738: error: "LogRecord" has no attribute "refs"  [attr-defined]
arelle/Cntlr.py:788: error: Incompatible types in assignment (expression has type "bytes", variable has type "str")  [assignment]
```

Update 1:
Cleared all pending errors, changed annotation style to python 3.10 style, resolved comments.
*Note* that following type check ignores were added for the following errors:
```bash
arelle/Cntlr.py:247: error: Cannot find implementation or library stub for module named "gtk"  [import] 
arelle/Cntlr.py:366: error: Name "_INT_TYPES" is not defined  [name-defined] # builtins
arelle/Cntlr.py:411: error: Name "_STR_BASE" is not defined  [name-defined] # builtins
arelle/Cntlr.py:413: error: Name "_STR_BASE" is not defined  [name-defined] # builtins
arelle/Cntlr.py:463: error: Name "_STR_UNICODE" is not defined  [name-defined] # builtins
arelle/Cntlr.py:595: error: "None" has no attribute "getrusage"  [attr-defined] # has to do with global import
arelle/Cntlr.py:595: error: "None" has no attribute "RUSAGE_SELF"  [attr-defined] # has to do with global import
```

#### Steps to Test

**review**:
@Arelle/arelle

### DIFF
--- a/arelle/Cntlr.py
+++ b/arelle/Cntlr.py
@@ -10,7 +10,7 @@
 """
 from __future__ import annotations
 from typing import Any, TYPE_CHECKING, TextIO, Mapping
-from .typing import TypeGetText
+from arelle.typing import TypeGetText
 import tempfile, os, io, sys, logging, gettext, json, re, subprocess, math
 from arelle import ModelManager
 from arelle.WebCache import WebCache
--- a/arelle/Cntlr.py
+++ b/arelle/Cntlr.py
@@ -9,7 +9,7 @@
    :synopsis: Common controller class to initialize for platform and setup common logger functions
 """
 from __future__ import annotations
-from typing import Optional, Union, Any, TYPE_CHECKING, TextIO
+from typing import Any, TYPE_CHECKING, TextIO, Mapping
 from .typing import TypeGetText
 from arelle import PythonUtil # define 2.x or 3.x string types
 import tempfile, os, io, sys, logging, gettext, json, re, subprocess, math
@@ -117,7 +117,7 @@ class Cntlr:
     systemWordSize: int
     uiLang: str
     uiLangDir: str
-    updateURL: Optional[str]
+    updateURL: str | None
     configDir: str
     imagesDir: str
     localeDir: str
@@ -128,26 +128,20 @@ class Cntlr:
     hasClipboard: bool
     contextMenuClick: str
     hasWebServer: bool
-    config: Optional[dict[str, Any]]
+    config: dict[str, Any] | None
     configJsonFile: str
     webCache: WebCache
     modelManager: ModelManager.ModelManager
-    logger: Optional[logging.Logger]
-    logHandler: Union[
-        LogToBufferHandler,
-        LogToPrintHandler,
-        LogToXmlHandler,
-        logging.Handler,
-        logging.FileHandler
-        ]
+    logger: logging.Logger | None
+    logHandler: logging.Handler
 
     def __init__(
         self,
         hasGui: bool = False,
-        logFileName: Optional[str] = None,
-        logFileMode: Optional[str] = None,
-        logFileEncoding: Optional[str] = None,
-        logFormat: Optional[str] = None
+        logFileName: str | None = None,
+        logFileMode: str | None = None,
+        logFileEncoding: str | None = None,
+        logFormat: str | None = None
     ) -> None:
         self.hasWin32gui = False
         self.hasGui = hasGui
@@ -250,7 +244,7 @@ class Cntlr:
                     self.userAppDir = os.path.join( os.path.expanduser("~/.config"), "arelle")
             if hasGui:
                 try:
-                    import gtk
+                    import gtk  # type: ignore[import]
                     self.hasClipboard = True
                 except ImportError:
                     self.hasClipboard = False
@@ -319,14 +313,14 @@ class Cntlr:
 
     def startLogging(
         self,
-        logFileName: Optional[str] = None,
-        logFileMode: Optional[str] = None,
-        logFileEncoding: Optional[str] = None,
-        logFormat: Optional[str] = None,
-        logLevel: Optional[str] = None,
-        logHandler: Optional[logging.Handler] = None,
+        logFileName: str | None = None,
+        logFileMode: str | None = None,
+        logFileEncoding:str | None = None,
+        logFormat: str | None = None,
+        logLevel: str | None = None,
+        logHandler: logging.Handler | None = None,
         logToBuffer: bool = False,
-        logTextMaxLength: Optional[int] = None,
+        logTextMaxLength: int | None = None,
         logRefObjectProperties: bool = True
     ) -> None:
         # add additional logging levels (for python 2.7, all of these are ints)
@@ -368,8 +362,8 @@ class Cntlr:
             except ValueError:
                 self.addToLog(_("Unknown log level name: {0}, please choose from {1}").format(
                     logLevel, ', '.join(logging.getLevelName(l).lower()
-                                        for l in sorted([i for i in logging.loggingLevelNums.keys()
-                                                         if isinstance(i,_INT_TYPES) and i > 0]))),
+                                        for l in sorted([i for i in logging._levelToName.keys()
+                                                         if isinstance(i,_INT_TYPES) and i > 0]))),  # type: ignore[name-defined]
                               level=logging.ERROR, messageCode="arelle:logLevel")
             setattr(self.logger, "messageCodeFilter", None)
             setattr(self.logger, "messageLevelFilter", None)
@@ -403,7 +397,7 @@ class Cntlr:
         :param file: File name (and optional line numbers) pertaining to message
         :type file: str
         """
-        args: Union[tuple[str], tuple[str, dict[str, Any]]]
+        args: tuple[str] | tuple[str, dict[str, Any]]
         if self.logger is not None:
             if messageArgs:
                 args = (message, messageArgs)
@@ -414,10 +408,9 @@ class Cntlr:
             if isinstance(file, (tuple,list,set)):
                 for _file in file:
                     refs.append( {"href": _file} )
-            elif isinstance(file, _STR_BASE):
+            elif isinstance(file, _STR_BASE):  # type: ignore[name-defined]
                 refs.append( {"href": file} )
-            if isinstance(level, _STR_BASE):
-                level = logging._checkLevel(level)
+            if isinstance(level, _STR_BASE):  # type: ignore[name-defined]
                 # given level is str at this point, level_int will always
                 # be an int but logging.getLevelName returns Any (int if
                 # input is str, and str if input is int)
@@ -433,7 +426,7 @@ class Cntlr:
                        .encode(sys.stdout.encoding, 'backslashreplace')
                        .decode(sys.stdout.encoding, 'strict')))
 
-    def showStatus(self, message: str, clearAfter: Optional[int] = None) -> None:
+    def showStatus(self, message: str, clearAfter: int | None = None) -> None:
         """Dummy method for specialized controller classes to specialize,
         provides user feedback on status line of GUI or web page
 
@@ -466,7 +459,8 @@ class Cntlr:
         """Save user preferences configuration (in json configuration file)."""
         if self.hasFileSystem:
             with io.open(self.configJsonFile, 'wt', encoding='utf-8') as f:
-                jsonStr = _STR_UNICODE(json.dumps(self.config, ensure_ascii=False, indent=2)) # might not be unicode in 2.7
+                # might not be unicode in 2.7
+                jsonStr = _STR_UNICODE(json.dumps(self.config, ensure_ascii=False, indent=2)) # type: ignore[name-defined]
                 f.write(jsonStr)  # 2.7 getss unicode this way
 
     # default non-threaded viewModelObject
@@ -536,7 +530,7 @@ class Cntlr:
         return 'cancel'
 
     # if no text, then return what is on the clipboard, otherwise place text onto clipboard
-    def clipboardData(self, text: Optional[str]=None) -> Optional[str]:
+    def clipboardData(self, text: str | None = None) -> Any:
         """Places text onto the clipboard (if text is not None), otherwise retrieves and returns text from the clipboard.
         Only supported for those platforms that have clipboard support in the current python implementation (macOS
         or ActiveState Windows Python).
@@ -581,7 +575,7 @@ class Cntlr:
         return None
 
     @property
-    def memoryUsed(self) -> Optional[float]:
+    def memoryUsed(self) -> float | None:
         try:
             global osPrcs # is this needed?
             # to tell mypy this is for windows
@@ -615,7 +609,7 @@ def logRefsFileLines(refs: list[dict[str, Any]]) -> str:
                     for file, lines in sorted(fileLines.items()))
 
 class LogFormatter(logging.Formatter):
-    def __init__(self, fmt: Optional[str] = None, datefmt: Optional[str] = None) -> None:
+    def __init__(self, fmt: str | None = None, datefmt: str | None = None) -> None:
         super(LogFormatter, self).__init__(fmt, datefmt)
 
     def fileLines(self, record: logging.LogRecord) -> str:
@@ -650,7 +644,7 @@ class LogToPrintHandler(logging.Handler):
     :param logOutput: 'logToStdErr' to cause log print to stderr instead of stdout
     :type logOutput: str
     """
-    logFile: Optional[Union[str, TextIO]]
+    logFile: str | TextIO | None
 
     def __init__(self, logOutput: str) -> None:
         super(LogToPrintHandler, self).__init__()
--- a/arelle/Cntlr.py
+++ b/arelle/Cntlr.py
@@ -25,7 +25,6 @@ if TYPE_CHECKING:
     from arelle.ModelXbrl import ModelXbrl
 
 osPrcs = None
-isPy3 = (sys.version[0] >= '3')
 LOG_TEXT_MAX_LENGTH = 32767
 cxFrozen = getattr(sys, 'frozen', False)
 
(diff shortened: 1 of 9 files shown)


aviary3-wk:
## Security Insights

## Action Items
---

arelle/Locale.py
```diff
@@ -12,6 +12,7 @@
 '''
 import sys, subprocess
```
ggravlingen, line 13: add `from __future__ import annotations` here

arelle/Locale.py
```diff
@@ -78,7 +79,7 @@ def getLanguageCode():
     except (AttributeError, ValueError): #language code and encoding may be None if their values cannot be determined.
         return "en"
 
-def getLanguageCodes(lang=None):
+def getLanguageCodes(lang: Optional[str] = None) -> list[str, str, str]:
```
ggravlingen, line 82: When using add from __future__ import annotations you can type using `lang: str | None = None`, which is the Python 3.10+ way of typing.

Not sure about the list though. If it's a list of strings that's being returned, the return type is simply `list[str]`

arelle/WebCache.py
```diff
@@ -84,7 +89,11 @@ class WebCache:
 
     default_timeout = None
 
-    def __init__(self, cntlr, httpProxyTuple):
+    def __init__(
+        self, cntlr: Cntlr,
+        httpProxyTuple: Optional[tuple[bool, str, str, str, str]]
```
ggravlingen, line 94:
```suggestion
        httpProxyTuple: tuple[bool, str, str, str, str] | None
```

ggravlingen: Also add `from future` etc in the import section

arelle/typing.py
```diff
@@ -1,5 +1,6 @@
 """Type hints for Arelle."""
 from __future__ import annotations
+from typing import Any
 from collections.abc import Callable
 
-TypeGetText = Callable[[str], str]
+TypeGetText = Callable[[str], Any]
```
ggravlingen, line 6: Are you sure about this? Doesn't the translation feature always return a translated string?

arelle/Locale.py
selgamal: Yes string thing looks funny!
Didn't know about the | thing, I will fix that every where... thank you!!!

arelle/Cntlr.py
```diff
@@ -265,8 +312,8 @@ def setUiLanguage(self, lang, fallbackToDefault=False):
             self.uiLang = langCodes[0].lower()
             self.uiLangDir = 'rtl' if self.uiLang[0:2] in {"ar","he"} else 'ltr'
             if not isPy3: # 2.7 gettext provides string instead of unicode from .mo files
-                installedGettext = __builtins__['_']
-                def convertGettextResultToUnicode(msg):
+                installedGettext: TypeGetText = __builtins__['_']
+                def convertGettextResultToUnicode(msg: str) -> Any:
                     translatedMsg = installedGettext(msg)
                     if isinstance(translatedMsg, _STR_UNICODE):
```
ggravlingen, line 318: Add `# type: ignore[name-defined]` here and in all places where you get the error `name-defined`

arelle/WebCache.py
selgamal: > Also add `from future` etc in the import section

Yep got it

ggravlingen:
When you see this:

```
arelle/Cntlr.py:384: error: "Logger" has no attribute "messageCodeFilter"  [attr-defined]
arelle/Cntlr.py:385: error: "Logger" has no attribute "messageLevelFilter"  [attr-defined]
arelle/Cntlr.py:386: error: "Handler" has no attribute "logTextMaxLength"  [attr-defined]
```

It means the property isn't defined on the logger or handler class. So you need to go to the logger instance and add eg:
`messageCodeFilter: str` or whatever it is (maybe just `Any`?) That's the reason I made this change for instance:
https://github.com/Arelle/Arelle/pull/276/files#diff-5aeb72f32f79c477f03e5d73f005abf50d15ae7b168d90e5fd4c544d0558d33cR269

arelle/typing.py
selgamal: > Are you sure about this? Doesn't the translation feature always return a translated string?

Not sure, I am digging into this, what I know for sure it caused me some trouble when returning arabic.

ggravlingen: What do you get if you run `print(type("Some text in arabic"))`?

selgamal:
> Contributor



> When you see this:
> 
> ```
> arelle/Cntlr.py:384: error: "Logger" has no attribute "messageCodeFilter"  [attr-defined]
> arelle/Cntlr.py:385: error: "Logger" has no attribute "messageLevelFilter"  [attr-defined]
> arelle/Cntlr.py:386: error: "Handler" has no attribute "logTextMaxLength"  [attr-defined]
> ```
> 
> It means the property isn't defined on the logger or handler class. So you need to go to the logger instance and add eg: `messageCodeFilter: str` or whatever it is (maybe just `Any`?) That's the reason I made this change for instance: https://github.com/Arelle/Arelle/pull/276/files#diff-5aeb72f32f79c477f03e5d73f005abf50d15ae7b168d90e5fd4c544d0558d33cR269

I don't think  its the same, the logger/ logRecord classes are defined in the logging module, the attr can be added for the custom handler.

arelle/typing.py
selgamal: string, but it will depend on where gettext is pulling the lines from

ggravlingen:
> > Contributor
> 
> > When you see this:
> > ```
> > arelle/Cntlr.py:384: error: "Logger" has no attribute "messageCodeFilter"  [attr-defined]
> > arelle/Cntlr.py:385: error: "Logger" has no attribute "messageLevelFilter"  [attr-defined]
> > arelle/Cntlr.py:386: error: "Handler" has no attribute "logTextMaxLength"  [attr-defined]
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > It means the property isn't defined on the logger or handler class. So you need to go to the logger instance and add eg: `messageCodeFilter: str` or whatever it is (maybe just `Any`?) That's the reason I made this change for instance: https://github.com/Arelle/Arelle/pull/276/files#diff-5aeb72f32f79c477f03e5d73f005abf50d15ae7b168d90e5fd4c544d0558d33cR269
> 
> I don't think its the same, the logger/ logRecord classes are defined in the logging module, the attr can be added for the custom handler.

I think there are ways around it but maybe it's not worth the effort at the moment. Maybe just add ignores for those lines?

arelle/typing.py
selgamal: https://docs.python.org/3.9/library/gettext.html#gettext.gettext 
Not sure which functions are the warning refers to.

ggravlingen: You can use this trick to find out what the type is. What I normally do is print the type and then run through a test suite with `-s` at the end. You should get the same results if you parse a stand-alone file with the command line.

arelle/Cntlr.py
selgamal: Got it.
On another note, I think these lines should be removed as a whole, there are for python 2.7, not supported anymore

selgamal:
> 

I think I did a workaround for a similar case, I will give it another shot tomorrow.
stubs may help?

ggravlingen:
> 

Yes, either stubs or ignoring missing imports for the module: https://mypy.readthedocs.io/en/stable/config_file.html#example-pyproject-toml

arelle/Cntlr.py
ggravlingen: Maybe something to consider in this issue? https://github.com/Arelle/Arelle/issues/309

arelle/typing.py
selgamal: I think this more about the po files (translation files) and how they are encoded. Normally for the fallback language (English) it will return the English text as is _("Text here"), so it will always be a string, when you start pulling text from the po files you might get a different type.

selgamal: Return type `Any` confirmed, steps:
```python
# add these lines befor by mypy run (say starting line Cntrl.py:287)
 reveal_type(_("user name if needed (with password) for web file retrieval"))
 self.setUiLanguage('ar-EG', True)
 reveal_type(_("user name if needed (with password) for web file retrieval"))
```
Result:
```bash
arelle/Cntlr.py:287: note: Revealed type is "Any"
arelle/Cntlr.py:289: note: Revealed type is "Any"
```
`print(type(<thing>))` is tricky, because `str` is `Any`

selgamal:
> > 
> 
> Yes, either stubs or ignoring missing imports for the module: https://mypy.readthedocs.io/en/stable/config_file.html#example-pyproject-toml

Actually turns out a simple getatter, setattr, delattr does the trick in this case (when adding custom attributes for objects defined out of scope), and actually it makes sense ...

arelle/Locale.py
selgamal: this done

arelle/Locale.py
selgamal: fixed

arelle/WebCache.py
selgamal: done

arelle/Cntlr.py
selgamal: removed py 2 code fd85d6ba4abe2ad1106765eb64726f490de58ac0

arelle/Locale.py
```diff
@@ -78,7 +80,7 @@ def getLanguageCode():
     except (AttributeError, ValueError): #language code and encoding may be None if their values cannot be determined.
         return "en"
 
-def getLanguageCodes(lang=None):
+def getLanguageCodes(lang: Optional[str] = None) -> list[str]:
```
ggravlingen, line 83:
```suggestion
def getLanguageCodes(lang: str | None = None -> list[str]:
```

arelle/Locale.py
```diff
@@ -10,8 +10,10 @@
 
 (c) Copyright 2011 Mark V Systems Limited, All rights reserved.
 '''
+from __future__ import annotations
 import sys, subprocess
 import regex as re
+from typing import Optional
```
ggravlingen, line 16: Not needed when you're using `from __future__ import annotations`

`Optional[str]` is the same as writing `str | None`

arelle/WebCache.py
```diff
@@ -9,6 +9,7 @@
 
 '''
 from __future__ import annotations
+from typing import TYPE_CHECKING, Optional
```
ggravlingen, line 12: `Optional` is now an unused import?

arelle/Locale.py
selgamal: Yep missed that one

arelle/WebCache.py
selgamal: Yes, I should have done the clean up

arelle/Locale.py
selgamal: yep i missed that one

selgamal: pushed

arelle/WebCache.py
selgamal: done

arelle/Cntlr.py
```diff
@@ -8,18 +8,26 @@
...
+from typing import Any, TYPE_CHECKING, TextIO, Mapping
+from .typing import TypeGetText
 import tempfile, os, io, sys, logging, gettext, json, re, subprocess, math
 from arelle import ModelManager
+from arelle.WebCache import WebCache
 from arelle.Locale import getLanguageCodes, setDisableRTL
 from arelle import PluginManager, PackageManager
 from collections import defaultdict
+
+_: TypeGetText
```
sagesmith-wf, line 21: I don't quite follow what this does

arelle/Cntlr.py
```diff
@@ -8,18 +8,26 @@
    :license: Apache-2.
    :synopsis: Common controller class to initialize for platform and setup common logger functions
 """
-from arelle import PythonUtil # define 2.x or 3.x string types
+from __future__ import annotations
+from typing import Any, TYPE_CHECKING, TextIO, Mapping
+from .typing import TypeGetText
```
sagesmith-wf, line 13: suggest full import path

arelle/Cntlr.py
```diff
@@ -8,18 +8,26 @@
...
 import tempfile, os, io, sys, logging, gettext, json, re, subprocess, math
 from arelle import ModelManager
+from arelle.WebCache import WebCache
```
sagesmith-wf, line 16: I don't know how much drag this stuff has, but should we be putting all of the imports that are just for typing inside of the  `if TYPE_CHECKING` block?

arelle/PackageManager.py
```diff
@@ -25,6 +27,8 @@
 TP_XSD = "http://www.xbrl.org/2016/taxonomy-package.xsd"
 CAT_XSD = "http://www.xbrl.org/2016/taxonomy-package-catalog.xsd"
 
+if TYPE_CHECKING:
+    from .Cntlr import Cntlr
```
sagesmith-wf, line 31: Also suggest full file path here

arelle/Cntlr.py
ggravlingen: I introduced this in another PR. `_` seems to be shorthand for translating a string using gettext so this line adds a type hint stating that `_` is a callable that takes a `str` as an argument and returns a `str` (or `Any` as discussed elsewhere in this PR).

arelle/Cntlr.py
selgamal: Actually this was moved from inside the Cntlr init function, it is needed to initialize WebCache [here](https://github.com/Arelle/Arelle/blob/6e9e22bf1def1d1a7c1411ae53698ede7561e85e/arelle/Cntlr.py#L242)

arelle/ModelManager.py
```diff
@@ -4,11 +4,17 @@
...
+from __future__ import annotations
+from typing import TYPE_CHECKING
 import gc, sys, traceback, logging
 from arelle import ModelXbrl, Validate, DisclosureSystem, PackageManager
 from arelle.PluginManager import pluginClassMethods
 
-def initialize(cntlr):
+
+if TYPE_CHECKING:
+    from .Cntlr import Cntlr
```
austinmatherne-wk, line 15: nit: absolute imports are preferred. We're trying to get away from relative imports.

arelle/PackageManager.py
austinmatherne-wk: Second this. We're trying to move away from relative imports.

austinmatherne-wk (approved):
Nice work! Thank you for taking this on. It's very much appreciated.

sagesmith-wf (approved):
Thank you!

ggravlingen:
Nice!

selgamal:
Should I change this pr to ready for review?

ggravlingen:
Yes

PR #324, opened 2022-09-09
[TABLE LINKBASE] Correct multiples testcases

RobinetDenisAcsone:
#### Reason for change
Table linkbase conformance is still failing

#### Description of change
This PR corrects multiples testcases, please look at each commit message (second line) to see which one of them is related.

Current status:
**_Without the layout validation_** (commented out for the run): 
before: 149 failed cases
after: 93 failed cases
[see xlsx file for more information](https://github.com/Arelle/Arelle/files/9534480/comformance_validation_result.xlsx) (without layout validation)

The failed validations left seems related to:
* tag on rules node (eg. 1200, 1300)
* conceptRelationshipNode (eg. 3110, 3130)
* xbrlte:expressionNotCastableToRequiredType not correctly trowed. (Usually, errors like err:FORG0001 is given)
* some validation not yet implemented

For the majority left, I don't really know yet how to correct it, I will try to work on the layout now, and will come back to it after.

#### Steps to Test
Run the table linkbase conformance suite

**review**:
@hefischer 
@austinmatherne-wk 

For review, I added a small commit for each case, tell me if you want to have all squash before merging.
Also, I am not quite sure for some commit (mostly 8646cd6ebc3268129b18991aa7ced76bfa88abe5, b653f32c2545dc4908e218a5a684ce39bf83d3b5). It seems to work, but for example 8646cd6ebc3268129b18991aa7ced76bfa88abe5 modify code that is used for other things than table linkbase


aviary3-wk:
## Security Insights

## Action Items
---

arelle/ModelRenderingObject.py
```diff
@@ -287,6 +289,10 @@ def siblingBreakdownNode(self):
                     return (sibling,)
         return ()
 
+    @ property
```
austinmatherne-wk, line 292: nit: spacing
```suggestion
    @property
```

RobinetDenisAcsone:
Rebased for tests, not a lot changed in the list yet, since a lot are failing due to the layout comparison (remember that the result I gave was without it).
Once I had time to work on it, the list should be around the number I gave.

austinmatherne-wk (approved):

arelle/ModelDocument.py
```diff
@@ -1164,6 +1164,8 @@ def linkbaseDiscover(self, linkbaseElement, inInstance=False):
                                             linkrole = XbrlConst.defaultLinkRole
                                         #index by both arcrole and linkrole#arcrole and dimensionsions if applicable
                                         baseSetKeys = [(arcrole, linkrole, linkQn, arcQn)]
+                                        baseSetKeys.append((arcrole, linkrole, linkQn, None))
```
sagesmith-wf, line 1167: Pre-existing/unrelated I think it would make it significantly easier to reason about if we made objects for these keys instead of using tuples

arelle/ModelRenderingObject.py
```diff
@@ -18,7 +18,7 @@
                                      aspectsMatch, factsPartitions, VariableBinding)
 from arelle.PrototypeInstanceObject import FactPrototype
 
-OPEN_ASPECT_ENTRY_SURROGATE = '\uDBFF'
+OPEN_ASPECT_ENTRY_SURROGATE = '\uFFEE' # this need to be a utf-8 compatible char
```
sagesmith-wf, line 21: big shrug if its easier, but you could past the actual character here I think its `￮`

sagesmith-wf (approved):

arelle/ModelRenderingObject.py
brettkail-wk: It looks like the intent was to return an arbitrarily "large" code point that would sort after "everything", so a private-use code point was chosen.  If that actually was the intent, then a code point number is definitely clearer than pasting the actual character.

What was the intent of this change?  I don't think U+DBFF actually is larger than all possible code points, but U+FFEE isn't either (there exist U+1xxxx code points too).  Perhaps trying to choose a string that is "arbitrarily large" is the wrong thinking and we instead need to return a high-level type that implements `__lt__` so that some special sentinel can always be sorted to the end.

RobinetDenisAcsone: This character is used as the label and for comparison (== or !=) at some point. I didn't find any comparison with < or > but I may be wrong. The problem I had with the previous char, is that it's put in lxml at some point, and not being an utf-8 character thrown an exception.

I don't think it's shown to the user, it's just for internal comparison.

RobinetDenisAcsone: The exception is thrown here: https://github.com/Arelle/Arelle/blob/88815e6ce4b82d6d5e3f02310c34f05f9b416d5b/arelle/ViewFileRenderedGrid.py#L443

RobinetDenisAcsone: A better fix would perhaps be to change the line to in ViewFileRenderedGrid.py by:
`valueElt.text = xsString(None,None,addQnameValue(self.xmlDoc, aspectValue if label and label != OPEN_ASPECT_ENTRY_SURROGATE else "\u00A0"))
`

It's what is usually done for other labels.
It seems to have the same result at the end.

The only difference is in debug, having '\uDBFF' in a variable make the debugger print exceptions (at least with pycharm)

brettkail-wk: > A better fix would be to change the line

That seems plausible to me.

> The only difference is in debug, having '\uDBFF' in a variable make the debugger print exceptions (at least with pycharm)

Changing the `OPEN_ASPECT_ENTRY_SURROGATE` value to be a non-string class might make that a bit easier and might help flush out other missing comparisons.

RobinetDenisAcsone: I think changing to a non-string class should be good.

@hefischer Can you check before I do the modification? To see if I didn't miss something? Thanks

derekgengenbacher-wf (approved):

arelle/ModelRenderingObject.py
hermfischer: The goal was to differentiate a placeholder for open axes entries so that it was last.  \uDBFF was a bad idea (from long ago for very-ancient python versions).   A new class which subtypes str would work but I'm not sure I'm as comfortable with that as changing \uDBFF to \uF8FF (the highest 2-byte unicode value in the private use area).  (\uFFFFD or \u10FFFD are even-longer byte private use unicode but not as likely to work everywhere.)  So my preference would be \uF8FF (over a new class sub typing str).

brettkail-wk: > A new class which subtypes str would work

Is there a reason the new class has to subtype str?  Can it not just be it's own thing?

> I'm not sure I'm as comfortable with that as changing \uDBFF to \uF8FF

Can you articulate your unease?  Forcing handling at call-sites of this special case at call-sites seems clearer than using a special value (still with special handling at call-sites).

RobinetDenisAcsone: I agree with @brettkail-wk, from what I found, it's only used for some equals, and avoiding using a special value seems to me to be a better practice

PR #339, opened 2022-09-16
Log error and avoid crash on plugin load failure

### LINKED ISSUE #252: Failure to load plugin causes GUI to crash
If the GUI has a plugin enabled that can't be found it will crash at startup with the following error:
> Exception at plug-in method freshenModuleInfos: [Errno 2] No such file or directory: '<repo-path>/arelle/arelle/plugin/<plugin-name>'

This can happen if you have used one of the pre-built distributions of arelle to enable a plugin that's not part of the core arelle repo (EdgarRenderer) and then later try to launch the arelle GUI from within the repo (`python arelleGUI.pyw`).

If the `arelleCmdLine` can't find a plugin it will log an error, but continue processing. The GUI should gracefully handle missing plugins in the same manner.

aaroncameron-wk:
#### Reason for change
Closes #252 

#### Description of change
FileNotFoundError was not being captured/logged on plug-in execution exception like other exception types.
Also consolidated and updated logging around plug-in initialization.

#### Steps to Test
- Move, rename, or otherwise make unfindable a non-repo plugin (like EdgarRenderer) referenced in `plugins.json`
- Launch `arelleGUI.pyw` from repo
- Should launch successfully with "Unable to load module ____" visible in messages window

**review**:
@Arelle/arelle

### DIFF
--- a/arelle/PluginManager.py
+++ b/arelle/PluginManager.py
@@ -447,7 +447,7 @@ def loadModule(moduleInfo: dict[str, Any], packagePrefix: str="") -> None:
                     logPluginTrace(_msg, error=True)
             for importModuleInfo in moduleInfo.get('imports', EMPTYLIST):
                 loadModule(importModuleInfo, packageImportPrefix)
-        except (AttributeError, ImportError, ModuleNotFoundError, TypeError, SystemError) as err:
+        except (AttributeError, ImportError, FileNotFoundError, ModuleNotFoundError, TypeError, SystemError) as err:
             # Send a summary of the error to the logger and retain the stacktrace for stderr
             _cntlr.addToLog(message=_ERROR_MESSAGE_IMPORT_TEMPLATE.format(name), level=logging.ERROR)
 
--- a/arelle/PluginManager.py
+++ b/arelle/PluginManager.py
@@ -157,14 +157,14 @@ moduleInfo = {
 def logPluginTrace(message: str, level: Number) -> None:
     """
     If plugin trace file logging is configured, logs `message` to it.
-    Otherwise, only logs to controller logger if log is an error.
+    Only logs to controller logger if log is an error.
     :param message: Message to be logged
     :param level: Log level of message (e.g. logging.INFO)
     """
     global pluginTraceFileLogger
     if pluginTraceFileLogger:
         pluginTraceFileLogger.log(level, message)
-    elif level >= logging.ERROR:
+    if level >= logging.ERROR:
         _cntlr.addToLog(message=message, level=level, messageCode='arelle:pluginLoadingError')
 
 
--- a/arelle/PluginManager.py
+++ b/arelle/PluginManager.py
@@ -31,11 +31,13 @@ if TYPE_CHECKING:
 
 PLUGIN_TRACE_FILE = None
 # PLUGIN_TRACE_FILE = "c:/temp/pluginerr.txt"
+PLUGIN_TRACE_LEVEL = logging.WARNING
 
 # plugin control is static to correspond to statically loaded modules
 pluginJsonFile = None
 pluginConfig = None
 pluginConfigChanged = False
+pluginLogger = None
 modulePluginInfos = {}
 pluginMethodsForClasses = {}
 _cntlr = None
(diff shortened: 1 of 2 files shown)


aviary3-wk:
## Security Insights

## Action Items
---

arelle/PluginManager.py
```diff
@@ -143,25 +143,37 @@ def close():  # close all loaded methods
 
 '''
 
+def logPluginTrace(message: str, error: bool=False) -> None:
+    """
+    If PLUGIN_TRACE_FILE is configured, logs `message` to it.
```
sagesmith-wf, line 148: Suggest using standard docstring syntax

arelle/PluginManager.py
```diff
@@ -143,25 +143,37 @@ def close():  # close all loaded methods
...
+    Otherwise, only prints to `stderr` if `error` flag is `True`
+    """
+    if PLUGIN_TRACE_FILE:
+        with open(PLUGIN_TRACE_FILE, "at", encoding='utf-8') as fh:
+            fh.write(message + '\n')
```
sagesmith-wf, line 153: suggest either breaking into 2 lines or using a formatted string to avoid an exception if message is `None`

sagesmith-wf: Also might be helpful to encode some general diagnostic info like time and thread ect. Suggest making a logger that either writes to a file or to std out instead of explicitly writing to a file. Maybe:

```python
import logging
import sys
...
logger = setup_logging(logging.StreamHandler(sys.stdout) if not PLUGIN_TRACE_FILE else logging.FileHandler(PLUGIN_TRACE_FILE))
...

def setup_logging(handler, formatter=logging.Formatter('%(asctime)s %(message)s', datefmt='%m/%d/%Y %I:%M:%S %p')):
    handler.setLevel(PLUGIN_TRACE_LEVEL)
    handler.setFormatter(formatter)
    logger = logging.getLogger(f'{__name__})
    logger.addHandler(handler)
    return logger
...
def logPluginTrace(message: str, level) -> None:
    logger.log(message, level)
```

arelle/PluginManager.py
```diff
@@ -104,6 +107,26 @@ def close():  # close all loaded methods
...
+
+def setupLogger(handler: logging.Handler) -> logging.Logger:
+    """
+    Logs message with given level to the configured plugin logger
+    :param message: Message to be logged
+    :type message: str
+    :param level: Log level of message (e.g. logging.INFO)
+    :type level: Number
+    """
+    logger = logging.getLogger(f'{__name__}')
```
sagesmith-wf, line 119: think the `f'{}'` is unnecessary and you can just do `getLogger(__name__)`

sagesmith-wf (dismissed):

sagesmith-wf (approved):

arelle/PluginManager.py
```diff
@@ -44,7 +46,8 @@
 _ERROR_MESSAGE_IMPORT_TEMPLATE = "Unable to load module {}"
 
 def init(cntlr: Cntlr, loadPluginConfig: bool = True) -> None:
-    global pluginJsonFile, pluginConfig, modulePluginInfos, pluginMethodsForClasses, pluginConfigChanged, _cntlr, _pluginBase
+    global pluginJsonFile, pluginConfig, pluginLogger, modulePluginInfos, pluginMethodsForClasses, pluginConfigChanged, _cntlr, _pluginBase
+    pluginLogger = setupLogger(logging.FileHandler(PLUGIN_TRACE_FILE) if PLUGIN_TRACE_FILE else logging.StreamHandler(sys.stdout))
```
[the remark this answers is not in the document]
aaroncameron-wk: This logger does not output to GUI, that happens through `cntlr.addToLog` when plugin loading fails [here](https://github.com/Arelle/Arelle/blob/319cda0c6a1baf1d6f5141e6ac247661d7a56303/arelle/PluginManager.py#L453). 
This logger was only implemented in place of the direct `print` calls and trace file writes that this PR ended up touching. I did experiment with using `cntlr.addToLog` as well as having this logger propagate to the cntrl logger, but that logger does not output timestamps and is not actually initialized by the time the first attempts are made at loading plugin info.
For now, it is probably better to just use the existing `cntlr.addToLog` and only use the separate logger instance for the plugin trace file logging, but I think a review of the logging setup as a whole would be valuable.

arelle/PluginManager.py
```diff
@@ -143,25 +153,41 @@ def close():  # close all loaded methods
...
+    """
+    If plugin trace file logging is configured, logs `message` to it.
+    Otherwise, only logs to controller logger if log is an error.
+    :param message: Message to be logged
+    :param level: Log level of message (e.g. logging.INFO)
+    """
+    global pluginTraceFileLogger
+    if pluginTraceFileLogger:
+        pluginTraceFileLogger.log(level, message)
+    elif level >= logging.ERROR:
```
austinmatherne-wk, line 167: I don't think the plugin trace file should cause errors to be removed from the cntrl. I'd make this an `if` instead of an `elif`.

austinmatherne-wk (approved):

allenbross-wf (approved):

sagesmith-wf (approved):

derekgengenbacher-wf (approved):

PR #351, opened 2022-09-20
XT-2345: Removing Python 2.7 references

allenbross-wf:
#### Reason for change
Python 2.7 is no longer supported

#### Description of change
1. Removed conditional 2.7 code
2. Removed/migrated Builtin modifications

#### Steps to Test
Conformance suite should be sufficient.

**review**:
@Arelle/arelle


aviary3-wk:
## Security Insights

## Action Items
---

arelle/ViewFile.py
```diff
@@ -8,12 +8,10 @@
...
 from arelle.FileSource import FileNamedStringIO
-if sys.version[0] >= '3':
-    csvOpenMode = 'w'
-    csvOpenNewline = ''
-else:
-    csvOpenMode = 'wb' # for 2.7
-    csvOpenNewline = None
+
+csvOpenMode = 'w'
+csvOpenNewline = ''
```
brettkail-wk, line 13: For these single use globals that only existed to gloss over 2-vs-3, it seems like it would be clearer to inline the values and remove the values altogether.

austinmatherne-wk (commented):
Nice work! 
Mostly just a bunch of minor nits. `str(“some string”)` is equivalent to `”some string”` so with the drop of python 2 support we can remove a lot of the wrapping `str()` calls. I’ve pointed these out.

There’s still an instance of `_RANGE` that needs to be replaced [here](https://github.com/Arelle/Arelle/blob/29ba43a3452b17e87b1fe6862a7041d3697e3148/arelle/XPathContext.py#L479).

austinmatherne-wk (approved):

derekgengenbacher-wf (approved):

derekgengenbacher-wf:
QA +10 CI passes successfully. Smoke testing locally was successful.

Issue #356, opened 2022-09-23
Run CI tests on Linux, macOS, and Windows


austinmatherne-wk:
### What should we change and why?

We ran into an issue (#354) where local typing (on macOS) didn't match the results in CI (Linux). We should update the [unit test/type checking/linting workflow](https://github.com/Arelle/Arelle/blob/cb10d2365fa4bfeb389f45275b6ac3f3257fb830/.github/workflows/tests.yml#L9) to run with a matrix of the latest macOS, Ubuntu, and Windows runners.

Initial investigation shows some issues with type checking and unit tests with the macOS runner that will need to be resolved.

PR #358, opened 2022-09-23
Add OS test matrix

ggravlingen:
#### Reason for change
https://github.com/Arelle/Arelle/issues/356

#### Description of change
Adds a matrix to run tests on different os versions.

#### Steps to Test

**review**:
@Arelle/arelle

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -6,7 +6,13 @@ on:
 
 jobs:
   tests:
-    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        os:
+          - macos-12
+          - ubuntu-latest
+          - windows-latest
+    runs-on: ${{ matrix.os }}
     steps:
       - uses: actions/checkout@v3.0.2
       - name: Install Python 3


.github/workflows/tests.yml
```diff
@@ -6,7 +6,13 @@ on:
 
 jobs:
   tests:
-    runs-on: ubuntu-latest
+    strategy:
```
ggravlingen, line 9: I've never done this before and got a bit curious how this works so I thought I'd give it a go. Let's see how it works out.

aviary3-wk:
## Security Insights

## Action Items
---

.github/workflows/tests.yml
```diff
@@ -6,7 +6,13 @@ on:
...
+      matrix:
+        os:
+          - macos-12
```
austinmatherne-wk, line 12: Should probably use latest for consistency:
```suggestion
          - macos-latest
```

ggravlingen: Good point, I copied it from one of the other workflows.

ggravlingen: When I think about it: maybe pin and be a bit more precise to the users on what OS versions this lib supports?

ggravlingen: Tests for Macos fail due to some lib missing. Did a bit of reading on how to solve it and maybe using `latest` is a solution here. Will give it a try.

ggravlingen: That seems to have done the trick 🎉

tests/unit_tests/arelle/test_pluginmanager.py
```diff
@@ -89,14 +89,14 @@ def test_plugin_manager_reset():
             # Test data
             ("tests/unit_tests/arelle", "functionsMaths", "xyz"),
             # Expected result
-            ("functionsMaths", "tests/unit_tests", "xyz")
+            ("functionsMaths", f"tests{os.sep}unit_tests", "xyz")
```
austinmatherne-wk, line 92: nit: I think using [os.path.normcase()](https://docs.python.org/3/library/os.path.html?highlight=path#os.path.normcase) to consume these paths in the test body would be a little more resilient (filesystem capitalization support, etc) to any potential future changes.

austinmatherne-wk (approved):

tests/unit_tests/arelle/test_pluginmanager.py
ggravlingen: You mean wrap the whole string of the exected result in that function?

austinmatherne-wk: Yes. I don't feel strongly though. I also think the `test_data` paths should be normalized to correctly exercise the code under test on windows.

ggravlingen: OK, I'll leave it to be then. `os.sep` is used alot in Arelle so maybe an issue should be raised to replace all of them at once?

ggravlingen: @stevenbronson-wk sorry, I don't quite understand what you mean here. Where should that go?

austinmatherne-wk (approved):

Issue #320, opened 2022-09-06
Easy environment reporting for bugs


austinmatherne-wk:
[As suggested](https://github.com/Arelle/Arelle/pull/319#issuecomment-1238354380) by @ggravlingen  we should create a utility from the CLI to facilitate copy pasting environment information (or uploading a diagnostics file) for bug reports.

sagesmith-wf:
This is a super good idea!

PR #360, opened 2022-09-24
Add initial support for system info reporting

ggravlingen:
#### Reason for change
As discussed in https://github.com/Arelle/Arelle/issues/320.

It doesn't solve the entire issue, but at least it's a start to be amended with additional information as needed.

#### Description of change
Adding a function to return some basic system information. I threw in a trivial unit test as well, just checking that the function returns. Given that the test is run on Mac, Windows and Linux it feels like an opportunity to create a flaky test to start checking what actual system values are returned by the function.

#### Steps to Test
```python -m arelleCmdLine --diagnostics```

Should return something like this:
```
{
'arch': 'aarch64',
 'docker': True,
 'os_name': 'Linux',
 'version_arelle': '1.2020.05.10',
 'version_os': '5.10.104-linuxkit',
 'version_python': '3.9.13',
 'virtualenv': True
}
```

**review**:
@Arelle/arelle

### DIFF
--- a/tests/unit_tests/arelle/test_system_info.py
+++ b/tests/unit_tests/arelle/test_system_info.py
@@ -1,0 +1,22 @@
+"""Tests for system info."""
+from arelle.SystemInfo import get_system_info
+
+
+def test_get_system_info() -> None:
+    """Test that function get_system_info returns."""
+    function_result = get_system_info()
+
+    expected_keys = {
+        "arch",
+        "docker",
+        "os_name",
+        "version_arelle",
+        "version_os",
+        "version_python",
+        "virtualenv"
+    }
+
+    resulting_keys = set(function_result.keys())
+
+    missing_keys = expected_keys - resulting_keys
+    assert len(missing_keys) == 0, f"Missing keys {missing_keys}"
--- a/arelle/CntlrCmdLine.py
+++ b/arelle/CntlrCmdLine.py
@@ -24,6 +24,8 @@ from arelle import PluginManager
 from arelle.PluginManager import pluginClassMethods
 from arelle.UrlUtil import isHttpUrl
 from arelle.WebCache import proxyTuple
+from arelle.SystemInfo import get_system_info
+from pprint import pprint
 import logging
 from lxml import etree
 win32file = win32api = win32process = pywintypes = None
@@ -382,6 +384,8 @@ def parseAndRun(args):
     parser.add_option("-a", "--about",
                       action="store_true", dest="about",
                       help=_("Show product version, copyright, and license."))
+    parser.add_option("--diagnostics", action="store_true", dest="diagnostics",
+                      help=_("output system diagnostics information"))
 
     if not args and cntlr.isGAE:
         args = ["--webserver=::gae"]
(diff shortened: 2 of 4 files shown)

### CODE REVIEW

**Comment by aviary2-wf** [comment]
## Security Insights

## Action Items
---

File: tests/unit_tests/arelle/test_system_info.py [hunk 1]
```diff
@@ -0,0 +1,7 @@
+"""Tests for system info."""
+from arelle.SystemInfo import get_system_info
+
+def test_get_system_info() -> None:
+    """Test that function get_system_info returns."""
+    get_system_info()
```
sagesmith-wf, line 6: Suggest verifying the expected keys are present

sagesmith-wf (dismissed):

tests/unit_tests/arelle/test_system_info.py
ggravlingen: Added in https://github.com/Arelle/Arelle/pull/360/commits/f63f58a818f177df61f8939e33c36f76ddf36ffd

tests/unit_tests/arelle/test_system_info.py
```diff
@@ -0,0 +1,18 @@
+"""Tests for system info."""
+from arelle.SystemInfo import get_system_info
+
+
+def test_get_system_info() -> None:
+    """Test that function get_system_info returns."""
+    function_result = get_system_info()
+
+    assert True
+    assert list(function_result.keys()) == [
```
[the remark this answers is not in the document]
ggravlingen: Sorry, didn't see this while I was trying to solve the issue. I hope the fix I just pushed works, else I'll go for your solution here.

ggravlingen:
I have intentionally not added anything to the issue template as I don't want to fiddle too much with your internal preferred methods of handling issues.

tests/unit_tests/arelle/test_system_info.py
```diff
@@ -0,0 +1,18 @@
+"""Tests for system info."""
+from arelle.SystemInfo import get_system_info
+
+
+def test_get_system_info() -> None:
+    """Test that function get_system_info returns."""
+    function_result = get_system_info()
+
+    assert True
```
sagesmith-wf, line 9: Remove?

tests/unit_tests/arelle/test_system_info.py
```diff
@@ -0,0 +1,18 @@
...
+    assert any(test_key in list(function_result.keys()) for test_key in [
+        "arch",
+        "docker",
+        "os_name",
+        "version_arelle",
+        "version_os",
+        "version_python",
+        "virtualenv",
+    ]) is True
```
sagesmith-wf, line 18: The `is True` is implicitly there so you can just remove it. You might restructure so that you can print out which one is there that isn't expected. I might restructure like:

```python
expected_keys = {
    "arch", 
    "docker", 
    "os_name", 
    "version_arelle", 
    "version_os", 
    "version_python", 
    "virtualenv"
}

resulting_keys = set(function_result.keys())

missing_keys = expected_keys - resulting_keys
assert len(missing_keys) == 0, f"Missing keys {missing_keys}"

extra_keys = resulting_keys - expected_keys
assert len(extra_keys) == 0, f"Extra keys {extra_keys}"
```

ggravlingen: Neat solution! Haven't used sets much in Python but this was a cool way of using them.

ggravlingen: I'll have to remove the last bit as an additional key is added on Mac environments. The resulting test should be good enough for this small function.

sagesmith-wf (approved):
Really appreciate the unit test, I think getting in the habit is going to be really sweet long term!

austinmatherne-wk (approved):
Thank you for writing a test!

PR #391, opened 2022-10-03
Adding more Type Hinting to ModelXbrl

allenbross-wf:
#### Reason for change
 
ModelXbrl was not fully type hinted.

#### Description of change

Added type hints

#### Steps to Test

CI should be sufficient

**review**:
@Arelle/arelle

### DIFF
--- a/arelle/FileSource.py
+++ b/arelle/FileSource.py
@@ -76,7 +76,7 @@ def openFileSource(
         # not archived content
         return FileSource(filename, cntlr, checkIfXmlIsEis)
 
-def archiveFilenameParts(filename: str, checkIfXmlIsEis: bool = False) -> tuple[str, str] | None:
+def archiveFilenameParts(filename: str | None, checkIfXmlIsEis: bool = False) -> tuple[str, str] | None:
     # check if path has an archive file plus appended in-archive content reference
     for archiveSep in archivePathSeparators:
         if (filename and
@@ -407,11 +407,12 @@ class FileSource:
                 return [f]  # standard package
         return [f for f in (self.dir or []) if os.path.split(f)[-1] in TAXONOMY_PACKAGE_FILE_NAMES]
 
-    def isInArchive(self, filepath: str, checkExistence: bool = False) -> bool:
+    def isInArchive(self, filepath: str | None, checkExistence: bool = False) -> bool:
         archiveFileSource = self.fileSourceContainingFilepath(filepath)
         if archiveFileSource is None:
             return False
         if checkExistence:
+            assert isinstance(filepath, str)
             assert isinstance(archiveFileSource.basefile, str)
             assert archiveFileSource.dir is not None
             archiveFileName = filepath[len(archiveFileSource.basefile) + 1:].replace("\\", "/") # must be / file separators
@@ -432,7 +433,7 @@ class FileSource:
                     break
         return url
 
-    def fileSourceContainingFilepath(self, filepath: str) -> FileSource | None:
+    def fileSourceContainingFilepath(self, filepath: str | None) -> FileSource | None:
         if self.isOpen:
             # archiveFiles = self.dir
             ''' change to return file source if archive would be in there (vs actually is in archive)
@@ -442,6 +443,7 @@ class FileSource:
                  filepath[len(self.baseurl) + 1:] in archiveFiles)):
                 return self
             '''
+            assert isinstance(filepath, str)
             assert isinstance(self.basefile, str)
             assert isinstance(self.baseurl, str)
             if (filepath.startswith(self.basefile) or
--- a/arelle/ModelObject.py
+++ b/arelle/ModelObject.py
@@ -3,7 +3,7 @@ See COPYRIGHT.md for copyright information.
 '''
 from __future__ import annotations
 from typing import TYPE_CHECKING, Any, Generator, Optional, cast
-from lxml import etree  # type: ignore[import]
+from lxml import etree
 from arelle import Locale
 from arelle.ModelValue import qname, qnameEltPfxName, QName
 
@@ -174,7 +174,7 @@ class ModelObject(etree.ElementBase):
         return attrValue
 
     @property
-    def localName(self) -> str:
+    def localName(self) -> Any:
         try:
             return self._localName
         except AttributeError:
@@ -182,7 +182,7 @@ class ModelObject(etree.ElementBase):
             return self._localName
 
     @property
-    def prefixedName(self) -> str:
+    def prefixedName(self) -> Any:
         try:
             return self._prefixedName
         except AttributeError:
@@ -190,7 +190,7 @@ class ModelObject(etree.ElementBase):
             return self._prefixedName
 
     @property
-    def namespaceURI(self) -> str | None:
+    def namespaceURI(self) -> Any | None:
         try:
             return self._namespaceURI
         except AttributeError:
@@ -249,12 +249,12 @@ class ModelObject(etree.ElementBase):
             return self._elementSequence
 
     @property
-    def parentQname(self) -> QName | None:
+    def parentQname(self: ModelObject) -> QName | None:
         try:
             return self._parentQname
         except AttributeError:
             parentObj = self.getparent()
-            self._parentQname = parentObj.elementQname if parentObj is not None else None
+            self._parentQname = parentObj.elementQname if parentObj is not None else None  # type: ignore[attr-defined]
             return self._parentQname
 
 
@@ -453,6 +453,7 @@ class ModelAttribute:
 
 class ObjectPropertyViewWrapper:  # extraProperties = ( (p1, v1), (p2, v2), ... )
     __slots__ = ("modelObject", "extraProperties")
+    modelObject: ModelObject
     def __init__(self, modelObject: ModelObject, extraProperties: tuple[Any, ...] = ()) -> None:
         self.modelObject = modelObject
         self.extraProperties = extraProperties
--- a/arelle/ModelObject.py
+++ b/arelle/ModelObject.py
@@ -149,7 +149,7 @@ class ModelObject(etree.ElementBase):
         except AttributeError:
             return None
 
-    def attr(self, attrname: str) -> Any:
+    def attr(self, attrname: str) -> str | None:
         return self.get(attrname)
 
     @property
@@ -167,14 +167,14 @@ class ModelObject(etree.ElementBase):
         else:
             self._prefixedName = self.localName
 
-    def getStripped(self, attrName: str) -> Any:
+    def getStripped(self, attrName: str) -> str | None:
         attrValue = self.get(attrName)
         if attrValue is not None:
             return attrValue.strip()
         return attrValue
 
     @property
-    def localName(self) -> Any:
+    def localName(self) -> str:
         try:
             return self._localName
         except AttributeError:
@@ -182,7 +182,7 @@ class ModelObject(etree.ElementBase):
             return self._localName
 
     @property
-    def prefixedName(self) -> Any:
+    def prefixedName(self) -> str:
         try:
             return self._prefixedName
         except AttributeError:
@@ -190,7 +190,7 @@ class ModelObject(etree.ElementBase):
             return self._prefixedName
 
     @property
-    def namespaceURI(self) -> Any | None:
+    def namespaceURI(self) -> str | None:
         try:
             return self._namespaceURI
         except AttributeError:
--- a/arelle/ModelObject.py
+++ b/arelle/ModelObject.py
@@ -120,7 +120,7 @@ class ModelObject(etree.ElementBase):
         self.isChanged = False
         parent = self.getparent()
         if parent is not None and hasattr(parent, "modelDocument"):
-            self.init(parent.modelDocument)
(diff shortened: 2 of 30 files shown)


aviary-wf:
## Security Insights

## Action Items
---

ggravlingen:
@allenbross-wf you probably want to add the file to `pyproject.toml` for mypy to pick it up in the CI/CD.

arelle/ModelXbrl.py
```diff
@@ -22,35 +22,34 @@
...
-AUTO_LOCATE_ELEMENT = '771407c0-1d0c-11e1-be5e-028037ec0200' # singleton meaning choose best location for new element
-DEFAULT = sys.intern(str("default"))
-NONDEFAULT = sys.intern(str("non-default"))
-DEFAULTorNONDEFAULT = sys.intern(str("default-or-non-default"))
-EMPTY_TUPLE = ()
+AUTO_LOCATE_ELEMENT: str = '771407c0-1d0c-11e1-be5e-028037ec0200' # singleton meaning choose best location for new element
+DEFAULT: str = sys.intern(str("default"))
+NONDEFAULT: str = sys.intern(str("non-default"))
+DEFAULTorNONDEFAULT: str = sys.intern(str("default-or-non-default"))
+EMPTY_TUPLE: Tuple[()]
```
austinmatherne-wk, line 40: This is a breaking change. `EMPTY_TUPLE` is no longer a defined value (also `typing.Tuple[x]` is deprecated, you should use `tuple[x]` instead).

arelle/ModelXbrl.py
```diff
@@ -22,35 +22,34 @@
...
+    from arelle.ModelValue import QName, DATETIME
+    from datetime import date, datetime
+    from arelle import ModelFormulaObject, ModelManager
+
 
 
 ModelFact = None
 
-profileStatNumber = 0
+profileStatNumber: int = 0
```
austinmatherne-wk, line 34: These literal values don't need to be typed. mypy can infer `x = 0` is an int or `x = 'foo'` is a string, etc.

austinmatherne-wk (commented):
If you add `arelle.ModelXbrl` to the [pyproject.toml list of type checking enabled modules](https://github.com/Arelle/Arelle/blob/34d846eff1b3df8c79dbf95dd86325eb4ca1892b/pyproject.toml#L108-L122) and run `mypy arelle --namespace-packages` you should get a list of type errors that need to be resolved.

Feel free to reach out if you'd like to work through some of these together. 🙂

arelle/ModelXbrl.py
```diff
@@ -108,6 +111,7 @@ def create(modelManager, newDocumentType=None, url=None, schemaRefs=None, create
                 loadSchemalocatedSchemas(modelXbrl)
     return modelXbrl
 
+
 def loadSchemalocatedSchemas(modelXbrl) -> None:
     from arelle import ModelDocument
```
selgamal, line 116:
```suggestion
    from arelle.ModelDocument import ModelDocument
```
This is not a change from this PR, but it affects lines 119 and 122

arelle/FileSource.py
```diff
@@ -104,7 +105,7 @@ def __init__(self, fileSource, errno, fileName):
         self.url = fileSource.url
 
 class FileSource:
-    def __init__(self, url, cntlr=None, checkIfXmlIsEis=False):
+    def __init__(self, url, cntlr=None, checkIfXmlIsEis=False) ->Any:
```
sagesmith-wf, line 108: Nit: Space between `-> Any`

arelle/ModelObject.py
```diff
@@ -149,7 +149,7 @@ def modelXbrl(self) -> ModelXbrl | None:
         except AttributeError:
             return None
 
-    def attr(self, attrname: str) -> str | None:
+    def attr(self, attrname: str) -> Any:
```
austinmatherne-wk, line 152: Change not needed?
```suggestion
    def attr(self, attrname: str) -> str | None:
```

arelle/ModelObject.py
```diff
@@ -167,30 +167,30 @@ def setNamespaceLocalName(self) -> None:
...
-    def getStripped(self, attrName: str) -> str | None:
+    def getStripped(self, attrName: str) -> Any:
         attrValue = self.get(attrName)
         if attrValue is not None:
             return attrValue.strip()
         return attrValue
 
     @property
-    def localName(self) -> str:
+    def localName(self) -> Any:
```
austinmatherne-wk, line 177: Change not needed?
```suggestion
    def localName(self) -> str:
```

arelle/ModelObject.py
```diff
@@ -167,30 +167,30 @@ def setNamespaceLocalName(self) -> None:
...
         try:
             return self._localName
         except AttributeError:
             self.setNamespaceLocalName()
             return self._localName
 
     @property
-    def prefixedName(self) -> str:
+    def prefixedName(self) -> Any:
```
austinmatherne-wk, line 185: Change not needed?
```suggestion
    def prefixedName(self) -> str:
```

arelle/ModelObject.py
```diff
@@ -167,30 +167,30 @@ def setNamespaceLocalName(self) -> None:
...
         try:
             return self._prefixedName
         except AttributeError:
             self.setNamespaceLocalName()
             return self._prefixedName
 
     @property
-    def namespaceURI(self) -> str | None:
+    def namespaceURI(self) -> Any | None:
```
austinmatherne-wk, line 193: Change not needed?
```suggestion
    def namespaceURI(self) -> str | None:
```

arelle/PrototypeInstanceObject.py
```diff
@@ -4,10 +4,11 @@
 from arelle import XmlUtil
 from arelle.ModelValue import QName
 from arelle.ModelObject import ModelObject
+from typing import Any
 Aspect = None
 
 class FactPrototype():      # behaves like a fact for dimensional validity testing
-    def __init__(self, v, aspectValues=None):
+    def __init__(self, v, aspectValues=None) -> Any:
```
austinmatherne-wk, line 11: `__init__` always returns `None`:
```suggestion
    def __init__(self, v, aspectValues=None) -> None:
```

arelle/PythonUtil.py
```diff
@@ -4,11 +4,14 @@
 
 do not convert 3 to 2
 '''
+from __future__ import annotations
 import sys
 from decimal import Decimal
 from fractions import Fraction
 from collections import OrderedDict
 from collections.abc import MappingView, MutableSet
+import math
```
austinmatherne-wk, line 13: Unused import?

austinmatherne-wk (commented):
Good work. There was a lot to do here. These are my initial comments. I'll have more once I've gone through all of the model xbrl changes.

arelle/ModelXbrl.py
```diff
@@ -266,73 +191,82 @@ class ModelXbrl:
...
-        self.logger = logging.getLogger("arelle")
-        self.logRefObjectProperties = getattr(self.logger, "logRefObjectProperties", False)
-        self.logRefHasPluginAttrs = any(True for m in pluginClassMethods("Logging.Ref.Attributes"))
-        self.logRefHasPluginProperties = any(True for m in pluginClassMethods("Logging.Ref.Properties"))
-        self.logRefFileRelUris = defaultdict(dict)
-        self.profileStats = {}
-        self.schemaDocsToValidate = set()
-        self.modelXbrl = self # for consistency in addressing modelXbrl
-        self.arelleUnitTests = {} # unit test entries (usually from processing instructions
+        self.logger: Any = logging.getLogger("arelle")
```
austinmatherne-wk, line 251: This is an example of where type information is lost. `logging.getLogger(...)` already has a return value, but it's discarded by the `Any` annotation.

```suggestion
        self.logger = logging.getLogger("arelle")
```

arelle/ModelXbrl.py
```diff
@@ -266,73 +191,82 @@ class ModelXbrl:
...
+        self.logRefObjectProperties: Any = getattr(self.logger, "logRefObjectProperties", False)
+        self.logRefHasPluginAttrs: bool = any(True for m in pluginClassMethods("Logging.Ref.Attributes"))
+        self.logRefHasPluginProperties: bool = any(True for m in pluginClassMethods("Logging.Ref.Properties"))
+        self.logRefFileRelUris: DefaultDict[Any, dict[str, str]] = defaultdict(dict)
```
austinmatherne-wk, line 255: When we define types we're making a contract with consumers of how the field/function/variable will behave (in other languages like Java this would be an interface). By defining the specific `defaultdict` implementation of `dict` as the type it makes it difficult to later change the implementation if we want to. It's almost always better to use the more generic type and retain the flexibility to change the implementation if needed.

```suggestion
        self.logRefFileRelUris: dict[Any, dict[str, str]] = defaultdict(dict)
```

arelle/ModelXbrl.py
```diff
@@ -266,73 +191,82 @@ class ModelXbrl:
...
+        self.profileStats: dict[str, tuple[int, Any, Any]] = {}
+        self.schemaDocsToValidate: set[modeldoc] = set()
+        self.closeFileSource: bool
+        self.fileSource: Any
+        self.entryLoadingUrl: Any
+        self.ixdsDocUrls: list[str]
```
austinmatherne-wk, line 261: These uninitialized class field types should be defined up at the class level (where the existing `contexts` and `dimensionDefaultConcepts` fields are)

arelle/ModelXbrl.py
```diff
@@ -379,69 +311,61 @@ def closeViews(self):
...
-        :type linkrole: str
         :param arcqname: Arc element qname (wild if None)
-        :type arcqname: QName
         :param includeProhibits: True to include prohibiting arc elements as relationships
-        :type includeProhibits: bool
-        :returns: [ModelRelationship] -- Ordered list of effective relationship objects per parameters
         """
         from arelle import ModelRelationshipSet
-        key = (arcrole, linkrole, linkqname, arcqname, includeProhibits)
+        key: tuple[str, tuple[str] | str | None, QName | None, QName | None, bool] = (arcrole, linkrole, linkqname, arcqname, includeProhibits)
```
austinmatherne-wk, line 333: `arcrole`, `linkrole`, etc are all already types, so the type of `key` can be inferred:
```suggestion
        key = (arcrole, linkrole, linkqname, arcqname, includeProhibits)
```

austinmatherne-wk (commented):
Some additional comments for ModelXbrl

arelle/ModelXbrl.py
austinmatherne-wk: Good catch. I didn't look closely enough.

arelle/ModelXbrl.py
```diff
@@ -771,51 +671,49 @@ def factsByQname(self): # indexed by fact (concept) qname
...
         :param periodType: Period type to match ("instant", "duration", or "forever")
-        :type periodType: str
-        :returns: set -- ModelFacts that have specified periodType
         """
         try:
             return self._factsByPeriodType[periodType]
         except AttributeError:
-            self._factsByPeriodType = fbpt = defaultdict(set)
+            self._factsByPeriodType: defaultdict[str, set[ModelFact]] = defaultdict(set)
+            fbpt: defaultdict[str,set[ModelFact]] = defaultdict(set)
```
austinmatherne-wk, line 716: This is a breaking change. `self._factsByPeriodType` and the local `fbpt` are no longer the same var. You're populating `fbpt`, but returning values from `self._factsByPeriodType`.

arelle/ModelXbrl.py
```diff
@@ -455,26 +379,26 @@ def matchSubstitutionGroup(self, elementQname, subsGrpMatchTable):
...
-                  qn:(qn is not None) for qn in (subsGrpQnames if hasattr(subsGrpQnames, '__iter__') else (subsGrpQnames,)) + (None,)})
 
-    def createInstance(self, url=None):
-        """Creates an instance document for a DTS which didn't have an instance document, such as
+        if isinstance(subsGrpQnames, list):
+            qnames = subsGrpQnames
+        else:
+            qnames = [cast('QName', subsGrpQnames)]
+        return cast('Optional[bool]', self.matchSubstitutionGroup(elementQname, {
+                  qn: (qn is not None) for qn in qnames}))
```
austinmatherne-wk, line 395: This is also a breaking change (causing table linkbase conformance suite failures). Previously this function would handle any iterable (`hasattr(subsGrpQnames, '__iter__')`), but now only accommodates lists. Additionally, the `+ (None,)` bit was gross, but important because `matchSubstitutionGroup()` falls back to looking up `None`. A better approach would be to explicitly handle `None` here.

```suggestion
        from collections.abc import Iterable
        if isinstance(subsGrpQnames, Iterable):
            qnames = subsGrpQnames
        else:
            qnames = [subsGrpQnames]
        matchingSubstitutionGroup = cast('Optional[bool]', self.matchSubstitutionGroup(elementQname, {
                  qn: (qn is not None) for qn in qnames}))
        return matchingSubstitutionGroup if matchingSubstitutionGroup is not None else False
```

austinmatherne-wk (approved):

derekgengenbacher-wf:
QA +10 Testing locally succeeded. CI.

PR #423, opened 2022-10-15
Correct gYearMonth, gMonthDay, gYear, gMonth, and gDay types

austinmatherne-wk:
#### Reason for change
`gYearMonth`, `gMonthDay`, `gYear`, `gMonth`, and `gDay` classes can accept `str` init params.

#### Description of change
Type updates

#### Steps to Test
* CI/mypy

**review**:
@Arelle/arelle

### DIFF
--- a/arelle/ModelValue.py
+++ b/arelle/ModelValue.py
@@ -631,7 +631,7 @@ class Time(datetime.time):
 
 
 class gYearMonth:
-    def __init__(self, year: int, month: int) -> None:
+    def __init__(self, year: int | str, month: int | str) -> None:
         self.year = int(year)  # may be negative
         self.month = int(month)
 
@@ -674,7 +674,7 @@ class gYearMonth:
 
 
 class gMonthDay:
-    def __init__(self, month: int, day: int) -> None:
+    def __init__(self, month: int | str, day: int | str) -> None:
         self.month = int(month)
         self.day = int(day)
 
@@ -719,7 +719,7 @@ class gMonthDay:
 
 
 class gYear:
-    def __init__(self, year: int):
+    def __init__(self, year: int | str):
         self.year = int(year)  # may be negative
 
     def __repr__(self) -> str:
@@ -763,7 +763,7 @@ class gYear:
 
 
 class gMonth:
-    def __init__(self, month: int):
+    def __init__(self, month: int | str):
         self.month = int(month)
 
     def __repr__(self) -> str:
@@ -807,7 +807,7 @@ class gMonth:
 
 
 class gDay:
-    def __init__(self, day: int) -> None:
+    def __init__(self, day: int | str) -> None:
         self.day = int(day)
 
     def __repr__(self) -> str:
--- a/arelle/ModelValue.py
+++ b/arelle/ModelValue.py
@@ -629,176 +629,226 @@ class Time(datetime.time):
         time.hour24 = hour24
         return time
 
-class gYearMonth():
+
+class gYearMonth:
     def __init__(self, year: int, month: int) -> None:
-        self.year = int(year) # may be negative
+        self.year = int(year)  # may be negative
         self.month = int(month)
+
     def __repr__(self) -> str:
         return self.__str__()
+
     def __str__(self) -> str:
-        return "{0:0{2}}-{1:02}".format(self.year, self.month, 5 if self.year < 0 else 4) # may be negative
-    def __eq__(self,other: Any) -> bool:
+        return "{0:0{2}}-{1:02}".format(self.year, self.month, 5 if self.year < 0 else 4)  # may be negative
+
+    def __eq__(self, other: Any) -> bool:
         if not isinstance(other, gYearMonth):
             return NotImplemented
         return self.year == other.year and self.month == other.month
-    def __ne__(self,other: Any) -> bool:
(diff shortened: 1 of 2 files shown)


aviary3-wk:
## Security Insights

## Action Items
---

arelle/ModelValue.py
```diff
@@ -629,177 +629,227 @@ def __new__(cls, hour: int = 0, minute: int = 0, second: int = 0, microsecond: i
...
+
     def __bool__(self) -> bool:
         return self.month != 0 or self.day != 0
 
-class gYear():
-    def __init__(self, year: int):
-        self.year = int(year) # may be negative
+
+class gYear:
+    def __init__(self, year: int | str):
```
ggravlingen, line 722:
```suggestion
    def __init__(self, year: int | str) -> None:
```

arelle/ModelValue.py
```diff
@@ -629,177 +629,227 @@ def __new__(cls, hour: int = 0, minute: int = 0, second: int = 0, microsecond: i
...
         return self.year >= other.year
+
     def __bool__(self) -> bool:
         return self.year != 0 != 0
 
-class gMonth():
-    def __init__(self, month: int):
+
+class gMonth:
+    def __init__(self, month: int | str):
```
ggravlingen, line 766:
```suggestion
    def __init__(self, month: int | str) -> None:
```

ggravlingen (changes requested):
Odd that mypy didn't fail on those missing return types.

austinmatherne-wk:
> Odd that mypy didn't fail on those missing return types.

[mypy will infer](https://mypy.readthedocs.io/en/stable/class_basics.html#annotating-init-methods) the return type of `__init__` (because it can only be `None`) as long as at least one param has a type hint:
> The [__init__](https://docs.python.org/3/reference/datamodel.html#object.__init__) method is somewhat special – it doesn’t return a value. This is best expressed as -> None. However, since many feel this is redundant, it is allowed to omit the return type declaration on [__init__](https://docs.python.org/3/reference/datamodel.html#object.__init__) methods if at least one argument is annotated. For example, in the following classes [__init__](https://docs.python.org/3/reference/datamodel.html#object.__init__) is considered fully annotated:
> ```python
> class C1:
>     def __init__(self) -> None:
>         self.var = 42
> 
> class C2:
>     def __init__(self, arg: int):
>         self.var = arg
>```
> However, if __init__ has no annotated arguments and no return type annotation, it is considered an untyped method:
>```python
> class C3:
>     def __init__(self):
>         # This body is not type checked
>         self.var = 42 + 'abc'
>```

[PEP 484](https://peps.python.org/pep-0484/#the-meaning-of-annotations) still suggests that the return type should be included:
> (Note that the return type of __init__ ought to be annotated with -> None. The reason for this is subtle. If __init__ assumed a return annotation of -> None, would that mean that an argument-less, un-annotated __init__ method should still be type-checked? Rather than leaving this ambiguous or introducing an exception to the exception, we simply say that __init__ ought to have a return annotation; the default behavior is thus the same as for other methods.)

ggravlingen:
> > Odd that mypy didn't fail on those missing return types.
> 
> [mypy will infer](https://mypy.readthedocs.io/en/stable/class_basics.html#annotating-init-methods) the return type of `__init__` (because it can only be `None`) as long as at least one param has a type hint:
> 
> > The [**init**](https://docs.python.org/3/reference/datamodel.html#object.__init__) method is somewhat special – it doesn’t return a value. This is best expressed as -> None. However, since many feel this is redundant, it is allowed to omit the return type declaration on [**init**](https://docs.python.org/3/reference/datamodel.html#object.__init__) methods if at least one argument is annotated. For example, in the following classes [**init**](https://docs.python.org/3/reference/datamodel.html#object.__init__) is considered fully annotated:
> > ```python
> > class C1:
> >     def __init__(self) -> None:
> >         self.var = 42
> > 
> > class C2:
> >     def __init__(self, arg: int):
> >         self.var = arg
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > However, if **init** has no annotated arguments and no return type annotation, it is considered an untyped method:
> > ```python
> > class C3:
> >     def __init__(self):
> >         # This body is not type checked
> >         self.var = 42 + 'abc'
> > ```
> 
> [PEP 484](https://peps.python.org/pep-0484/#the-meaning-of-annotations) still suggests that the return type should be included:
> 
> > (Note that the return type of **init** ought to be annotated with -> None. The reason for this is subtle. If **init** assumed a return annotation of -> None, would that mean that an argument-less, un-annotated **init** method should still be type-checked? Rather than leaving this ambiguous or introducing an exception to the exception, we simply say that **init** ought to have a return annotation; the default behavior is thus the same as for other methods.)

Oh, this was new to me. Must've had some linting rule in previous projects I've worked in that cought this. What's your preference here?

ggravlingen (approved):
LGTM

austinmatherne-wk:
> Oh, this was new to me. Must've had some linting rule in previous projects I've worked in that cought this. What's your preference here?

Better to follow the PEP guidance and include them, but I don't feel strongly.

derekgengenbacher-wf (approved):

PR #459, opened 2022-10-27
Custom/local frozen builds

### LINKED ISSUE #375: Support Custom Frozen Builds

There is a need to support custom frozen builds for users who either would like to bundle their own plugins or are working on not yet public code (EFM updates prior to release). They should be able to make changes to the Arelle repo and create custom frozen builds without using GitHub.

Our builds are handled by GitHub workflows which others can reuse in their own forks, but that doesn't help if they're using GitLab or another alternative to GitHub.

The Linux workflow uses docker to setup system dependencies and bundle our particular set of plugins, but uses [buildLinuxDist.sh](https://github.com/Arelle/Arelle/blob/39a2775573f1d538e370ad72c1bc498347dbcff2/scripts/buildLinuxDist.sh) to handle the actually build commands while macOS and Windows run the build commands directly in their workflows.

We should extract the build commands from the Windows and macOS workflows (to a batch and shell script respectively), so that if a user has configured their system appropriately (necessary system libs are installed) then they can run the build scripts without using GitHub workflows.

There are also a bunch of `buildX` scripts in the root of the repo that we can remove.

aaroncameron-wk:
#### Reason for change
Resolves #375 

#### Description of change
- Cleanup obsolete build scripts
- Update Mac/Windows scripts to run similar steps as workflow (assuming pre-configured environment)
- Update workflows to use updated scripts

#### Steps to Test
- Build and run on Windows
- Build and run on Mac
- Run Windows/Mac build workflows

**review**:
@Arelle/arelle


aviary2-wf:
## Security Insights

## Action Items
---

allenbross-wf (approved):

sagesmith-wf (approved):

.github/workflows/build-mac.yml
```diff
@@ -80,13 +80,10 @@ jobs:
...
         run: echo "SETUPTOOLS_SCM_PRETEND_VERSION=$(python -W ignore distro.py --version)" >> $GITHUB_ENV
-      - name: Rebuild messages.pot internationalization file
-        run: python pygettext.py -v -o arelle/locale/messages.pot arelle/*.pyw arelle/*.py
-      - name: Regenerate messages catalog (doc/messagesCatalog.xml)
-        run: python generateMessagesCatalog.py
-      - run: cp -p arelleGUI.pyw arelleGUI.py
       - name: Build app
-        run: python distro.py bdist_mac
+        run: |
+          chmod 755 buildMacDist.sh
```
austinmatherne-wk, line 85: Is this needed if the script is checked in with executable permissions (`git update-index --chmod=+x buildMacDist.sh`)?

buildWinDist.bat
```diff
@@ -0,0 +1,8 @@
+rem Make directories
```
austinmatherne-wk, line 1: Move these build scripts into the `scripts` directory, please. I'd like to cleanup the repo root.

buildMacDist.sh
```diff
@@ -1,117 +1,9 @@
...
 
-# remove old build
-/bin/rm -rf build
-/bin/rm -rf dist
-/bin/rm -rf dist_dmg
-
-# set the build date in version.py
-python3.4 buildVersion.py
-
+# Rebuild messages.pot internationalization file
```
austinmatherne-wk, line 3: At a minimum `set -x` should provide improved logging in the github workflow runs. `set -xeu` is a good practice for any of these workflow scripts.

sagesmith-wf (approved):

austinmatherne-wk (approved):

.github/workflows/build-mac.yml
```diff
@@ -80,13 +80,8 @@ jobs:
           rm -rf tmp
       - name: Capture version before localization
         run: echo "SETUPTOOLS_SCM_PRETEND_VERSION=$(python -W ignore distro.py --version)" >> $GITHUB_ENV
```
austinmatherne-wk, line 82: You should be able to remove the `SETUPTOOLS_SCM_PRETEND_VERSION` env var from the mac and windows workflows now that's handled in the build scripts.

aaroncameron-wk: Right, didn't mean to push. Was trying to get git history clean to make sure the version string is clean locally.

austinmatherne-wk (approved):

PR #478, opened 2022-11-09
cntlr can be none

austinmatherne-wk:
#### Reason for change
Remove not none assertions on cntlr field which can in fact be none

#### Description of change
Remove non none assertions

#### Steps to Test
* CI/mypy

**review**:
@Arelle/arelle

### DIFF
--- a/arelle/FileSource.py
+++ b/arelle/FileSource.py
@@ -200,7 +200,6 @@ class FileSource:
                 return  # an error should have been logged
             if self.isZip:
                 try:
-                    assert self.cntlr is not None
                     assert isinstance(self.basefile, str)
                     fileStream = openFileStream(self.cntlr, self.basefile, 'rb')
                     self.fs = zipfile.ZipFile(fileStream, mode="r")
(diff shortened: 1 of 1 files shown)


aviary-wf:
## Security Insights

## Action Items
---

sagesmith-wf (commented):
What code path does this happen in? It seems odd?

arelle/FileSource.py
```diff
@@ -200,7 +200,6 @@ def open(self, reloadCache: bool = False) -> None:
                 return  # an error should have been logged
             if self.isZip:
                 try:
-                    assert self.cntlr is not None
```
austinmatherne-wk, line 203: arbitrary line for discussion:

> What code path does this happen in? It seems odd?

From building a [xule rule set](https://github.com/xbrlus/xule/blob/60354a037977f0e0442f0c92b265fc594501b99c/plugin/xule/XuleRuleSet.py#L159).

sagesmith-wf (approved):

aaroncameron-wk (approved):

arelle/FileSource.py
sagesmith-wf: Huh, is that the expected usage there? I haven't ever seen someone loading using the filesource directly

derekgengenbacher-wf: Is this path hit when loading the XULE validation zips?

austinmatherne-wk: > Huh, is that the expected usage there? I haven't ever seen someone loading using the filesource directly

Yes, the file stream methods specifically check for and handle the none cntlr case.

derekgengenbacher-wf (approved):

derekgengenbacher-wf:

PR #513, opened 2022-12-14
Reset locale after changing

### LINKED ISSUE #505: Arelle alters the `LC_ALL` environment variable, affecting other processes
### What happened?

We use Arelle to extract Federal Energy Regulatory Commission filings distributed in XBRL format, and turn them into a relational database covering many utilities across multiple filing periods. (See e.g. [the 2021 FERC Form 1](https://data.catalyst.coop/ferc1_xbrl)) and our [FERC XBRL Extractor repo](https://github.com/catalyst-cooperative/ferc-xbrl-extractor).

This extraction is an early step in a longer ETL pipeline, and it runs successfully. However, later on in the pipeline, we encounter `UnicodeDecodeError` on some platforms but not others. These errors go away when a single line is commented out in `arelle/Locale.py`. See [this commit](https://github.com/catalyst-cooperative/Arelle/commit/cef2d830e229ae5c5577ac426acf04dd8a40b0a6) in our fork of the Arelle repo from earlier this year.

That line appears to correspond to [line 77 in Locale.py today](https://github.com/Arelle/Arelle/blob/776c34e578f64299a94adc3b1868511c22f7cac3/arelle/Locale.py#L77).

Testing earlier today with v2.2.2 of Arelle rather than our fork, the exact same problem still arises. See [this comment](https://github.com/catalyst-cooperative/ferc-xbrl-extractor/issues/36#issuecomment-1342764869). The exact ultimate error is:

```
    def decode(self, input, final=False):
>       return codecs.ascii_decode(input, self.errors)[0]
E       UnicodeDecodeError: 'ascii' codec can't decode byte 0xc3 in position 7113: ordinal not in range(128)
```

The XBRL inputs that were being processed by Arelle are contained in [this zipfile](https://zenodo.org/record/7314437/files/ferc1-xbrl-2021.zip?download=1) which is part of [this FERC Form 1 data archive on Zenodo](https://zenodo.org/record/7314437).

To work around this issue without needing to use a fork of Arelle, @zschira created a context manager that resets the `LC_ALL` environment variable after Arelle is finished running. See [this commit](https://github.com/catalyst-cooperative/ferc-xbrl-extractor/pull/42/commits/54d18d2d31a788dbd7d65298c001cc712e4d14a7):

```py
class ArelleInterface:
    """Clear LC_ALL after Arelle has set it internally.
    Arelle sets LC_ALL by default, which can cause some
    downstream encoding issues. This context manager will
    make sure it is cleared before this becomes a problem.
    """

    def __enter__(self):
        """Do nothing."""
        return self

    def __exit__(self, *exc):
        """Unset LC_ALL and return."""
        locale.setlocale(locale.LC_ALL, "")
        return False


def load_xbrl(path: str | FileSource.FileSource):
    """Load XBRL (either taxonomy or individual filing).
    Args:
        path: URL or local path pointing to an XBRL taxonomy or instance.
    """
    with ArelleInterface():
        cntlr = Cntlr.Cntlr()
        cntlr.startLogging(logFileName="logToPrint")
        model_manager = ModelManager.initialize(cntlr)
        return ModelXbrl.load(model_manager, path)
```

We've experienced this problem on:
* Ubuntu Linux 20.04 (GitHub runners)
* MacOS 13.0.1 (Ventura)

## What was expected?

We expected that Arelle would not result in any persistent side-effects altering the computational environment it and subsequent processes were running in.

### Documents

_No response_

_No response_

### Interface

Python library (pip install)

### Version

version 2.2.2

### Download

pip install

### Operating System

macOS 13 Ventura

aaroncameron-wk:
#### Reason for change
Resolves #505

#### Description of change
To improve on Arelle's current usage of the locale usage, ensure that the locale is reset to the previous value whenever Arelle makes a change to it.

#### Steps to Test
Unit tests added.
Run Arelle as GUI, as a library, and via command line. Change locale and expect same results.
**review**:
@Arelle/arelle @zaneselvans

### DIFF
--- a/arelle/Locale.py
+++ b/arelle/Locale.py
@@ -35,8 +35,13 @@ defaultLocaleCodes = {
     "sk": "SK", "sl": "SI", "sq": "AL", "sr": "RS", "sv": "SE", "th": "TH",
     "tr": "TR", "uk": "UA", "ur": "PK", "vi": "VN", "zh": "CN"}
 
-def getUserLocale(localeCode: str = '') -> tuple[LocaleDict, str | None]:
-    # get system localeconv and reset system back to default
+
+def _getUserLocaleUnsafe(localeCode: str = '') -> tuple[LocaleDict, str | None]:
+    """
+    Get locale conventions dictionary. May change the global locale if called directly.
+    :param localeCode: The locale code to use to retrieve conventions. Defaults to system default.
+    :return: Tuple of local conventions dictionary and a user-directed setup message
+    """
     import locale
     conv = None
     localeSetupMessage = None
@@ -71,11 +76,26 @@ def getUserLocale(localeCode: str = '') -> tuple[LocaleDict, str | None]:
                 conv = locale.localeconv()
             except locale.Error:
                 pass
-    locale.setlocale(locale.LC_ALL, 'C')
     if conv is None:  # some other issue prevents getting culture code, use 'C' defaults (no thousands sep, no currency, etc)
+        locale.setlocale(locale.LC_ALL, 'C')
         localeSetupMessage = f"locale code \"{localeCode}\" is not available on this system"
         conv = locale.localeconv() # use 'C' environment, e.g., en_US
     return cast(LocaleDict, conv), localeSetupMessage
+
+
+def getUserLocale(localeCode: str = '') -> tuple[LocaleDict, str | None]:
+    """
+    Get locale conventions dictionary. Ensures that the locale (global to the process) is reset afterwards.
+    :param localeCode: The locale code to use to retrieve conventions. Defaults to system default.
+    :return: Tuple of local conventions dictionary and a user-directed setup message
+    """
+    import locale
+    currentLocale = locale.getlocale()
+    try:
+        return _getUserLocaleUnsafe(localeCode)
+    finally:
+        locale.setlocale(locale.LC_ALL, currentLocale)
+
 
 def getLanguageCode() -> str:
     if sys.platform == "darwin": # MacOS doesn't provide correct language codes
--- a/arelle/Locale.py
+++ b/arelle/Locale.py
@@ -286,6 +286,17 @@ def languageCodes() -> dict[str, str]:  # dynamically initialize after gettext i
         }
         return _languageCodes
 
+
+def setApplicationLocale() -> None:
+    """
+    Sets the locale to C, to be used when running Arelle as a standalone application
+    (e.g., `arelleCmdLine`, `arelleGUI`.)
+    :return:
+    """
+    import locale
+    locale.setlocale(locale.LC_ALL, 'C')
+
+
 _disableRTL: bool = False # disable for implementations where tkinter supports rtl
 def setDisableRTL(disableRTL: bool) -> None:
     global _disableRTL
--- a/arelle/Locale.py
+++ b/arelle/Locale.py
@@ -25,8 +25,6 @@ LC_MONETARY = 4
 LC_NUMERIC = 1
 LC_TIME = 2
 
-C_LOCALE = None # culture-invariant locale
-
 defaultLocaleCodes = {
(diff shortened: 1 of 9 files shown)


aviary3-wk:
## Security Insights

## Action Items
---

arelle/Locale.py
```diff
@@ -74,14 +76,27 @@ def getUserLocale(localeCode: str = '') -> tuple[LocaleDict, str | None]:
...
     if conv is None:  # some other issue prevents getting culture code, use 'C' defaults (no thousands sep, no currency, etc)
+        locale.setlocale(locale.LC_ALL, 'C')
         localeSetupMessage = f"locale code \"{localeCode}\" is not available on this system"
         conv = locale.localeconv() # use 'C' environment, e.g., en_US
-    if C_LOCALE is None:  # load culture-invariant C locale
-        C_LOCALE = locale.localeconv()
     return cast(LocaleDict, conv), localeSetupMessage
 
+
+def getUserLocale(localeCode: str = '') -> tuple[LocaleDict, str | None]:
```
aaroncameron-wk, line 86: In initial discussion, we talked about how this change was most critical when Arelle is being used as a library, and not as critical when being used independently via GUI or command line. This implied that we might skip resetting the locale in those contexts.

Based on what I'm seeing, calls to `getUserLocale` are used sparingly in those contexts, generally once per run during initialization or on demand from the GUI. Because of this, I don't think it's worth the complexity to conditionally perform the reset based on the running context, and instead just commit to it being reset every time Arelle interacts with it.

austinmatherne-wk: My concern with no longer setting the locale for the application context is that it could cause a regression somewhere that depends on the locale being `C` (default file encoding, time str formating, etc.) which Arelle currently sets it to during initialization. Admittedly I couldn't find anywhere in Arelle where this would be an issue, but I still think it's reasonable to set the locale to `C` when it's deployed as an app (and avoid potentially introducing a regression for one of our many builds/platforms.)

There are a couple ways to handle that.
1. Only set for the frozen builds:
```python
if getattr(sys, "frozen", False):
    locale.setlocale(locale.LC_ALL, 'C')
```
2. Set the locale to `C` from the application entry points ([`CntlrCmdLine`](https://github.com/Arelle/Arelle/blob/1a294dcf6d6aa05f9fb2de698a8fd72f5003f0fc/arelle/CntlrCmdLine.py#L548) and [`CntlrWinMain`](https://github.com/Arelle/Arelle/blob/1a294dcf6d6aa05f9fb2de698a8fd72f5003f0fc/arelle/CntlrWinMain.py#L63))

I'd lean towards option #2 as it covers running arelle as an app from source.

aaroncameron-wk: My approach did cause regressions, because the conformance suites are now running without locale being set to `C`. This initially caused some failures due to some seemingly unnecessary localization of date/time parts. My last commit corrects these, but having `C` set in the application contexts should protect us from any unknown regressions that the conformance suites don't cover.

arelle/Locale.py
```diff
@@ -37,10 +35,14 @@
     "sk": "SK", "sl": "SI", "sq": "AL", "sr": "RS", "sv": "SE", "th": "TH",
     "tr": "TR", "uk": "UA", "ur": "PK", "vi": "VN", "zh": "CN"}
 
-def getUserLocale(localeCode: str = '') -> tuple[LocaleDict, str | None]:
-    # get system localeconv and reset system back to default
+
+def getUserLocaleUnsafe(localeCode: str = '') -> tuple[LocaleDict, str | None]:
```
austinmatherne-wk, line 39: Prefix this function with an underscore? I don't see any good reason why we'd want someone to call this from another module.

austinmatherne-wk (dismissed):

austinmatherne-wk (approved):

sagesmith-wf (approved):

derekgengenbacher-wf (approved):

allenbross-wf:
QA +10

PR #541, opened 2023-01-07
Optimize QName operators

### LINKED ISSUE #528: Optimize QName

The [QName class](https://github.com/Arelle/Arelle/blob/98e372ed61803c95882bcde61de7a4271fa50c50/arelle/ModelValue.py#L173) is a hot code path (instances can be created and compared millions of times during validation). We’re using the `@total_ordering` decorator as a convenience, but it is not recommended for situations where [performance is important](https://docs.python.org/3/library/functools.html?highlight=total_ordering#functools.total_ordering).

> While this decorator makes it easy to create well behaved totally ordered types, it does come at the cost of slower execution and more complex stack traces for the derived comparison methods. If performance benchmarking indicates this is a bottleneck for a given application, implementing all six rich comparison methods instead is likely to provide an easy speed boost.

We should remove the decorator and additionally optimize the equality checks by inlining `_conforms` and removing the redundant `isinstance`/`hasattr` checks.

austinmatherne-wk:
#### Reason for change
Resolves #528 

#### Description of change
No significant improvements here, just low hanging fruit that can shave a minute or two off of long running validations.

#### Steps to Test
* CI (unit tests cover QName class)

**review**:
@Arelle/arelle
…onforms

### DIFF
--- a/arelle/ModelValue.py
+++ b/arelle/ModelValue.py
@@ -4,7 +4,6 @@ See COPYRIGHT.md for copyright information.
 from __future__ import annotations
 import datetime, isodate
 from decimal import Decimal
-from functools import total_ordering
 from typing import TYPE_CHECKING, Any, cast, overload, Optional, Union
 from fractions import Fraction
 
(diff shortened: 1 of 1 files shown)


aviary-wf:
## Security Insights

## Action Items
---

PR #548, opened 2023-01-12
[FIX] Reset _dimensionsInUse when a new context is created, to be re-evaluated when needed

RobinetDenisAcsone:
#### Reason for change
Once the cache  `_dimensionsInUse` (see [dimensionsInUse](https://github.com/Arelle/Arelle/blob/master/arelle/ModelXbrl.py#L865) method) is computed, it's never updated, which can cause some issue when new context are created

#### Description of change
When a new context is created, delete the cache to be recomputed when needed

#### Steps to Test
It's possible to test it with the EBA taxonomy, entrypoint http://www.eba.europa.eu/eu/fr/xbrl/crr/fws/corep/its-005-2020/2022-03-01/mod/corep_lcr_da.xsd with table C 77.00 ([here is the taxonomy](https://www.eba.europa.eu/sites/default/documents/files/document_library/Risk%20Analysis%20and%20Data/Reporting%20Frameworks/Reporting%20framework%203.2/Phase%203/1050698/Taxonomy%20Package.zip), add every zip in it with the `Help` > `Manage packages` option):
* open the file attached. It already contains a fact. It's zipped because GitHub doesn't accept .xbrl files
* open table `C 76.00.w` (it will call dimensionsInUse if a fact exist)
* Create a fact in `C 77.00`, reloading the table will show an empty table with current master. It will show the created fact with this fix.

**review**:
@Arelle/arelle

[dimensionsInUse.zip](https://github.com/Arelle/Arelle/files/10402217/dimensionsInUse.zip)

### DIFF
--- a/arelle/ModelXbrl.py
+++ b/arelle/ModelXbrl.py
@@ -37,6 +37,7 @@ if TYPE_CHECKING:
     _: TypeGetText  # Handle gettext
 else:
     ModelFact = None
+    ModelContext = None
 
 
 profileStatNumber = 0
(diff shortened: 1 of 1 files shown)


aviary-wf:
## Security Insights

## Action Items
---

arelle/ModelXbrl.py
```diff
@@ -696,6 +696,8 @@ def createContext(
 
         XmlValidate.validate(self, newCntxElt)
         self.modelDocument.contextDiscover(newCntxElt)
+        if hasattr(self, "_dimensionsInUse"):
```
austinmatherne-wk, line 699: Good catch! It would be better to only do this if the context adds new dimensions. Or even better update the set (if it's defined) with the new dimensions.

austinmatherne-wk (approved):

arelle/ModelXbrl.py
RobinetDenisAcsone: Ok, I will change that, adding the missing dimensions should be possible

RobinetDenisAcsone: Done!

austinmatherne-wk (approved):

sagesmith-wf (approved):

Issue #529, opened 2023-01-03
Type Hints: XPathContext


austinmatherne-wk:
### What should we change and why?

The [XPathContext module](https://github.com/Arelle/Arelle/blob/master/arelle/XPathContext.py) is extensively used by the formula evaluator. We should add type hints prior to refactoring to minimize the potential for regressions.

Issue #536, opened 2023-01-03
Formula Package


austinmatherne-wk:
### What should we change and why?

The formula implementation in Arelle is complex and lacks documentation. The large functions with short cryptic variable names makes  it difficult for new developers to understand what’s going on.

We should move the formula modules into a package (`arelle.formula`), break apart some of the large functions, and give them better more meaningful names.

PR #555, opened 2023-01-17
XPathParser type hints and cleanup

austinmatherne-wk:
#### Reason for change
I found some minor issues in the formula XPath implementation while working on #529.

#### Description of change
* Selectively ran black to improve readability (particularly with the pyparsing grammar)
* Fixed some typos/broken code paths
* Updated deprecated pyparsing class and function names
* Add type hints to modules:
  * XPathParser
  * ModelObjectFactory
  * plugin.functionsMath

#### Steps to Test
* CI

**review**:
@Arelle/arelle

### DIFF
--- a/arelle/ModelManager.py
+++ b/arelle/ModelManager.py
@@ -3,6 +3,7 @@ See COPYRIGHT.md for copyright information.
 '''
 from __future__ import annotations
 
+from collections.abc import Callable
 from typing import TYPE_CHECKING
 import gc, sys, traceback, logging
 from arelle import ModelXbrl, Validate, DisclosureSystem, PackageManager
@@ -12,6 +13,7 @@ from arelle.typing import LocaleDict
 
 if TYPE_CHECKING:
     from arelle.Cntlr import Cntlr
+    from arelle.ModelValue import QName
 
 def initialize(cntlr: Cntlr) -> ModelManager:
     modelManager = ModelManager(cntlr)
@@ -73,7 +75,7 @@ class ModelManager:
         self.abortOnMajorError = False
         self.collectProfileStats = False
         self.loadedModelXbrls = []
-        self.customTransforms = None
+        self.customTransforms: dict[QName, Callable[[str], str]] | None = None
         self.isLocaleSet = False
         self.setLocale()
 
@@ -100,13 +102,11 @@ class ModelManager:
         """
         self.cntlr.addToLog(message, messageCode=messageCode, file=file, refs=refs, level=level)
 
-    def showStatus(self, message, clearAfter=None) -> str:
+    def showStatus(self, message: str | None, clearAfter: int | None = None) -> None:
         """Provide user feedback on status line of GUI or web page according to type of controller.
 
         :param message: Message to display on status widget.
-        :type message: str
         :param clearAfter: Time, in ms., after which to clear the message (e.g., 5000 for 5 sec.)
-        :type clearAfter: int
         """
         self.cntlr.showStatus(message, clearAfter)
 
--- a/arelle/XPathParser.py
+++ b/arelle/XPathParser.py
@@ -102,9 +102,9 @@ def pushDecimal(sourceStr, loc, toks):
 
 
 def pushQuotedString(sourceStr, loc, toks):
-    str = toks[0]
-    q = str[0]
-    dequotedStr = str[1:-1].replace(q + q, q)
+    _str = toks[0]
+    q = _str[0]
+    dequotedStr = _str[1:-1].replace(q + q, q)
     exprStack.append(dequotedStr)
     return dequotedStr
 
--- a/arelle/XPathParser.py
+++ b/arelle/XPathParser.py
@@ -34,7 +34,7 @@ from arelle import ModelValue, XbrlConst, XmlUtil
 from arelle.Locale import format_string
 from arelle.PluginManager import pluginClassMethods
 
-FunctionIxt = None
+ixtNamespaceFunctions = None
 
 
 # Debugging flag can be set to either "debug_flag=True" or "debug_flag=False"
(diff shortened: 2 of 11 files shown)


aviary3-wk:
## Security Insights
The items listed below may not capture all security relevant changes. Before providing a security review, be sure to review the entire PR for security impact.
<details>
<summary>(1) Security relevant changes were detected</summary>
<li>Watched keyword <code>XMLParser(</code> in <code>arelle/ModelObjectFactory.py</code> line(s) ['50'] added</li>
</details>

## Action Items
* Obtain a [security review](https://wiki.atl.workiva.net/display/SECURITY/Development+Security+Review+Guidelines); reviewer should pay special attention to insights listed above
---

arelle/XPathParser.py
```diff
@@ -1,227 +1,343 @@
...
 
 
 # Debugging flag can be set to either "debug_flag=True" or "debug_flag=False"
-debug_flag=True
+debug_flag = True
+
+exprStack: ExpressionStack = []
+xmlElement: ModelObject | None = None
+modelXbrl: ModelXbrl | None = None
+pluginCustomFunctions: dict[QName, Callable[[XPathContext, OperationDef, ModelObject, list[list[FormulaToken]]], Any]] | None = None
```
brettkail-wk, line 81: Out of curiosity, is the value of the dict actually used somewhere?  It seemed like `in` is the only operator used, so perhaps it could just be a set.

Anyway, I noticed that the `list[list[FormulaToken]]` here seems to mismatch `MATH_FUNCTION_ARGS`, and I was trying to understand if `MATH_FUNCTION_ARGS` is actually the correct type, or whether those functions should actually be verifying the types of the arguments.

arelle/ModelManager.py
```diff
@@ -73,7 +75,7 @@ def __init__(self, cntlr: Cntlr):
         self.abortOnMajorError = False
         self.collectProfileStats = False
         self.loadedModelXbrls = []
-        self.customTransforms = None
+        self.customTransforms: dict[QName, Callable[[str], str]] | None = None
```
sagesmith-wf, line 78: Seems like this should/could default to an empty dict instead?

arelle/XPathParser.py
austinmatherne-wk: > Out of curiosity, is the value of the dict actually used somewhere? It seemed like `in` is the only operator used, so perhaps it could just be a set.

Not here. I kept the signature so I'd have it as reference while I'm working on the rest of the formula types.

> Anyway, I noticed that the `list[list[FormulaToken]]` here seems to mismatch `MATH_FUNCTION_ARGS`, and I was trying to understand if `MATH_FUNCTION_ARGS` is actually the correct type, or whether those functions should actually be verifying the types of the arguments.

Good catch! The numeric args func handles `list[list[FormulaToken]]` -> `Number`. Updated!

arelle/XPathParser.py
```diff
@@ -1,227 +1,343 @@
...
+    from arelle.ModelXbrl import ModelXbrl
+    from arelle.ModelManager import ModelManager
+    from arelle.ModelObject import ModelObject
+    from arelle.ModelValue import QName
+    from arelle.XPathContext import XPathContext, XPathException
+    from arelle.typing import TypeGetText
+
+    _: TypeGetText  # Handle gettext
+
+    FormulaToken = Union[
```
sagesmith-wf, line 54: This feels really helpful, what do you think about breaking out into `FormulaGrammar.py` or something that makes it obvious that its defining just the tokens? As a side note, with this being so compilery I don't know if you are already planning some of this but it seems easier to understand for people if we eventualize organize this more like a traditional parser with the  cutoffs, `grammar -> tokenizer -> ast ... ect`

arelle/plugin/formulaLoader.py
```diff
@@ -1283,9 +1283,9 @@ def compileXfsGrammar( cntlr, debugParsing ):
     cntlr.showStatus(_("Compiling Formula Grammar"))
     from pyparsing import (Word, Keyword, alphas,
                  Literal, CaselessLiteral,
-                 Combine, Optional, nums, Or, Forward, Group, ZeroOrMore, OneOrMore, StringEnd, alphanums,
-                 ParserElement, quotedString, dblQuotedString, sglQuotedString, QuotedString,
-                 delimitedList, Suppress, Regex, FollowedBy,
+                 Combine, Opt, nums, Or, Forward, Group, ZeroOrMore, OneOrMore, StringEnd, alphanums,
```
sagesmith-wf, line 1286: Whats the difference between `Optional` and `Opt` I saw you changed a bunch of them?

sagesmith-wf (commented):
This actually goes surprisingly far into helping me understand this, thanks. I think a markdown grammar file or a link to a grammar file could go a long ways also. Also reorganizing some of the code at some point would also be wonderful

arelle/plugin/formulaLoader.py
austinmatherne-wk: They're the same. pyparsing deprecated `Optional` when a class of the same name was created for the base python `typing` module.

arelle/XPathParser.py
austinmatherne-wk: That'll be the first commit in #536

arelle/ModelManager.py
austinmatherne-wk: That could cause additional loads when there aren't any custom transforms (can't tell if they haven't been loaded yet or there just aren't any)? Anyway, I'm only doing the minimal touching of `ModelManager` while working on formula

sagesmith-wf: 👍 with what you have, just getting at having collection set to null in general is an anti-pattern but not specifically calling for that to be addressed here. Perhaps `bool customTransformsIsSet` could be a work around? It seems odd longterm that it should be loading custom transforms in multiple places within the flow

aaroncameron-wk (approved):

PR #570, opened 2023-02-06
Add Windows UI tests.

stevenbronson-wk:
#### Steps to Test
Test on a fork.  I haven't tested the S3 and Alicloud uploads, so you could try those if you're interested and have access to the accounts.

**review**:
@Arelle/arelle


aviary-wf:
## Security Insights

## Action Items
---

tests/integration_tests/ui_tests/ArelleGUITest/ArelleGUITest.sln
```diff
@@ -0,0 +1,31 @@
...
+EndProject
+Global
+	GlobalSection(SolutionConfigurationPlatforms) = preSolution
+		Debug|Any CPU = Debug|Any CPU
+		Debug|x64 = Debug|x64
+		Release|Any CPU = Release|Any CPU
+		Release|x64 = Release|x64
+	EndGlobalSection
+	GlobalSection(ProjectConfigurationPlatforms) = postSolution
+		{0CEE6A09-141B-4DAA-9409-4416A50E60F0}.Debug|Any CPU.ActiveCfg = Debug|Any CPU
```
sagesmith-wf, line 16: Any way to alias this?

tests/integration_tests/ui_tests/ArelleGUITest/ArelleGUITest/Tests.cs
```diff
@@ -0,0 +1,362 @@
...
+                : ExecutionTypeT(ArellePath, ArelleResourcesPath);
+            public record Build(string? ArellePath, string ArelleResourcesPath) : ExecutionTypeT(ArellePath, ArelleResourcesPath);
+        }
+
+        readonly ExecutionTypeT ExecutionType;
+        Window Window;
+
+        public ArelleGUITests()
+        {
+            var useSource = Environment.GetEnvironmentVariable("ARELLE_USE_BUILD") != "true";
```
sagesmith-wf, line 35: Any reason to not use strong type like `boolean`?

tests/integration_tests/ui_tests/ArelleGUITest/ArelleGUITest/Tests.cs
```diff
@@ -0,0 +1,362 @@
...
+            // ARELLE_PATH
+            // source - path to checkout
+            // portable build - path to extract directory
+            // installed build - absent/empty, checks Program Files
+            // convert empty string to null to simplify GitHub Actions conditionals
+            var arellePathEnv1 = Environment.GetEnvironmentVariable("ARELLE_PATH");
```
sagesmith-wf, line 41: Not sure I understand why is this returning empty instead of `null`?

tests/integration_tests/ui_tests/ArelleGUITest/ArelleGUITest/Tests.cs
```diff
@@ -0,0 +1,362 @@
...
+            var arellePathEnv = String.IsNullOrEmpty(arellePathEnv1) ? null : arellePathEnv1;
+            var arelleResourcesPathEnv = Environment.GetEnvironmentVariable("ARELLE_RESOURCES_PATH");
+
+            const string up = @"..\..\..\..\..\..\..\..\..";
```
sagesmith-wf, line 45: Is this a relative path to somewhere? If so you might use an absolute path here, or atleast comment where you are trying to end up

tests/integration_tests/ui_tests/ArelleGUITest/ArelleGUITest/Tests.cs
```diff
@@ -0,0 +1,362 @@
...
+                var arelleGUIPyPath = Path.Combine(executionType.ArellePath!, "arelleGUI.pyw");
+                var parentProcess = Process.Start(
+                    "cmd", $"/c \"{executionType.PythonExe} {arelleGUIPyPath} > {TestsMediaPath}\\{TestContext.CurrentContext.Test.MethodName}-log.txt 2>&1\"");
+
+                Process? MaybeFindProcess()
+                {
+                    foreach (var p in Process.GetProcesses())
+                    {
+                        var isPython = p.ProcessName.EndsWith("python", StringComparison.OrdinalIgnoreCase);
+                        if (!isPython) continue;
```
sagesmith-wf, line 95: Probably avoid the single line if statments and use the

```cs
if (!isPython) {
    continue;
}
```

form

tests/integration_tests/ui_tests/ArelleGUITest/ArelleGUITest/Tests.cs
```diff
@@ -0,0 +1,362 @@
...
+                        var ppid = GetParentProcessId(p);
+                        if (ppid == null) continue;
```
sagesmith-wf, line 97: Probably avoid the single line if statments and use the

tests/integration_tests/ui_tests/ArelleGUITest/ArelleGUITest/Tests.cs
```diff
@@ -0,0 +1,362 @@
...
+                            ppid = GetParentProcessId(Process.GetProcessById(ppid.Value));
+                        }
+                        if (ppid == parentProcess.Id)
+                        {
+                            return p;
+                        }
+                    }
+                    return null;
+                }
+                var process = Retry.WhileNull(MaybeFindProcess, timeout: TimeSpan.FromMinutes(2), interval: TimeSpan.FromMilliseconds(500)).Result;
```
sagesmith-wf, line 110: Suggest constants, and maybe even env vars for all time configuration

tests/integration_tests/ui_tests/ArelleGUITest/ArelleGUITest/Tests.cs
sagesmith-wf: No. I would expect it to return `null` if unset. Why does it return `''` sometimes?

tests/integration_tests/ui_tests/ArelleGUITest/ArelleGUITest/Tests.cs
sagesmith-wf: Sure, but its weakly typed using a `var` instead of strongly typed using `bool useSource = ...;`

tests/integration_tests/ui_tests/ArelleGUITest/ArelleGUITest/Tests.cs
sagesmith-wf: > Creating two constants for every retry seems excessive and makes you have to navigate/hover to read the code.

Sure, but it also documents what the timeout is, and lets you mess with values if something starts failing randomly without having to read the code to understand.

tests/integration_tests/ui_tests/ArelleGUITest/ArelleGUITest/Tests.cs
sagesmith-wf: According to the [docs](https://learn.microsoft.com/en-us/dotnet/api/system.environment.getenvironmentvariable?view=net-7.0#system-environment-getenvironmentvariable(system-string)) its either been set and it returns the value or it hasn't been set and it returns `null`. I am not following why is it variably returning `null` sometimes and `''` other times, it should be one or the other consistently I would think

tests/integration_tests/ui_tests/ArelleGUITest/ArelleGUITest/Tests.cs
sagesmith-wf: Sorry, wrong phrasing here, the point is the same though, why use `var` instead of `bool`. According to the [docs](http://msdn.microsoft.com/en-us/library/bb384061.aspx):
> The var keyword can also be useful when the specific type of the variable is tedious to type on the keyboard, or is obvious, or does not add to the readability of the code. One example where var is helpful in this manner is with nested generic types such as those used with group operations. In the following query, the type of the query variable is IEnumerable>. As long as you and others who must maintain your code understand this, there is no problem with using implicit typing for convenience and brevity.

> However, the use of var does have at least the potential to make your code more difficult to understand for other developers. For that reason, the C# documentation generally uses var only when it is required.

explicit types should be preferred when possible to make the code less difficult to understand

tests/integration_tests/ui_tests/ArelleGUITest/ArelleGUITest.sln
sagesmith-wf: It seems like this maybe doesn't need to exist unless we want people to be using multiple projects? If we are going to keep this, it seems specifically catered towards visual studio (since we don't have multiple projects) so we should address the newlines by providing a config file also [here](https://github.com/Arelle/Arelle/pull/570#discussion_r1097676712)

tests/integration_tests/ui_tests/ArelleGUITest/ArelleGUITest/Tests.cs
```diff
@@ -0,0 +1,373 @@
...
+            Mouse.LeftClick(new Point(
+                (int)(selectFileFromArchivePane.BoundingRectangle.X + selectFileFromArchivePane.BoundingRectangle.Width / 2.0f),
+                (int)(selectFileFromArchivePane.BoundingRectangle.Y
+                // +1 for the header, +0.5 to end up in the middle
+                + selectFileFromArchivePane.BoundingRectangle.Height * (indexOfEntryPoint + 1.5) / (nFilesInDocumentZip + 1))
+            ));
+            Keyboard.Type(VirtualKeyShort.TAB, VirtualKeyShort.ENTER);
+
+            Logger.Default.Info("wait for document to load and UI to populate");
+            Retry.WhileNull(() => Window.FindFirstByXPath("/Pane/Pane/Pane/Pane/Pane/Pane/Pane"), timeout: TimeSpan.FromMinutes(1), ignoreException: true);
```
sagesmith-wf, line 208: What's this about?

tests/integration_tests/ui_tests/ArelleGUITest/ArelleGUITest/Tests.cs
```diff
@@ -0,0 +1,373 @@
...
+            Logger.Default.Info("open document");
+            Keyboard.Type(VirtualKeyShort.ENTER);
+            Wait.UntilInputIsProcessed();
+
+            Logger.Default.Info("select file inside document archive");
+            var selectFileFromArchiveWindow = Retry.WhileNull(
+                () => Window.FindFirstChild(Automation.ConditionFactory.ByControlType(ControlType.Window)), timeout: TimeSpan.FromSeconds(3)).Result;
+            var selectFileFromArchivePane = selectFileFromArchiveWindow.FindFirstByXPath("/Pane/Pane/Pane/Pane[1]");
+            Mouse.LeftClick(new Point(
+                (int)(selectFileFromArchivePane.BoundingRectangle.X + selectFileFromArchivePane.BoundingRectangle.Width / 2.0f),
```
sagesmith-wf, line 200: Maybe abstract to a click in center that takes the thing that you want to click?

tests/integration_tests/ui_tests/ArelleGUITest/ArelleGUITest/Tests.cs
```diff
@@ -0,0 +1,373 @@
...
+            copyToClipboardMenuItem.Click();
+            Wait.UntilInputIsProcessed();
+            var copyToClipboardSubMenu = Automation.GetDesktop()
+                .FindAllChildren(cf => cf.ByControlType(ControlType.Menu).And(cf.ByProcessId(Application.ProcessId)))
+                .Select(e => e.AsMenu())
+                .Single(e => e.Items.Length == 3);
+            var copyTableMenuItem = copyToClipboardSubMenu.Items[2];
+            copyTableMenuItem.Click();
+            Wait.UntilInputIsProcessed();
+            Thread.Sleep(200);
```
sagesmith-wf, line 237: Why is this needed? Does the wait not wait long enough?

tests/integration_tests/ui_tests/ArelleGUITest/ArelleGUITest/Tests.cs
```diff
@@ -0,0 +1,373 @@
...
+            var messagesAreaAndTabBar = mainAreaChildren[0];
+            var tablesAndFactView = mainAreaChildren[1];
+            var messagesArea = messagesAreaAndTabBar.FindFirstChild();
+            var tablesAndFactViewChildren = tablesAndFactView.FindAllChildren();
+            var factViewAndTabBar = tablesAndFactViewChildren[0];
+            var tablesAndTabBar = tablesAndFactViewChildren[1];
+            var factView = factViewAndTabBar.FindFirstChild();
+            var tables = tablesAndTabBar.FindFirstChild();
+
+            Logger.Default.Info("copy structure to clipboard");
```
sagesmith-wf, line 224: What structure? The presentation tree?

tests/integration_tests/ui_tests/ArelleGUITest/ArelleGUITest/Tests.cs
```diff
@@ -0,0 +1,373 @@
...
+            return (int?)typeof(Process)
+                    .GetProperty("ParentProcessId", System.Reflection.BindingFlags.NonPublic | System.Reflection.BindingFlags.Instance)!
+                    .GetGetMethod(true)!
+                    .Invoke(process, null);
+        }
+
+        static string GetClipboardText()
+        {
+            string s = "";
+            var t = new Thread(() => { s = System.Windows.Forms.Clipboard.GetText(); });
```
sagesmith-wf, line 254: Why does this need to be multithreaded? It seems like it immediately waits anyways?

tests/integration_tests/ui_tests/ArelleGUITest/ArelleGUITest/Tests.cs
```diff
@@ -0,0 +1,373 @@
...
+                .FindAllChildren(cf => cf.ByControlType(ControlType.Menu).And(cf.ByProcessId(Application.ProcessId)))
+                .Select(e => e.AsMenu())
+                .Single(e => e.Items.Length == 3);
+            var copyTableMenuItem = copyToClipboardSubMenu.Items[2];
+            copyTableMenuItem.Click();
+            Wait.UntilInputIsProcessed();
+            Thread.Sleep(200);
+            Logger.Default.Info("get clipboard text");
+            var tableText = GetClipboardText();
+            Assert.That(tableText, Is.EqualTo(TABLE_STRUCTURE.Replace("\r", "")));
```
sagesmith-wf, line 240: This should probably either be screen recording the entire time as well as providing a screenshot when it fails?

tests/integration_tests/ui_tests/ArelleGUITest/ArelleGUITest/Tests.cs
sagesmith-wf: I am not sure what that means, but [this](https://learn.microsoft.com/en-us/dotnet/api/system.windows.forms.clipboard.gettext?view=windowsdesktop-7.0) doesn't imply to me that you should need a thread if your main thread is setting the correct sta attribute.

tests/integration_tests/ui_tests/ArelleGUITest/ArelleGUITest/Tests.cs
sagesmith-wf: Maybe `waitForNestedPane(depth: 7)`?

tests/integration_tests/ui_tests/ArelleGUITest/ArelleGUITest/Tests.cs
sagesmith-wf: want to update the log to say its grabbing the xbrl structure or presentation structure or something?

tests/integration_tests/ui_tests/ArelleGUITest/ArelleGUITest/Tests.cs
sagesmith-wf: 🤷 it seems like an interesting decision to go against the guidance provided by the language maintainers but that said I suppose I also could have picked a better one to comment on. Here is an example of a place where I would have to execute/debug the code/have the api up to figure out what type I am expecting to be returned here :

```cs
var iconsAndMainArea = Window.FindFirstByXPath("/Pane/Pane");
var iconsAndMainAreaChildren = iconsAndMainArea.FindAllChildren();
var mainArea = iconsAndMainAreaChildren[0];
var icons = iconsAndMainAreaChildren[1];
var statusBar = iconsAndMainAreaChildren[2];
var mainAreaChildren = mainArea.FindAllChildren();
var messagesAreaAndTabBar = mainAreaChildren[0];
var tablesAndFactView = mainAreaChildren[1];
var messagesArea = messagesAreaAndTabBar.FindFirstChild();
var tablesAndFactViewChildren = tablesAndFactView.FindAllChildren();
var factViewAndTabBar = tablesAndFactViewChildren[0];
var tablesAndTabBar = tablesAndFactViewChildren[1];
var factView = factViewAndTabBar.FindFirstChild();
var tables = tablesAndTabBar.FindFirstChild();
```

tests/integration_tests/ui_tests/ArelleGUITest/ArelleGUITest/Tests.cs
austinmatherne-wk: A comment explaining what in the UI this is looking for would be appreciated (and help support debugging if it breaks).

sagesmith-wf (dismissed):

allenbross-wf (approved):

austinmatherne-wk (approved):

aaroncameron-wk (approved):

austinmatherne-wk (dismissed):

austinmatherne-wk (approved):

austinmatherne-wk:
# Verified

* workflows on fork

PR #592, opened 2023-02-20
Only self-close selected XHTML elements

### LINKED ISSUE #574: Don't self-close all empty tags when extracting text block content from iXBRL with escape="true"

When extracting values from an iXBRL document with `escape="true"`, Arelle will always self-close empty elements.  For example, if we have some iXBRL like this:

```xml
<ix:nonNumeric escape="true"><b>Foo<i></i>Bar</b></ix:nonNumeric>
```

This will result in a fact value of (escaped):

```xml
<b>Foo<i />Bar</b>
```
This is fine as long as the value is treated as XHTML, rather than HTML.  If it's treated as HTML, `<i />` will be considered to be an opening tag, and the word "Bar" will be rendered in italics.

Unfortunately, expanding all tags also creates problems.  Converting `<br />` to `<br></br>`  results in a double `<br>` if treated as HTML, so this expanding needs to be done selectively.  There is code to do this in the iXBRL viewer [here](https://github.com/Workiva/ixbrl-viewer/blob/master/iXBRLViewerPlugin/xhtmlserialize.py#L18)

Arelle uses a fully custom serializer, and always self-closes (see [here](https://github.com/Arelle/Arelle/blob/master/arelle/XmlUtil.py#L278))

I think Arelle should either preserve the self-closing vs empty distinction from the source document (which I think is impossible when using lxml), or apply the same logic as used in the iXBRL viewer to produce XHTML that will render correctly when treated as HTML.

See https://github.com/Workiva/ixbrl-viewer/pull/379#pullrequestreview-1288286580 for context.

paulwarren-wk:
#### Reason for change

Fixes #574 

#### Description of change

Before this change, all empty elements in `ix:nonNumeric` elements with `escape="true"` were self-closed.  With this change, only those these elements the following elements can be are closed: 'area', 'base', 'basefont', 'br', 'col', 'frame', 'hr', 'img','input', 'isindex', 'link', 'meta', 'param'.  This list is taken from https://www.w3.org/TR/html401/index/elements.html (empty elements).

#### Steps to Test

Create a nonNumeric tag like this:

```xml
<ix:nonNumeric name="eg:StringConcept" contextRef="c1" id="f100" escape="true">
    <div>                
Line 1<br></br>Line 2.  Empty div <div style="color: red"></div> Is this red?
    </div>
</ix:nonNumeric>
```

View the content of the fact in Arelle, and confirm that in the output, the `br` tag becomes `<br />` but the empty `div` remains `<div style="color: red"></div>`.  This is important because if the text block is rendered as HTML (rather than XHTML) a self-closed `div` would be treated as an opening tag, and the "is this red?" text would be incorrectly rendered in red.

**review**:
@Arelle/arelle

### DIFF
--- a/arelle/XmlUtil.py
+++ b/arelle/XmlUtil.py
@@ -181,6 +181,12 @@ def textNotStripped(element: ModelObject | PrototypeObject | None) -> str:
         return ""
     return element.textValue  # allows embedded comment nodes, returns '' if None
 
+def selfClosable(elt):
+    return elt.qname.localName in (
+        'area', 'base', 'basefont', 'br', 'col', 'frame', 'hr', 'img',
+        'input', 'isindex', 'link', 'meta', 'param'
+    )
+
 # ixEscape can be None, "html" (xhtml namespace becomes default), "xhtml", or "xml"
 def innerText(
     element: ModelObject,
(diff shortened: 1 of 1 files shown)


aviary3-wk:
## Security Insights

## Action Items
---

brettkail-wk (dismissed):

austinmatherne-wk (approved):

PR #643, opened 2023-04-05
Formula Performance: additional improvements and cap cache memeory usage

austinmatherne-wk:
#### Reason for change
It was discovered that for some extreme documents (with hundreds of dimensions per fact) that the formula cache could grow to many GB in size. This puts a cap on that and also introduces additional performance improvements targeting large documents.

#### Description of change
Various formula improvements 

#### Steps to Test
* CI (formula conformance suite)

**review**:
@Arelle/arelle


aviary3-wk:
## Security Insights

## Action Items
---

arelle/CntlrCmdLine.py
```diff
@@ -886,6 +888,8 @@ def monitorParentProcess():
             fo.compileOnly = True
         if options.formulaAction:
             fo.formulaAction = options.formulaAction
+        if options.formulaCacheSize:
+            options.formulaCacheSize
```
brettkail-wk, line 892: Assign?
```
fo.cacheSize = options.formulaCacheSize
```

arelle/formula/FactAspectsCache.py
```diff
@@ -10,24 +10,60 @@
...
     def cacheNotMatch(self, fact1: ModelFact, fact2: ModelFact, aspect: int | QName) -> None:
-        self._register(fact1, fact2, aspect, False)
+        self._prioritizedAspects.add(aspect)
+        if not self._maxSizePreviouslyReached:
+            self._register(fact1, fact2, aspect, False)
 
     def _register(self, fact1: ModelFact, fact2: ModelFact, aspect: int | QName, value: bool) -> None:
+        if self._size >= self._maxSize:
+            # Disabling the cache entirely is somewhat rudimentary. Alternative caching strategies, such as LRU, were
+            # explored, but they demonstrated poorer performance on average across all tested documents. The current
```
brettkail-wk, line 50: Just discussion, but this creates a surprising performance "cliff" where if a document gets to a specific size, it will suddenly get much worse performance.

arelle/formula/FactAspectsCache.py
```diff
@@ -10,24 +10,60 @@
...
 
-    def evaluations(self, fact1: ModelFact, fact2: ModelFact) -> defaultdict[int | QName, bool | None]:
-        return self._matchingAspects[fact1][fact2]
+    @property
+    def prioritizedAspects(self) -> set[int | QName]:
+        return self._prioritizedAspects
+
+    def evaluations(self, fact1: ModelFact, fact2: ModelFact) -> defaultdict[int | QName, bool | None] | None:
+        if not self._maxSizePreviouslyReached:
+            if fact1 in self._matchingAspects and fact2 in self._matchingAspects[fact1]:
```
brettkail-wk, line 32: Out of curiosity, did you consider adding a unique fact integer ID to each fact?  The dict key could be a (min(id1,id2),max(id1,id2),aspects) tuple.

brettkail-wk (approved):

arelle/formula/FactAspectsCache.py
austinmatherne-wk: I experimented with tuple keys (it wasn't sorted and still performed a double lookup), but that had worse performance. I'll run another experiment with this specifically and open a followup PR if it's better

brettkail-wk: Interesting.  My intuition was that a single dict lookup would be better than multiple, but perhaps tuple overhead makes that not worth it.  I suppose even with multi-level dict, having an orderable id so you know which order to do the dict lookup might help, but perhaps not much.

arelle/CntlrCmdLine.py
```diff
@@ -302,6 +302,8 @@ def parseAndRun(args):
     parser.add_option("--formularunids", action="store", dest="formulaRunIDs", help=SUPPRESS_HELP)
     parser.add_option("--formulaCompileOnly", action="store_true", dest="formulaCompileOnly", help=_("Specify formula are to be compiled but not executed."))
     parser.add_option("--formulacompileonly", action="store_true", dest="formulaCompileOnly", help=SUPPRESS_HELP)
+    parser.add_option("--formulaCacheSize", action="store", dest="formulaCacheSize", help=_("Specify the number of fact aspect combinations to cache during formula evaluations. Negative numbers have no limit. (10_000_000 is default)"))
+    parser.add_option("--formulacachesize", action="store", dest="formulaCacheSize", help=_("Specify the number of fact aspect combinations to cache during formula evaluations. Negative numbers have no limit. (10_000_000 is default)"))
```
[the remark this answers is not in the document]
sagesmith-wf: Alternatively because this seems confusing, is there any reason to not just pass multiple flags and call `add_option` once instead?

```bash
>>> import optparse
>>> parser = optparse.OptionParser()
>>> parser.add_option('--integers', '--Integers', type=int)
<Option at 0x10577f190: --integers/--Integers>
>>> parser.parse_args(['--integers', "1"])
(<Values at 0x1057abd10: {'integers': 1}>, [])
>>> parser.parse_args(['--Integers', "1"])
(<Values at 0x1057aba90: {'integers': 1}>, [])
```

Also seems confusing to not use the library to handle the default:

```python
>>> parser.add_option('--integers', '--Integers', type=int, default=100000)
<Option at 0x101802210: --integers/--Integers>
>>> parser.parse_args([])
(<Values at 0x10159b710: {'integers': 100000}>, [])
```

arelle/formula/FactAspectsCache.py
```diff
@@ -10,24 +10,60 @@
...
 
 class FactAspectsCache:
-    def __init__(self) -> None:
-        self._matchingAspects: defaultdict[ModelFact, defaultdict[ModelFact, defaultdict[int | QName, bool | None]]] = defaultdict(lambda: defaultdict(lambda: defaultdict(lambda: None)))
+    def __init__(self, maxSize: int) -> None:
+        self._size = 0
+        self._maxSize = maxSize if maxSize >= 0 else float("inf")
+        self._maxSizePreviouslyReached = False
+        self._prioritizedAspects: set[int | QName] = set()
+        self._matchingAspects: defaultdict[
```
sagesmith-wf, line 18: pre-existing: `defaultdict` is an implementation detail, it seems more accurate to type as `dict[ModelFact, dict[ModelFact, dict[int | QName, bool | None]]]`

arelle/formula/FactAspectsCache.py
```diff
@@ -10,24 +10,60 @@
...
+            ModelFact,
+            defaultdict[
+                ModelFact,
+                defaultdict[int | QName, bool | None]
+            ]
+        ] = defaultdict(lambda: defaultdict(lambda: defaultdict(lambda: None)))
```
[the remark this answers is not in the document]
sagesmith-wf: `defaultdict(lambda: None)` is just a regular `dict` right?

arelle/CntlrCmdLine.py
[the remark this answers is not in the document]
austinmatherne-wk: We also have to provide a default for the GUI and Python library paths

arelle/formula/FactAspectsCache.py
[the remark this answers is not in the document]
austinmatherne-wk: Use `clear()` to assign `_matchingAspects`, `_size`, and `_prioritizedAspects` from `__init__`? I'd rather duplicate that code between __init__ and clear and not complicate the type hints by not having those attributes directly defined within `__init__`

arelle/formula/FactAspectsCache.py
austinmatherne-wk: In practice, there isn't a sharp drop-off here. The truly massive, multi-dimensional documents end up with many unique values in the cache (that are never looked up), and the overhead of inserting values overtakes the benefit of the occasional cache hit. There's probably a better heuristic in detecting the number of dimensions, but I don't think it's worth the effort without a larger set of example documents to fine-tune the algorithm. I even tried disabling positive matches in the cache for large documents and continued inserting only false values (to enable short-circuiting). However, in the end, prioritized aspect sorting benefits these documents far more than the cached values. This is because, out of around 200 dimensions across 1000+ facts, all but 1-5 of them will match. If you can identify those 1-5 and check them first, the cache no longer provides a meaningful improvement.

I'll give Steven's suggestion a try and take another round of measurements, but I suspect the performance will be similar, with larger sustained memory usage.

On a related note, I initially implemented prioritized aspects using a counter to provide a more detailed weight for the sorting function. Surprisingly, through random experiments, I discovered that a simple set was more performant (because sorting was more expensive, or updating the counter, I don't know).

brettkail-wk: I agree that for the docs+formulas you've looked at that there isn't a short drop-off.  I was thinking of a scenario with docs+formulas where the cache is effective and has 9M entries, and then the docs or formulas grow such that they would need a cache with 12M entries to hold them all.  Depending on what the formulas are doing, continuing to use the 10M cached entries could be significantly better than not using the cache at all, right?

arelle/formula/FactAspectsCache.py
austinmatherne-wk: defaultdict is guaranteed to return a value even if the key doesn't exist. That isn't an implementation detail. FormulaEvaluator depends on that behavior (instead of using `dict.get()` which turned out to be slower)

arelle/formula/FactAspectsCache.py
austinmatherne-wk: That's reasonably. It's entirely possible I'm over fitting the optimizations to the limited docs that I have. I went ahead and changed this to retain the cache.

derekgengenbacher-wf (approved):

PR #665, opened 2023-04-17
Add option to disable persistent config

derekgengenbacher-wf:
#### Reason for change
Adding this option should allow Arelle to work better with Docker containers that are not setup to have permissions to write directly to the image for a cache.

#### Description of change
Added option to disable config cache

#### Steps to Test
CI

**review**:
@Arelle/arelle


arelle/CntlrCmdLine.py
```diff
@@ -29,6 +29,7 @@
 win32file = win32api = win32process = pywintypes = None
 STILL_ACTIVE = 259 # MS Windows process status constants
 PROCESS_QUERY_INFORMATION = 0x400
+DISABLE_CONFIG_CACHE_OPTION = "--disableConfigCache"
```
austinmatherne-wk, line 32: The config isn't really a cache. Perhaps `--disablePersistentConfig`?

arelle/CntlrCmdLine.py
```diff
@@ -753,9 +758,9 @@ def monitorParentProcess():
         if options.outputAttribution:
             self.modelManager.outputAttribution = options.outputAttribution
         self.modelManager.validateTestcaseSchema = options.validateTestcaseSchema
-        if options.internetConnectivity == "offline":
+        if options.internetConnectivity == "offline" or options.disableConfigCache:
```
austinmatherne-wk, line 761: I know I'm the one who said offline mode should be triggered by this setting, but if we use `--disablePersistentConfig` then I think it would be better to not change the offline mode behavior and instead require users to pass both flags (beyond the issue that triggered this, I can see a desire to not persist config settings while still having a dynamic cache)

austinmatherne-wk (approved):

allenbross-wf (approved):

sagesmith-wf (approved):

PR #667, opened 2023-04-19
Make cntrl unit tests work the same regardless of if files exist on disk

austinmatherne-wk:
#### Reason for change
The mock open is called a varying number of times depending on what config files exist on disk.

#### Description of change
This updates the tests to also mock the path exists call to make sure the tests run the same on systems without any config files and those that do.

Additionally, creates a pytest mark to run only slow (or fast) unit tests
`> pytest tests/unit_tests -m slow`
`> pytest tests/unit_tests -m "not slow"`

#### Steps to Test
* CI
* Run unit tests locally

**review**:
@Arelle/arelle

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -88,6 +88,10 @@ write_to = "arelle/_version.py"
 [tool.pytest.ini_options]
 xfail_strict = true
 empty_parameter_set_mark = "fail_at_collect"
+addopts = "--strict-markers"
+markers = [
+    "slow: marks tests as slow (deselect with '-m \"not slow\"')",
+]
 
 [tool.mypy]
 # Warn when a # type: ignore comment does not specify any error codes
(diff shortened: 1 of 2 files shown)


aviary3-wk:
## Security Insights

## Action Items
---

derekgengenbacher-wf (approved):

pyproject.toml
```diff
@@ -88,6 +88,10 @@ write_to = "arelle/_version.py"
 [tool.pytest.ini_options]
 xfail_strict = true
 empty_parameter_set_mark = "fail_at_collect"
+addopts = "--strict-markers"
+markers = [
+    "slow: marks tests as slow (deselect with '-m \"not slow\"')",
```
sagesmith-wf, line 93: Would you consider deselect being the `fast` flag instead or is that not something you control?

sagesmith-wf (approved):

pyproject.toml
austinmatherne-wk: That would require adding a mark to all tests except those that are slow. I copied this from [the pytest docs](https://docs.pytest.org/en/6.2.x/mark.html#registering-marks)

PR #754, opened 2023-06-26
calc 1.1, fix spurious EFM.6.5.57 msg

austinmatherne-wk:
#### Reason for change
For EDGAR update 23.2.u2

#### Description of change
Fixes calc 1.1 and spurious efm 6.5.57 msg

#### Steps to Test
* test with attached calc 1.1 version of EFM test suite and regression tests

[614-calc1.1-syntax.zip](https://github.com/Arelle/Arelle/files/11869222/614-calc1.1-syntax.zip)

**review**:
@Arelle/arelle

### DIFF
--- a/arelle/ModelRelationshipSet.py
+++ b/arelle/ModelRelationshipSet.py
@@ -98,6 +98,12 @@ def baseSetRelationship(arcElement):
             return rel
     return None
 
+def baseSetSortkey(baseSetKey): # for base set sorting functions
+    arcrole, linkrole, _linkqname, _arcqname = baseSetKey
+    # arcrole may be a string or tuple of strings.
+    return (",".join(arcrole) if isinstance(arcrole, (tuple,list)) else arcrole or "",
+            linkrole or "")
+
 class ModelRelationshipSet:
     __slots__ = ("isChanged", "modelXbrl", "arcrole", "linkrole", "linkqname", "arcqname",
                  "modelRelationshipsFrom", "modelRelationshipsTo", "modelConceptRoots", "modellinkRoleUris",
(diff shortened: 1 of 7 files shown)


derekgengenbacher-wf (approved):

arelle/ModelRelationshipSet.py
```diff
@@ -98,6 +98,12 @@ def baseSetRelationship(arcElement):
             return rel
     return None
 
+def baseSetSortkey(baseSetKey): # for base set sorting functions
+    arcrole, linkrole, _linkqname, _arcqname = baseSetKey
+    # arcrole may be a string or tuple of strings.
+    return (",".join(arcrole) if isinstance(arcrole, (tuple,list)) else arcrole or "",
+            linkrole or "")
```
brettkail-wk, line 105: Would it be more robust to normalize to `list`/`tuple` rather than joining with an arbitrary delimiter?

hermfischer-wf: we need it lexicographic, where some newer base sets have multiple alternative arcroles (calc, TLB, probably others).  It was doing as array/string before but no longer works where calc 1.1 has multiple in some base sets and single in other.

brettkail-wk: I believe `list<list` is still lexicographic for inner strings (`['a']<['b']` vs `['b']<['a']`), it just generally seems cleaner (avoids needing to pick an unused delimiter; `['a','b']<['a,b']` produces different result than joining).

austinmatherne-wk: This is going out at the SEC as is. But they should address this in their next update

PR #811, opened 2023-08-02
Fix "string index out of range" if remapping exactly matches entry po…

paulwarren-wk:
#### Reason for change

If a taxonomy package has a remapping in `catalog.xml` that is the same as an entry point URL in `taxonomyPackage.xml` you get the following error:

```
arelle --packages build/codelist-common-WGWD-YYYY-MM-DD.zip -v -f build/currency-current-WGWD-YYYY-MM-DD/currency-en.xsd 
  File "/home/pdw/Arelle/arelleCmdLine.py", line 43, in <module>
    CntlrCmdLine.main()
  File "/home/pdw/Arelle/arelle/CntlrCmdLine.py", line 52, in main
    parseAndRun(args)
  File "/home/pdw/Arelle/arelle/CntlrCmdLine.py", line 456, in parseAndRun
    cntlr.run(options)
  File "/home/pdw/Arelle/arelle/CntlrCmdLine.py", line 692, in run
    packageInfo = PackageManager.addPackage(self, cmd, options.packageManifestName)
  File "/home/pdw/Arelle/arelle/PackageManager.py", line 605, in addPackage
    newPackageInfo = packageInfo(cntlr, url, packageManifestName=packageManifestName)
  File "/home/pdw/Arelle/arelle/PackageManager.py", line 511, in packageInfo
    parsedPackage = parsePackage(_cntlr, filesource, packageFileUrl, packageFilePrefix, errors)
  File "/home/pdw/Arelle/arelle/PackageManager.py", line 256, in parsePackage
    if not (_remappedUrl[0] in (os.sep, '/') or mapTo[-1] in (os.sep, '/')):
IndexError: string index out of range
```
`_remappedURL` is the remainder of the URL that isn't covered by the remapping, so in this case it's a zero length string.

#### Description of change

Check the length of `_remappedURL` before access.

#### Steps to Test

Attempt to add a taxonomy package with a remapping that exactly matches an entry point URL.  

[codelist-common-WGWD-YYYY-MM-DD.zip](https://github.com/Arelle/Arelle/files/12242325/codelist-common-WGWD-YYYY-MM-DD.zip)

**review**:
@Arelle/arelle

### DIFF
--- a/arelle/PackageManager.py
+++ b/arelle/PackageManager.py
@@ -253,7 +253,7 @@ def parsePackage(cntlr, filesource, metadataFile, fileBase, errors=[]):
                     prefixLength = len(mapFrom)
                     if prefixLength > longestPrefix:
                         _remappedUrl = remappedUrl[prefixLength:]
-                        if not (_remappedUrl[0] in (os.sep, '/') or mapTo[-1] in (os.sep, '/')):
+                        if len(_remappedUrl) > 0 and (not (_remappedUrl[0] in (os.sep, '/') or mapTo[-1] in (os.sep, '/'))):
                             _remappedUrl = mapTo + os.sep + _remappedUrl
                         else:
                             _remappedUrl = mapTo + _remappedUrl
--- a/arelle/PackageManager.py
+++ b/arelle/PackageManager.py
@@ -253,7 +253,7 @@ def parsePackage(cntlr, filesource, metadataFile, fileBase, errors=[]):
                     prefixLength = len(mapFrom)
                     if prefixLength > longestPrefix:
                         _remappedUrl = remappedUrl[prefixLength:]
-                        if len(_remappedUrl) > 0 and (not (_remappedUrl[0] in (os.sep, '/') or mapTo[-1] in (os.sep, '/'))):
+                        if len(_remappedUrl) > 0 and not _remappedUrl.startswith((os.sep, '/')) and not mapTo.endswith((os.sep, '/')):
                             _remappedUrl = mapTo + os.sep + _remappedUrl
                         else:
                             _remappedUrl = mapTo + _remappedUrl


aviary-wf:
## Security Insights

## Action Items
---

arelle/PackageManager.py
```diff
@@ -253,7 +253,7 @@ def parsePackage(cntlr, filesource, metadataFile, fileBase, errors=[]):
                     prefixLength = len(mapFrom)
                     if prefixLength > longestPrefix:
                         _remappedUrl = remappedUrl[prefixLength:]
-                        if not (_remappedUrl[0] in (os.sep, '/') or mapTo[-1] in (os.sep, '/')):
+                        if len(_remappedUrl) > 0 and (not (_remappedUrl[0] in (os.sep, '/') or mapTo[-1] in (os.sep, '/'))):
```
brettkail-wk, line 256: It seems easier to read with the `not` were distributed:
```
if len(_remappedUrl) > 0 and not _remappedUrl[0] in (os.sep, '/') and not mapTo[-1] in (os.sep, '/'):
```
Does the same issue also exist for empty `mapTo`?  Perhaps `startswith(tuple)`/`endswith(tuple)` is even better:
```
if len(_remappedUrl) > 0 and not _remappedUrl.startswith((os.sep, '/')) and not mapTo.endswith((os.sep, '/')):
```
Well, I'm somewhat surprised that `os.sep` is involved when it seems like URL remappings should always use `/`.

brettkail-wk (dismissed):

austinmatherne-wk (approved):

arelle/PackageManager.py
paulwarren-wk: Agreed on all points, and have implemented your second suggestion.

I'm not even sure what this code is doing.  It seems to do some fixing up of remappings that is not justified by the [spec](https://www.oasis-open.org/committees/download.php/14809/xml-catalogs.html#s.rewriteuri):

> Rewriting removes the matching prefix from the supplied URI reference and replaces it with the value of the value of the rewritePrefix attribute.

I don't  think there's anything that requires a remapping to split on a path segment boundary, and I can't see anything that suggests that `/` should be inserted.

austinmatherne-wk (approved):

arelle/PackageManager.py
pdw-mb: > I don't think there's anything that requires a remapping to split on a path segment boundary, and I can't see anything that suggests that / should be inserted.

I've raised a bug for the [Taxonomy Package spec](https://gitlab.xbrl.org/base-spec/taxonomy-package/-/issues/14) for this (XII gitlab, access restricted).

PR #838, opened 2023-08-23
Use xbrli namespace prefix in newly created xbrl documents

derekgengenbacher-wf:
#### Reason for change
When the `saveTargetDocument` code of InlineXBRLDocumentSet is triggered, this code creates a new xbrl instance model. During which ModelDocument creates new xml under the covers. However, the xml is missing a namespace prefix and thus all newly created xml documents use the default namespace.

This change updates the default namespace to be `xbrli` in the under the cover xml generation. 

#### Steps to Test
Take and ixbrl file and run it through the `saveTargetDocument` code in InlineXBRLDocumentSet. With out the fix, all xbrl elements will be written under a generic `<xbrl>` tag. All the tags for contexts including context, entity, etc. will also just be there generic forms. 

With the fix in place the initial '<xbrl>' tag now looks like `<xbrli:xbrl>` and Context tags will also be prefixed correctly with `xbrli`.

**review**:
@Arelle/arelle

### DIFF
--- a/arelle/ModelDocument.py
+++ b/arelle/ModelDocument.py
@@ -420,7 +420,7 @@ def create(modelXbrl, type, uri, schemaRefs=None, isEntry=False, initialXml=None
     elif type == Type.INSTANCE:
         # modelXbrl.uriDir = os.path.dirname(normalizedUri)
         Xml = ('<nsmap>{}'
-               '<xbrl xmlns="http://www.xbrl.org/2003/instance"'
+               '<xbrli:xbrl xmlns:xbrli="http://www.xbrl.org/2003/instance"'
                ' xmlns:link="http://www.xbrl.org/2003/linkbase"'
                ' xmlns:xlink="http://www.w3.org/1999/xlink">').format(initialComment)
         if schemaRefs:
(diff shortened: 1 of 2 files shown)


aviary-wf:
## Security Insights

## Action Items
---

allenbross-wf (dismissed):

arelle/ModelDocument.py
```diff
@@ -420,7 +420,7 @@ def create(modelXbrl, type, uri, schemaRefs=None, isEntry=False, initialXml=None
     elif type == Type.INSTANCE:
         # modelXbrl.uriDir = os.path.dirname(normalizedUri)
         Xml = ('<nsmap>{}'
-               '<xbrl xmlns="http://www.xbrl.org/2003/instance"'
+               '<xbrli:xbrl xmlns:xbrli="http://www.xbrl.org/2003/instance"'
```
austinmatherne-wk, line 423: Not strictly needed, but for consistency it would be good to add the prefix to the closing element on line 429 below:
`Xml += '</xbrli:xbrl></nsmap>'`

austinmatherne-wk (approved):

PR #851, opened 2023-08-29
Add defaultNamespace as an option from the command line

derekgengenbacher-wf:
#### Reason for change
Automatically defaulting to using `xbrli` as the default namespace broke other plugins. So instead add a command line option to InlineXbrlDocumentSet plugin that allows users to set a namespace if they so desire, otherwise default to the default namespace. 

#### Steps to Test
Use any ixbrl document to test the following:
1. run `python -m arelleCmdLine --plugin 'inlineXbrlDocumentSet' -f '[{"ixds":[{"file":"<directory to unzipped ixbrl filing>"}]}]' --saveInstance`
     - This should produce an extracted instance document in the defined directory that uses the default namespace: `<xbrl>`
 2. run 'python -m arelleCmdLine --plugin 'inlineXbrlDocumentSet' -f '[{"ixds":[{"file":"/Users/derekgengenbacher/workspaces/wf/arelle/workiva-2021-2Q-10K"}]}]' --saveInstance --defaultNamespace xbrli'
      - This should produce an extracted instance document in the defined directory that uses the xbrli namespace: `<xbrli:xbrl>`
 3. run `python -m arelleCmdLine --plugin 'inlineXbrlDocumentSet' -f '[{"ixds":[{"file":"/Users/derekgengenbacher/workspaces/wf/arelle/workiva-2021-2Q-10K"}]}]' --saveInstance --defaultNamespace i`
      - This should produce an extracted instance document in the defined directory that uses the i namespace: `<i:xbrl>`

**review**:
@Arelle/arelle

### DIFF
--- a/arelle/ModelDocument.py
+++ b/arelle/ModelDocument.py
@@ -391,7 +391,7 @@ def loadSchemalocatedSchema(modelXbrl, element, relativeUrl, namespace, baseUrl)
             doc.inDTS = False
     return doc
 
-def create(modelXbrl, type, uri, schemaRefs=None, isEntry=False, initialXml=None, initialComment=None, base=None, discover=True, documentEncoding="utf-8") -> ModelDocument:
+def create(modelXbrl, type, uri, schemaRefs=None, isEntry=False, initialXml=None, initialComment=None, base=None, discover=True, documentEncoding="utf-8", xbrliNamespacePrefix=None) -> ModelDocument:
     """Returns a new modelDocument, created from scratch, with any necessary header elements
 
     (such as the schema, instance, or RSS feed top level elements)
@@ -419,14 +419,21 @@ def create(modelXbrl, type, uri, schemaRefs=None, isEntry=False, initialXml=None
         Xml = '<nsmap>{}{}</nsmap>'.format(initialComment or '', initialXml or '')
     elif type == Type.INSTANCE:
         # modelXbrl.uriDir = os.path.dirname(normalizedUri)
+        if xbrliNamespacePrefix is not None:
+            xbrli_instance_namespace = f'<{xbrliNamespacePrefix}:xbrl xmlns:{xbrliNamespacePrefix}="http://www.xbrl.org/2003/instance"'
+        else:
+            xbrli_instance_namespace = '<xbrl xmlns="http://www.xbrl.org/2003/instance"'
         Xml = ('<nsmap>{}'
-               '<xbrli:xbrl xmlns:xbrli="http://www.xbrl.org/2003/instance"'
+               '{}'
                ' xmlns:link="http://www.xbrl.org/2003/linkbase"'
-               ' xmlns:xlink="http://www.w3.org/1999/xlink">').format(initialComment)
+               ' xmlns:xlink="http://www.w3.org/1999/xlink">').format(initialComment, xbrli_instance_namespace)
         if schemaRefs:
             for schemaRef in schemaRefs:
                 Xml += '<link:schemaRef xlink:type="simple" xlink:href="{0}"/>'.format(schemaRef.replace("\\","/"))
-        Xml += '</xbrl></nsmap>'
+        if xbrliNamespacePrefix is not None:
+            Xml += f'</{xbrliNamespacePrefix}:xbrl></nsmap>'
+        else:
+            Xml += '</xbrl></nsmap>'
     elif type == Type.SCHEMA:
         Xml = ('<nsmap>{}<schema xmlns="http://www.w3.org/2001/XMLSchema" /></nsmap>').format(initialComment)
     elif type == Type.RSSFEED:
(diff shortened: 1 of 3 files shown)


aviary3-wk:
## Security Insights

## Action Items
---

arelle/ModelDocument.py
```diff
@@ -391,7 +391,7 @@ def loadSchemalocatedSchema(modelXbrl, element, relativeUrl, namespace, baseUrl)
             doc.inDTS = False
     return doc
 
-def create(modelXbrl, type, uri, schemaRefs=None, isEntry=False, initialXml=None, initialComment=None, base=None, discover=True, documentEncoding="utf-8") -> ModelDocument:
+def create(modelXbrl, type, uri, schemaRefs=None, isEntry=False, initialXml=None, initialComment=None, base=None, discover=True, documentEncoding="utf-8", defaultNamespace='') -> ModelDocument:
```
austinmatherne-wk, line 394: The language here is a little confusing. It's the `xbrliNamespacePrefix`, yes? Also, setting the default value to `None` is more clear for handling the case where we want it to be the default namespace IMO.

```suggestion
def create(modelXbrl, type, uri, schemaRefs=None, isEntry=False, initialXml=None, initialComment=None, base=None, discover=True, documentEncoding="utf-8", xbrliNamespacePrefix=None) -> ModelDocument:
```

arelle/ModelDocument.py
```diff
@@ -419,14 +419,20 @@ def create(modelXbrl, type, uri, schemaRefs=None, isEntry=False, initialXml=None
         Xml = '<nsmap>{}{}</nsmap>'.format(initialComment or '', initialXml or '')
     elif type == Type.INSTANCE:
         # modelXbrl.uriDir = os.path.dirname(normalizedUri)
+        instance_namespace = '<xbrl xmlns="http://www.xbrl.org/2003/instance"'
+        if defaultNamespace:
+            instance_namespace = f'<{defaultNamespace}:xbrl xmlns:{defaultNamespace}="http://www.xbrl.org/2003/instance"'
```
austinmatherne-wk, line 424: nit: use an else block (or ternary) both here and below for the closing tag:
```suggestion
        if xbrliNamespacePrefix:
            instance_namespace = f'<{xbrliNamespacePrefix}:xbrl xmlns:{xbrliNamespacePrefix}="http://www.xbrl.org/2003/instance"'
        else:
            instance_namespace = '<xbrl xmlns="http://www.xbrl.org/2003/instance"'
```

arelle/plugin/inlineXbrlDocumentSet.py
```diff
@@ -514,8 +515,20 @@ def commandLineOptionExtender(parser, *args, **kwargs):
...
     parser.add_option("--encodesavedxmlchars",  # for WEB SERVICE use
                       action="store_true",
-                      dest="encodeSavedXmlChars",
+                      dest="defaultNamespace",
                       help=SUPPRESS_HELP)
+    parser.add_option("--defaultNamespace",
+                      action="store",
+                      default='',
+                      dest="defaultNamespace",
+                      help=_("The namespace to use for the base namespace, defaults to the default namespace"),
```
austinmatherne-wk, line 524: `xbrliNamespacePrefix` and use `None` as the default value.
```suggestion
    parser.add_option("--xbrliNamespacePrefix",
                      action="store",
                      dest="xbrliNamespacePrefix",
                      help=_("The namespace prefix to use for http://www.xbrl.org/2003/instance. It's used as the default namespace when unset."),
```

arelle/plugin/inlineXbrlDocumentSet.py
```diff
@@ -562,6 +575,7 @@ def commandLineFilingStart(cntlr, options, filesource, entrypointFiles, *args, *
 
 
 def commandLineXbrlRun(cntlr, options, modelXbrl, *args, **kwargs):
+    print('GETTING HERE')
```
austinmatherne-wk, line 578: Delete the print statement.

hermfischer:
The parameter name is in need of more specificity.  This feature doesn't actually specify which namespace is defaulted (for example if one wanted to default us-gaap's namespace), but rather which prefix gets applied to the namespace http://www.xbrl.org/2003/instance. So maybe one could name this parameter something like --xbrlInstanceNamespacePrefix or shorter --xbrliPrefix.  Then the default would be --xbrliPrefix "", walter's favorite would be --xbrliPrefix i, and others would prefer --xbrliPrefix xbrli.  Or maybe somebody has a more precise and succinct parameter name thought.

brettkail-wk:
An idea would be `--xmlns http://www.xbrl.org/2003/instance` (current default) or
`--xmlns xbrli=http://www.xbrl.org/2003/instance` (this PR), which would allow adding other prefixes as needed.  If that's too much complexity, we could keep the general form but only allow `http://www.xbrl.org/2003/instance` as a value.

austinmatherne-wk (approved):

allenbross-wf:
QA +10

PR #869, opened 2023-09-22
Prevent package manager from loading schemas in offline mode

### LINKED ISSUE #684: Package manager downloads schemas in offline mode
### What happened?

The package manager ignores the internet connectivity setting. Even in offline mode it will download the XBRL package and base XML schemas if they aren't available in the cache to validate the package.

> http://www.xbrl.org/2016/taxonomy-package-catalog.xsd (offending [line of code](https://github.com/Arelle/Arelle/blob/3867a07d64e074e0708c7e1d2ad1165b55af7a39/arelle/PackageManager.py#L171))
> http://www.xbrl.org/2016/taxonomy-package.xsd (offending [line of code](https://github.com/Arelle/Arelle/blob/3867a07d64e074e0708c7e1d2ad1165b55af7a39/arelle/PackageManager.py#L77))
> http://www.w3.org/2001/03/xml.xsd (offending [line of code](https://github.com/Arelle/Arelle/blob/3867a07d64e074e0708c7e1d2ad1165b55af7a39/arelle/PackageManager.py#L78))

### Documents
[esef_taxonomy_2022.zip](https://github.com/Arelle/Arelle/files/12072084/esef_taxonomy_2022.zip)

python arelleCmdLine.py --file <any> --internetConnectivity offline --packages <any-xbrl-package>

### Interface

CLI (arelleCmdLine), GUI (arelleGUI), Python library (pip install)

### Version

2.9.0

### Download

arelle.org (bundled Python), GitHub release (bundled Python), From source (git checkout), pip install

### Operating System

Linux (Red Hat 7), Linux (Red Hat 8), Linux (Red Hat 9), Linux (Ubuntu 18.04), Linux (Ubuntu 20.04), Linux (Ubuntu 22.04), macOS 10.15 Catalina, macOS 11 Big Sur, macOS 12 Monterey, macOS 13 Ventura, Windows 8.1, Windows 10, Windows 11

aaroncameron-wk:
#### Reason for change
Resolves #684

#### Description of change
Adds a `workingOnlineOrInCache` helper that can be used to check whether the web cache is in offline mode OR if the specified file is already in the cache.

Updates the package manager to make use of this by skipping validation against the schema (and logging a warning explaining this) if the web cache is in offline mode and the schema isn't cached.

#### Steps to Test
_Note: the below steps can be modified accordingly to test via CLI_
- Find a taxonomy package available online as an example (here's one: https://xbrl.sec.gov/2023.zip)
- Make sure Arelle is "online" (not configured to "Work offline")
- Clear cache
- *Confirming package loads successfully when "online"*
  - Confirm the package loads as expected, no warnings/errors
    - Help > Manage packages > On Web > (paste URL) > OK 
  - Confirm files in cache: xml.xsd, taxonomy-package.xsd, taxonomy-package-catalog.xsd, your .zip
  - Remove package 
- *Confirming package loads when cached in "offline" mode* 
  - Enable "Work offline" mode
  - Confirm the package loads as expected, no warnings/errors
  - Remove package 
- *Confirming xml.xsd is not loaded when in "offline" mode*
  - Still in "Work offline" mode 
  - Remove xml.xsd from cache
  - Confirm the package loads as expected, no warnings/errors
  - Confirm xml.xsd not added back to cache
  - Remove package 
- *Confirming  taxonomy-package.xsd, taxonomy-package-catalog.xsd are not loaded when in "offline" mode*
  - Still in "Work offline" mode 
  - Remove taxonomy-package.xsd, taxonomy-package-catalog.xsd from cache (your .zip should be only thing left in cache)
  - Confirm the package loads as expected
  - Confirm warnings are logged for both of the 2 missing schemas
  - Confirm nothing added back to the cache

**review**:
@Arelle/arelle

### DIFF
--- a/arelle/Cntlr.py
+++ b/arelle/Cntlr.py
@@ -604,6 +604,23 @@ class Cntlr:
             pass
         return 0
 
+    def workingOnlineOrInCache(self, url: str) -> bool:
+        """
+        Determines if the given URL should be requested based on the web cache's internet connectivity status
+        and whether the URL already exists in the cache.
+        :param url: Web URL
+        :return: True if the URL should be requested, False if not
+        """
+        if not self.webCache.workOffline:
+            # Working online, can proceed regardless of presence in cache
+            return True
+        cacheFilepath = self.webCache.urlToCacheFilepath(url)
+        if os.path.exists(cacheFilepath):
+            # The file exists in cache, we can proceed despite working offline
+            return True
+        return False
+
+
 def logRefsFileLines(refs: list[dict[str, Any]]) -> str:
     fileLines: defaultdict[Any, set[str]] = defaultdict(set)
     for ref in refs:
--- a/arelle/PackageManager.py
+++ b/arelle/PackageManager.py
@@ -51,6 +51,27 @@ def langCloseness(l1, l2):
             return i
     return _len
 
+
+def _parseFile(cntlr, parser, filepath, file, schemaUrl):
+    """
+    Returns tree from `file`, parsed with `parser`, and validated against the provided schema at `schemaUrl`.
+    :return: Tree if parsed and validated, None if `schemaUrl` could not be loaded.
+    """
+    tree = etree.parse(file, parser=parser)
+    # schema validate tp xml
+    if cntlr.workingOnlineOrInCache(schemaUrl):
+        xsdTree = etree.parse(schemaUrl, parser=parser)
+        etree.XMLSchema(xsdTree).assertValid(tree)
+    else:
+        cntlr.addToLog(_("File could not be validated against the schema (%(schemaUrl)s) because the schema was not "
+                         "found in the cache and Arelle is configured to work offline."),
+                       messageArgs={"schemaUrl": schemaUrl},
+                       messageCode="tpe:workingOffline",
+                       file=filepath,
+                       level=logging.ERROR)
+    return tree
+
+
 def parsePackage(cntlr, filesource, metadataFile, fileBase, errors=[]):
     global ArchiveFileIOError
     if ArchiveFileIOError is None:
@@ -72,10 +93,7 @@ def parsePackage(cntlr, filesource, metadataFile, fileBase, errors=[]):
     _file = filesource.file(metadataFile)[0] # URL in zip, plain file in file system or web
     parser = lxmlResolvingParser(cntlr)
     try:
-        tree = etree.parse(_file,parser=parser)
-        # schema validate tp xml
-        xsdTree = etree.parse(TP_XSD,parser=parser)
-        etree.XMLSchema(xsdTree).assertValid(tree)
+        tree = _parseFile(cntlr, parser, metadataFile, _file, TP_XSD)
     except (etree.XMLSyntaxError, etree.DocumentInvalid) as err:
         cntlr.addToLog(_("Taxonomy package file syntax error %(error)s"),
                        messageArgs={"error": str(err)},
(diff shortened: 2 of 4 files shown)


aviary3-wk:
## Security Insights

## Action Items
---

derekgengenbacher-wf (approved):

arelle/PackageManager.py
```diff
@@ -51,6 +51,27 @@ def langCloseness(l1, l2):
...
+    if cntlr.workingOnlineOrInCache(schemaUrl):
+        xsdTree = etree.parse(schemaUrl, parser=parser)
+        etree.XMLSchema(xsdTree).assertValid(tree)
+    else:
+        cntlr.addToLog(_("File could not be validated against the schema (%(schemaUrl)s) because the schema was not "
+                         "found in the cache and Arelle is configured to work offline."),
+                       messageArgs={"schemaUrl": schemaUrl},
+                       messageCode="tpe:workingOffline",
+                       file=filepath,
+                       level=logging.WARNING)
```
brettkail-wk, line 71: I wonder if this should be an ERROR?  If you're offline without a taxonomy-package.xsd, then we can't know whether the taxonomy package is valid or not.

arelle/PackageManager.py
```diff
@@ -72,10 +93,7 @@ def parsePackage(cntlr, filesource, metadataFile, fileBase, errors=[]):
     _file = filesource.file(metadataFile)[0] # URL in zip, plain file in file system or web
     parser = lxmlResolvingParser(cntlr)
     try:
-        tree = etree.parse(_file,parser=parser)
-        # schema validate tp xml
-        xsdTree = etree.parse(TP_XSD,parser=parser)
-        etree.XMLSchema(xsdTree).assertValid(tree)
+        tree = _parseFile(cntlr, parser, metadataFile, _file, TP_XSD)
```
brettkail-wk, line 96: What happens if you have taxonomy-package.xsd in the cache but are missing xml.xsd (`schemaRef`)?

arelle/PackageManager.py
aaroncameron-wk: The unknown aspect of that is why I opted for WARNING, we don't know there's an error. I don't know enough about the UX to say for sure, but my thinking was that at this point in the UX, the user isn't looking to validate the package, just load it into Arelle. I interpreted the validation here as a courtesy to give a "heads-up" what's been provided doesn't look like a valid taxonomy package, but I suppose the fact that validation errors [cause package manager to abort loading](https://github.com/Arelle/Arelle/blob/8a882f2fb2047d7bdd5e9249746e980b3085b5bb/arelle/PackageManager.py#L86) doesn't really align with that. I'm fine with bumping up the log level, but then I think the bigger question becomes: do we prevent the package from loading if we can't validate it?

brettkail-wk: > at this point in the UX, the user isn't looking to validate the package

If that's true, then the schema validation is misplaced, and it should be wherever validations are found, but the presence of other `errors`/`cntlr.addToLog(level=ERROR)` just below make me think that this is the place in Arelle where invalid taxonomy package structures are caught.

I don't think it needs to be a fatal error that prevents opening, but it seems like it should be an ERROR level.

arelle/PackageManager.py
aaroncameron-wk: xml.xsd isn't loaded into the cache, but validation still completes successfully.

brettkail-wk: Thanks.  I knew xml.xsd was optionalish, but I wasn't sure at what level it would be ignored.

austinmatherne-wk: From the command line I get:

> [Exception] Entry point loading Failed to complete request:
Element '{http://www.w3.org/2001/XMLSchema}import': Failed to parse the XML resource 'http://www.w3.org/2001/xml.xsd'., line 16

arelle/Cntlr.py
```diff
@@ -619,6 +619,23 @@ def memoryUsed(self) -> float | int:
...
         return 0
 
+    def workingOnlineOrInCache(self, url: str) -> bool:
+        """
+        Determines if the given URL should be requested based on the web cache's internet connectivity status
+        and whether the URL already exists in the cache.
+        :param url: Web URL
+        :return: True if the URL should be requested, False if not
+        """
+        if not self.webCache.workOffline:
```
austinmatherne-wk, line 629: At the moment this doesn't behave as expected from the command line. The work offline setting is parsed after taxonomy packages, so the default setting (online) is always set at this point. Switching the order that those two cli args are handled should resolve that that.

arelle/PackageManager.py
austinmatherne-wk: From the command line I get:

> [Exception] Entry point loading Failed to complete request:
Element '{http://www.w3.org/2001/XMLSchema}import': Failed to parse the XML resource 'http://www.w3.org/2001/xml.xsd'., line 16

arelle/Cntlr.py
aaroncameron-wk: Moved `internetConnectivity` option handling up.

arelle/PackageManager.py
austinmatherne-wk: Discussed offline. The error I ran into doesn't appear to be caused by the package manager.

austinmatherne-wk (approved):

derekgengenbacher-wf (approved):

austinmatherne-wk (approved):

PR #875, opened 2023-09-26
NL FR KVK 2.03, 5.01, 5.02 validations

aaroncameron-wk:
#### Description of change
- FR-KVK-2.03: The attribute 'href' of the 'link:schemaRef' element MUST refer to the full web location of the published entrypoint from the annual reporting taxonomy.
- FR-KVK-5.01: The attribute 'decimals' to monetary values MUST be filled with allowed values.
- FR-KVK-5.02: The attribute 'decimals' for non-monetary numeric facts MUST be filled with 'INF'.

#### Steps to Test
```
python arelleCmdLine.py -f ./tests/resources/validation/NL/example/a.xbrl 
--plugins="validate/NL" --validate --disclosureSystem NT17-preview
```
Result: no NL validation errors
___
```
python arelleCmdLine.py -f ./tests/resources/validation/NL/fr_kvk/2.03/fail/invalid.xbrl
--plugins="validate/NL" --validate --disclosureSystem NT17-preview
```
Result: Errors include: `[NL.FR-KVK-2.03]`
___

```
python arelleCmdLine.py -f ./tests/resources/validation/NL/fr_kvk/5.01/fail/invalid.xbrl
--plugins="validate/NL" --validate --disclosureSystem NT17-preview
```
Result: Errors include: `[NL.FR-KVK-5.01]`
___

```
python arelleCmdLine.py -f ./tests/resources/validation/NL/fr_kvk/5.02/fail/invalid.xbrl
--plugins="validate/NL" --validate --disclosureSystem NT17-preview
```
Result: Errors include: `[NL.FR-KVK-5.02]`
___

**review**:
@Arelle/arelle


aviary-wf:
## Security Insights

## Action Items
---

arelle/plugin/validate/NL/rules/fr_kvk.py
```diff
@@ -80,3 +82,104 @@ def rule_fr_kvk_2_01(
...
+                if "href" not in docRef.referenceTypes:
+                    continue
+                if docRef.referringModelObject.localName != "schemaRef":
+                    continue
+                href = refDoc.uri
+                ext = None
+                if href:
+                    path = urlparse(href.lower()).path
+                    ext = os.path.splitext(path)[1]
+                if not UrlUtil.isAbsolute(href) or ext != '.xsd':
```
brettkail-wk, line 117: It seems like the disclosure system should have the set of supported set of XSDs for that version rather than just heuristically checking the shape of the URI.  e.g., EFM checks for specific URLs to ensure you're not referencing something too old.

aaroncameron-wk: I've just pushed a fairly simple solution, but I did notice that EFM uses a more complicated [XML format](https://github.com/Arelle/Arelle/blob/72aa068e8b7e53d59f1a3f3d371f7fec25a994f2/arelle/plugin/validate/EFM/resources/edgartaxonomies/edgartaxonomies-23-3.xml) to [define attributes](https://github.com/Arelle/Arelle/blob/72aa068e8b7e53d59f1a3f3d371f7fec25a994f2/arelle/DisclosureSystem.py#L272-L326) associated with entry points. It seems fairly specific to EFM, I didn't see any other built-in validation plugins making use of this design, and I don't see any immediate value-add of using that design over this simple solution.

brettkail-wk: The advantage is `<schemaRef href="http://mycompany.com/myextensiontaxonomy.xsd"/>` passes your filter, but the rule says it shouldn't.  The rule says you must use a pre-approved entry points, not an arbitrary one that happens to be an absolute URL that ends with `.xsd`.

aaroncameron-wk: I pushed a [simple solution](https://github.com/Arelle/Arelle/pull/875/commits/b459e9816fe1bbb516593e886de18e0fb9c8c1aa) (but not as simple as my initial solution) with my last comment that does check against pre-approved entry points. That comment was more about whether or not we should make use of the same XML-based approach that EFM uses, or if this simple solution checking against a predefined list is sufficient for this validation.

brettkail-wk: Oh, sorry for the misunderstand.  I have no idea, sorry :-).

derekgengenbacher-wf (approved):

allenbross-wf:
All tests pass.

PR #876, opened 2023-09-26
NL BR KVK validations (2.04, 3.01, 4.07, 4.10, 4.16)

aaroncameron-wk:
#### Description
BR-KVK-2.04: The period in the context MUST correspond to the period of the current or
previous financial reporting period, or one day before the start of the comparative financial
where the context must be instant.

Example:
Reporting Period Start Date: 2018-01-01
Reporting Period End Date: 2018-12-31

| Context | Instant    | Start Date | End Date   | Note                                                    |
| ------- | ---------- | ---------- | ---------- | ------------------------------------------------------- |
| 1       | 2018-12-31 |            |            | End value as context for the current reporting period   |
| 2       |            | 2018-12-31 | 2018-12-31 | Current reporting period                                |
| 3       | 2017-12-31 |            |            | End value as context for previous reporting period      |
| 4       |            | 2017-12-31 | 2017-12-31 | Previous reporting period                               |
| 5       | 2016-12-31 |            |            | Starting value as context for previous reporting period |

___
BR-KVK-3.01: A measure element with a namespace prefix that refers to the
"http://www.xbrl.org/2003/iso4217" namespace MUST appear exactly once in the instance document.

___
BR-KVK-4.07: The jenv-bw2-i:FinancialReportingPeriodCurrentEndDate MUST be before the date of filing.

___
BR-KVK-4.10: The jenv-bw2-i:DocumentAdoptionDate MUST NOT be after the date of filing.

___
BR-KVK-4.16: A corrected financial statement MUST be established.
If kvk-i:DocumentResubmissionDueToUnsurmountableInaccuracies is "Ja" (Yes),
the following facts must be filled for a corrected financial statement:
jenv-bw2-i:DocumentAdoptionStatus
jenv-bw2-i:DocumentAdoptionDate

#### Steps to Test
Run validation on each failure example document using NT16 disclosure system and confirm that the associated error is logged.

Run validation on `nt16.xbrl` and `nt17.xbrl` and confirm no NL validation plugin errors are logged.
___
**review**:
@Arelle/arelle


aviary-wf:
## Security Insights

## Action Items
---

arelle/plugin/validate/NL/rules/br_kvk.py
```diff
@@ -0,0 +1,251 @@
...
+    validDurations = (currentDuration, previousDuration)
+    validInstants = (
+        currentDuration[1],
+        (currentDuration[0] + timedelta(days=-1)),
+        (previousDuration[0] + timedelta(days=-1))
+    )
+
+    for contextId, context in modelXbrl.contexts.items():
+        if context.isInstantPeriod:
+            # instantDatetime getter adds a day, we need to subtract for comparison
```
brettkail-wk, line 93: It seems like it would be clearer to add `instantDate` and `endDate` helpers that subtract a day if midnight or else return the day (per [4.7.2](https://www.xbrl.org/Specification/XBRL-2.1/REC-2003-12-31/XBRL-2.1-REC-2003-12-31+corrected-errata-2013-02-20.html#_4.7.2)).

arelle/plugin/validate/NL/rules/br_kvk.py
```diff
@@ -0,0 +1,251 @@
...
+if TYPE_CHECKING:
+    from arelle import ModelXbrl
+    from arelle.ModelValue import QName
+
+_: TypeGetText
+
+
+def _getReportingPeriodDateValue(modelXbrl: ModelXbrl, qname: QName) -> date:
+    facts = modelXbrl.factsByQname.get(qname)
+    assert facts and len(facts) == 1, _('Exactly one fact must exist for reporting period concept: {0}').format(qname)
```
brettkail-wk, line 38: It seems wrong to raise an exception if there are no facts, if there are duplicate facts (I assume there's a separate rule that intends to diagnose that, or if not, we should somehow warn once if we find duplicate period facts?), or the fact has an invalid value.

aaroncameron-wk: There are separate checks for making sure there are exactly one of most of concepts involved in these validations. Would the best practice be to have these validations fail silently and let those other checks draw attention to any missing/duplicate fact issues? Or should an issue like this that prevents the actual logic of the validation from running trigger the failure of that validation in the same way?

arelle/plugin/validate/NL/rules/br_kvk.py
aaroncameron-wk: I believe the helpers just need to *not* pass in `addOneDay`, I've implemented them.

arelle/plugin/validate/NL/rules/br_kvk.py
brettkail-wk: > There are separate checks for making sure there are exactly one of most of concepts involved in these validations.

Which ones don't have a "missing" separate check?  If it's previous period (e.g., new company?), perhaps we just need to tolerate the facts as missing (i.e., current and current-1day rather than current,previous,previous-1day).  Skipping this validation for missing mandatory facts seems reasonable.

Which ones don't have a separate "duplicate" check?  I'm not sure what to do since XBRL doesn't forbid duplicate facts ([notes](https://www.xbrl.org/WGN/xbrl-duplicates/WGN-2015-12-09/xbrl-duplicates-WGN-2015-12-09.html#sec-approaches)), so unless their manual does, it's unclear what the regulator enforces (no duplicates anywhere, no duplicates for these facts, or allow _all_ values as valid).  Perhaps start by adding duplicate checking for the key facts in this rule, report them with this rule's error code, and add a comment explaining we're not sure which behavior is actually intended.

If any single key fact value is invalid, I'd say skip the validation.  There's no use in reporting an issue for every other fact in the report saying they have invalid context just because some key fact has a typo.

arelle/ModelInstanceObject.py
```diff
@@ -941,6 +944,17 @@ def startDatetime(self, value):
...
 
+    @property
+    def endDate(self) -> date:
+        """
+        :return: endDate or instant attribute as date, *not* adjusted by a day for midnight values
+        """
+        try:
+            return self._endDate
+        except AttributeError:
+            self._endDate = XmlUtil.datetimeValue(XmlUtil.child(self.period, XbrlConst.xbrli, ("endDate", "instant"))).date()
```
brettkail-wk, line 955: Does this return `2022-12-31` for `2023-01-01T00:00:00`?  I believe it should, so I think start with `endDatetime` but subtract one day if (hour,minute,second,microsecond)==(0,0,0,0).  Similarly for `instantDate`.

arelle/plugin/validate/NL/rules/br_kvk.py
brettkail-wk: Follow up [here](https://github.com/Arelle/Arelle/pull/876#discussion_r1338497169).

arelle/plugin/validate/NL/rules/br_kvk.py
austinmatherne-wk: Agreed with Brett. Further, these validation rules should never raise an exception. They should either pass or fail (by returning one or more validation warnings/errors).

aaroncameron-wk: It might be worthwhile to have a certain exception type (could be `AssertionError`, could be something specific to this validation plugin design) that can be thrown from validations and then be logged as a debug message and ignored. 

That way rather than having to, for example, modify `_getReportingPeriodDateValue` to potentially return `None`, then handle the `None` case to gracefully exit the validation, we can just immediately `raise IncompleteValidationError('Exactly one fact must exist for...')` once we know the validation won't be possible to check. I suppose this only really matters if we often make use of helper functions, otherwise it's no better than having a `return` statement. Thoughts?

brettkail-wk: Seems plausible.  It might be nice if the exception contained a code+message so that we could fail harder if the code was not produced elsewhere/later (even if just as part of testing; e.g., a new plugin callback for handling IncompleteValidationError so the test harness can collect them until the end).

arelle/plugin/validate/NL/rules/__init__.py
```diff
@@ -0,0 +1,13 @@
+from arelle.ModelValue import qname
+
+
+JENV_BW2_DATA_NS = 'http://www.nltaxonomie.nl/nt16/jenv/20211208/dictionary/jenv-bw2-data'
```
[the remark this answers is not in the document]
aaroncameron-wk: Implemented this within NL's `PluginValidationDataExtension`

derekgengenbacher-wf (approved):

allenbross-wf (approved):

allenbross-wf:
QA +10

PR #910, opened 2023-10-12
Download dedicated caches for conformance suites in CLI

aaroncameron-wk:
#### Reason for change
The current method of pre-populating caches on CLI runners testing conformance suites is not ideal:
- Each runner downloads the same all-inclusive cache.
- The service providing the cache download won't be updated base solely on Arelle's CI needs.

#### Description of change
Instead of downloading an all-inclusive cache archives, download individual cache archives dedicated to each conformance suite. There will be duplication between these cache's but it's preferable to download and unpack a single cache rather than try to combine from different sources.

We already make use of a private AWS S3 bucket to download public and private conformance suites into the runners. We can also use this bucket to host the dedicate cache archives.

Initially, the runners will check if a cache archive matching the current test's name exists. If it doesn't, the runner will fall back to downloading the current all-inclusive cache. We can remove this condition if we eventually build dedicated caches for every conformance suite.

#### Steps to Test
Won't run on this PR due to workflow being `pull_request_target`.

**review**:
@Arelle/arelle


aviary2-wf:
## Security Insights

## Action Items
---

derekgengenbacher-wf (approved):

PR #920, opened 2023-10-19
Add NL.FR-NL-1.01 testcase

aaroncameron-wk:
#### Reason for change
FR-NL-1.01: A BOM character MUST NOT be used
A BOM (Byte Order Mark) is a unicode character that represents the byte order in a file:
16 or 32 bits. Within the UTF-8 standard, this is an optional, but unnecessary, character (because the 8
stands for 8 bits).
This rule cannot be implemented using XBRL Formula since the BOM
character is a physical document attribute that is not available in the logical document model
used by XPath.

#### Description of change
Like the validation checking for a CDATA end sequence, we shouldn't rely on a parsed XML model to find this character, as there's no guarantee that XML parsing won't trim/ignore it wherever it may be included in a document. This PR makes a reusable `findDisallowedPatterns` function (with a cached result so the instance documents don't need to be loaded multiple times) to check for the CDATA sequence, a BOM character, or any other pattern that makes sense to be checked for this way.

#### Steps to Test
CI

**review**:
@Arelle/arelle


aviary-wf:
## Security Insights

## Action Items
---

arelle/plugin/validate/NL/PluginValidationDataExtension.py
```diff
@@ -20,3 +34,47 @@ class PluginValidationDataExtension(PluginValidationData):
...
+        :param modelXbrl: Model containing instance documents to check for disallowed patterns
+        :return: List of tuples (basename, line number, matched pattern)
+        """
+        if self.disallowed_patterns_found is not None:
+            return self.disallowed_patterns_found
+        pattern = re.compile(DISALLOWED_PATTERN_REGEX)
+        disallowed_patterns_found = []
+        for doc in modelXbrl.urlDocs.values():
+            if doc.type == ModelDocument.Type.INSTANCE:
+                file = modelXbrl.fileSource.file(doc.filepath, encoding='utf-8')[0]
```
brettkail-wk, line 75: Pre-existing, I think `file` needs to be closed (can we use `with`?).

arelle/plugin/validate/NL/rules/fr_nl.py
```diff
@@ -17,11 +15,40 @@
...
+        pluginData: PluginValidationDataExtension,
+        val: ValidateXbrl,
+        *args: Any,
+        **kwargs: Any,
+) -> Iterable[Validation] | None:
+    """
+    FR-NL-1.01: A BOM character MUST NOT be used.
+    """
+    for basename, lineNumber, match in pluginData.findDisallowedPatterns(val.modelXbrl):
+        matchEncoded = match.encode()
```
brettkail-wk, line 41: Perhaps `findDisallowedPatterns` should do the `encode()`?

I'm somewhat confused how this all works.  It seems like we're reading as bytes, running a regex on bytes, so the match results are in bytes, so we need to encode here in order to compare against the str literal, is that right?  If so, why isn't `encode()` needed in `rule_fr_nl_2_06`?

arelle/plugin/validate/NL/PluginValidationDataExtension.py
```diff
@@ -20,3 +34,47 @@ class PluginValidationDataExtension(PluginValidationData):
...
+                for i, line in enumerate(file):
+                    for match in re.findall(pattern, line):
```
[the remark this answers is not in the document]
aaroncameron-wk: I wasn't familiar with BOM before but researching more I think you're right.

arelle/plugin/validate/NL/rules/fr_nl.py
aaroncameron-wk: Per Steven's comment, ended up not going with the approach of checking for BOM throughout the file. It seems BOM is, by definition, a character at the beginning of a text stream. So the whole regex/pattern matching part of this is abandoned in favor of simply checking the first character.

arelle/plugin/validate/NL/rules/fr_nl.py
```diff
@@ -21,6 +23,55 @@
...
+) -> Iterable[Validation] | None:
+    """
+    FR-NL-1.01: A BOM character MUST NOT be used.
+    """
+    modelXbrl = val.modelXbrl
+    encoding = 'utf-8'
+    for doc in modelXbrl.urlDocs.values():
+        if doc.type == ModelDocument.Type.INSTANCE:
+            with modelXbrl.fileSource.file(doc.filepath, encoding=encoding)[0] as file:
+                firstByte = file.read(1).encode(encoding)
```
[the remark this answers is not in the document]
aaroncameron-wk: Looks like I was overly concerned about the unexpected UTF16/32 BOMs, but after testing, Arelle would fail long before this validation is reached. Empty files as well (I don't know if they'd qualify as INSTANCE type docs anyway).

> the underlying openXmlFileStream and XmlUtil.encoding do much more than we need

My approach uses the same method of loading the file, but I don't know if I want to rely on the `encoding` returned as an indication of whether or not a BOM exists, since [XmlUtil.encoding](https://github.com/Arelle/Arelle/blob/2e79b956186c633bf9fc728ad1a536da803a4a04/arelle/XmlUtil.py#L147-L171) can feasibly return values other than `utf-8-sig` when a BOM exists.

allenbross-wf (approved):

allenbross-wf:
QA +10 CI Passed successfully

PR #926, opened 2023-10-25
Add NL.BR-KVK-4.20 testcase

aaroncameron-wk:
#### Reason for change
BR-KVK-4.20: Each date in an XBRL instance MUST be a valid date. The date is in format YYYY-MM-DD.

#### Description of change
Arelle's date parsing performs all the necessary logic internally. This validation simply finds date elements with `xValid == INVALID` and fires a validation error. I added `xValueError` so that the actual error message can be captured for inclusion int he validation error message.

Separately, I've started work on capturing value parsing errors in a more global way, rather than specifically checking for period date values then date fact values (and potentially missing somewhere else where invalid dates could show up), a better solution would fire a validation error wherever an invalid date is found regardless of context. I felt this would be too large of an internal change for this PR, so the included approach is what I intend to move forward with for now.

#### Steps to Test
CI

**review**:
@Arelle/arelle


aviary2-wf:
## Security Insights

## Action Items
---

arelle/plugin/validate/NL/rules/br_kvk.py
```diff
@@ -282,3 +283,52 @@ def rule_br_kvk_4_16(
...
+    ],
+)
+def rule_br_kvk_4_20(
+        pluginData: PluginValidationDataExtension,
+        val: ValidateXbrl,
+        *args: Any,
+        **kwargs: Any,
+) -> Iterable[Validation] | None:
+    """
+    BR-KVK-4.20: Each date in an XBRL instance MUST be a valid date. The date is in format YYYY-MM-DD.
```
brettkail-wk, line 302: > Arelle's date parsing performs all the necessary logic internally.

I assume Arelle's date parsing logic allows the full lexical space of [dates](https://www.w3.org/TR/xmlschema-2/#date-lexical-representation) (`'-'? yyyy '-' mm '-' dd zzzzzz?`) and [dateTimes](https://www.w3.org/TR/xmlschema-2/#dateTime-lexical-representation) (`'-'? yyyy '-' mm '-' dd 'T' hh ':' mm ':' ss ('.' s+)? (zzzzzz)?`), which both allow 4+ digit years and timezones.  Do we think it's trying to constraint the period start/end or just date facts?  If period/start, is it trying to limit to full days (no times)?  Do we think the `YYYY-MM-DD` is colloquial, or is it trying to require 4-digit years and no timezone?

allenbross-wf (approved):

arelle/plugin/validate/NL/rules/br_kvk.py
aaroncameron-wk: > I assume Arelle's date parsing logic allows the full lexical space of [dates](https://www.w3.org/TR/xmlschema-2/#date-lexical-representation) ('-'? yyyy '-' mm '-' dd zzzzzz?) and [dateTimes](https://www.w3.org/TR/xmlschema-2/#dateTime-lexical-representation) ('-'? yyyy '-' mm '-' dd 'T' hh ':' mm ':' ss ('.' s+)? (zzzzzz)?), which both allow 4+ digit years and timezones.

The full text of the rule goes into specific detail about the valid values for the various *date* parts, but makes no mention of *time* parts, so I'm inclined to believe this rule is only meant to apply to `date` values and says nothing about `dateTime` values (They might not define any `dateTime` concepts, which would explain the lack of rules applying to them). This is why the validation is set up to only check `dateItemType` fact values.

> Do we think it's trying to constraint the period start/end or just date facts? If period/start, is it trying to limit to full days (no times)?

I decided to include period values because of the translated wording "Each date in an XBRL instance..." not seeming specific to facts. Plus, context clues from other rules suggest, if not implicitly require, that the period values should be dates, particularly:
> BR-KVK-2.04: The period in the context MUST correspond to the period of the current or previous financial reporting period, or one day before the start of the comparative financial year where the context must be instant.

Where "financial reporting period" values are `dateItemType` values, so any period dates corresponding with those values should also be able to be represented with a pure date value (but I suppose a period date value of YYYY-MM-DDT00:00:00 could still satisfy BR-KVK-2.04 while being in apparent violation of this rule)

> Do we think the YYYY-MM-DD is colloquial, or is it trying to require 4-digit years and no timezone?

My initial approach was trusting that Arelle knows best on what date values should be allowed, and that this rule was more about making it clear that other formats (DD/MM/YYYY) would not be allowed. I suppose there could be some reason that whatever KvK is using to parse the XBRL may strictly require the listed format without exception, so I'll add an explicit check for this pattern on top of checking for Arelle's parsing errors. In general, I think we'll more quickly discover issues with our validations if they're too strict rather than if they're too lenient.

derekgengenbacher-wf (approved):

allenbross-wf (dismissed):

derekgengenbacher-wf (approved):

derekgengenbacher-wf:

PR #952, opened 2023-11-03
FR-NL-5.11 validation

aaroncameron-wk:
**Note: this PR updates existing NT18 testcases to import NT18 files. I will be adding an `nl_nt18.zip` cache to S3 as part of satisfying CI for this PR**
___

#### Reason for change
FR-NL-5.11: Formatting using a limited set of 'escaped XHTML' elements MAY ONLY be included in fact values of facts tagged with concepts of type 'formattedExplanationItemType'.

Only certain tags are allowed (see manual). "ul" and "ol" tags have a restriction on the "type" attribute".

Any valid tag may have a style attribute, but only certain style properties may be used. (font-family, font-size, color)

The manual specifies general readabilty in styling, but that is not practically enforced here.
    - No font smaller than "3" (font size can be altered in many ways that depend on browser support)
    - No white text on white background (many potential "color" values are not easily readable)

#### Description of change
Enforce XHTML content only being allowed in facts of certain types by attempting to parse the content of any fact. If the parsing yields an element, that triggers an error.

Then, where XHTML inclusion is allowed, compares against valid tags and style attributes defined in the manual. Importantly, the validation does not check against "readability" requirements defined in the manual (font size and color) because it's not practical to enforce the various ways font size and color can be modified.

#### Steps to Test
CI, after I've added `nl_nt18.zip` cache to S3.

**review**:
@Arelle/arelle


aviary2-wf:
## Security Insights

## Action Items
---

arelle/plugin/validate/NL/rules/fr_nl.py
```diff
@@ -436,6 +451,88 @@ def rule_fr_nl_5_06(
...
+    - No white text on white background (many potential "color" values are not easily readable)
+    """
+    invalidTypeFacts = []
+    for fact in val.modelXbrl.facts:
+        validType = fact.concept.instanceOfType(pluginData.formattedExplanationItemTypeQn)
+        wrappedContent = f'<div>{fact.textValue}</div>'  # Single root element required, wrap in valid div element
+        invalidTags = set()
+        invalidStyles = set()
+        hasElts = False
+        for elt in etree.fromstring(wrappedContent).iter():
```
brettkail-wk, line 484: Does this throw an exception if an arbitrary string fact doesn't parse as XML, e.g., `H&R Block`?

arelle/plugin/validate/NL/rules/fr_nl.py
```diff
@@ -436,6 +451,88 @@ def rule_fr_nl_5_06(
...
+                msg=_('Only a limited set of style properties (%(allowedStyles)s) are allowed in escaped XHTML. '
+                      'Found invalid style properties: %(invalidStyles)s'),
+                allowedStyles=', '.join(sorted(XHTML_ALLOWED_STYLES)),
+                invalidStyles=sorted(invalidStyles),
+                modelObject=fact
+            )
+    if len(invalidTypeFacts) > 0:
+        yield Validation.error(
+            codes='NL.FR-NL-5.11',
+            msg=_('Formatting using "escaped XHTML" elements MAY ONLY be included in '
```
brettkail-wk, line 530: It seems wrong to "speculatively" parse all facts rather than only `formattedExplanationItemType`.  It seems like this should be at most a warning if not removed altogether...

aaroncameron-wk: I downgraded to a warning. Based on the wording of the rule, I think it's important that the validation check and communicate in some way that certain elements shouldn't have XHTML, but I agree an error is heavy handed based on the speculation.

arelle/plugin/validate/NL/rules/fr_nl.py
aaroncameron-wk: I've updated this to yield a warning when a fact of type `formattedExplanationItemType` fails parsing and thus can't have it's XHTML contents validated.

allenbross-wf (approved):

derekgengenbacher-wf (approved):

allenbross-wf:
QA +10
Tests ran successfully.

PR #977, opened 2023-11-28
Implement built-in cache for effectively permanent files

### LINKED ISSUE #793: Bundle base schemas into frozen builds

A collection of base schemas from w3.org (xml.xsd and XMLSchema.xsd) and xbrl.org (specifically the taxonomy package schemas) are widely used by Arelle. To cater more effectively to users who wish to operate Arelle in offline mode by using taxonomy packages we should incorporate these schemas (and possibly others from xbrl.org) directly into the precompiled frozen builds.

aaroncameron-wk:
#### Reason for change
Closes #793 
Users using offline mode have to populate their cache ahead of time with common w3.org or xbrl.org files. As a matter of convenience, we can include these directly in the application since we don't expect them to change.

#### Description of change
Refactoring of the `WebCache.getfilename` function for improved readability.
Addition of a step that uses a built-in cache at `/arelle/resources/cache` if a corresponding file exists there.

#### Steps to Test
Run Arelle in offline mode and confirm operations requiring built-in cache files use them successfully.
Run Arelle in online mode and confirm operations requiring built-in cache files do not download them into the system cache and instead use the built-in cache.

**review**:
@Arelle/arelle


arelle/resources/cache/http/www.xbrl.org/2013/inlineXBRL/xhtml-inlinexbrl-1_1.xsd
```diff
@@ -0,0 +1,40 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
austinmatherne-wk, line 1: We should include both the 1.1 (http://www.xbrl.org/2013/inlineXBRL/xhtml-inlinexbrl-1_1.xsd) and 1.0 (http://www.xbrl.org/2008/inlineXBRL/xhtml-inlinexbrl-1_0.xsd) inline schemas.

Also, I'm inclined to think these should work for both http and https URLs.

aaroncameron-wk: If enabled, the `httpRedirect` feature would apply to the built-in cache as well. But if you're suggesting that we should have http/https be essentially equivalent by default in the built-in cache, I can make it so.

austinmatherne-wk (approved):

arelle/WebCache.py
```diff
@@ -113,6 +113,7 @@ def __init__(
                 self.encodeFileChars = re.compile(r'[<>:"\\|?*^]')
             else:
                 self.encodeFileChars = re.compile(r'[:^]')
+        self.builtInCacheDir = os.path.abspath(os.path.join(__file__, '..', 'resources', 'cache'))
```
austinmatherne-wk, line 116: I'm assuming this doesn't work in the frozen builds.

[distro.py needs to be configured](https://github.com/Arelle/Arelle/blob/d210baf36f5c161852639c2169d8c5fb0cb1ba03/distro.py#L16-L24) to include the files and then the correct path has to be [determined based on the environment](https://github.com/Arelle/Arelle/blob/d210baf36f5c161852639c2169d8c5fb0cb1ba03/arelle/Cntlr.py#L169-L170).

PR #987, opened 2023-12-06
Replace packages based off of package identifier

### LINKED ISSUE #964: Arelle silently replaces taxonomy packages based on name and version
### What happened?

If Arelle encounters two taxonomy packages with the same name and version information in `taxonomyPackage.xml`, then the most recently encountered one will silently replace the previous one.

I would expect:

* A warning message to be logged.
* Packages to be keyed on the `identifier` field, not name and version.  

I tripped over this because I was trying to untangle some badly packaged taxonomies and accidentally ended up with two different packages with an identical `taxonomyPackage.xml` file.  Clearly this is a mistake on my part, but equally, encountering two packages with the same metadata should trigger at least a warning, unless we've actually verified that the whole package is the same (e.g. using hashes).

Code is [here](https://github.com/Arelle/Arelle/blob/master/arelle/PackageManager.py#L631)

### Documents

_No response_

_No response_

### Interface

CLI (arelleCmdLine)

### Version

941f64be

### Download

From source (git checkout)

### Operating System

Linux (Ubuntu 20.04), Other (please specify)

aaroncameron-wk:
#### Reason for change
Resolves #964 

#### Description of change
Load "identifier" into package config. Use it instead of "name" and "version" when determining whether or not to replace or add a package.

#### Steps to Test
Create two identical packages A and B locally. In taxonomy package B, modify the `identifier` in `taxonomyPackage.xml`.

Use this command:
```
python arelleCmdLine.py --packages="path to package A.zip" --packages="path to package B.zip" -f test.xsd
```

Before this change, running this command will result in only one package ending up loaded into `packagesConfig` (since they still have matching `name` and `version` values).
After this change, two packages will end up loaded into `packagesConfig` (since they have different `identifier` values).

**review**:
@Arelle/arelle

### DIFF
--- a/arelle/PackageManager.py
+++ b/arelle/PackageManager.py
@@ -543,6 +543,7 @@ def packageInfo(cntlr, URL, reload=False, packageManifestName=None, errors=[]):
                 return None
             package = {'name': ", ".join(packageNames),
                        'status': 'enabled',
+                       'identifier': parsedPackage.get('identifier'),
                        'version': parsedPackage.get('version'),
                        'license': parsedPackage.get('license'),
                        'fileDate': time.strftime('%Y-%m-%dT%H:%M:%S UTC', fileDateTuple),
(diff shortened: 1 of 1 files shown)


aviary2-wf:
## Security Insights

## Action Items
---

derekgengenbacher-wf (approved):

austinmatherne-wk (approved):
Testing steps should include making sure packages enabled in the GUI in older versions of Arelle don't break when updating.

PR #1010, opened 2023-12-21
Multi-Doc GUI loaded dimension errors

hermfischer-wf:
#### Reason for change
GUI-loaded multiple documents (such as SEC pri doc and fee exhibit or EX-26) from same zip or directory caused secondary IXDS to have contexts with dimension contexts attempted to be resolved from primary document, causing spurious xbrldi dimension invalidity errors.

#### Description of change
Assign multi-target and multi-IXDS contexts to consuming IXDS and target instance.

#### Steps to Test
Load sec test suite multi-doc examples (EX-26 or fee exhibit) via GUI.
Check all relevant test suites for any newly-introduced regressions (IX-11 suite, EFM suite, ESEF suite...)

**review**:
@Arelle/arelle

### DIFF
--- a/arelle/ModelDocument.py
+++ b/arelle/ModelDocument.py
@@ -178,11 +178,16 @@ def load(modelXbrl, uri, base=None, referringElement=None, isEntry=False, isDisc
         if not isIncluded and namespace and namespace in XbrlConst.standardNamespaceSchemaLocations and uri != XbrlConst.standardNamespaceSchemaLocations[namespace]:
             return load(modelXbrl, XbrlConst.standardNamespaceSchemaLocations[namespace],
                         base, referringElement, isEntry, isDiscovered, isIncluded, namespace, reloadCache)
+        if modelXbrl.modelManager.abortOnMajorError and (isEntry or isDiscovered):
+            modelXbrl.error("IOerror",
+                _("%(fileName)s: file error: %(error)s \nLoading terminated."),
+                modelObject=referringElement, fileName=os.path.basename(uri), error=str(err))
+            raise LoadingException()
+        #import traceback
+        #print("traceback {}".format(traceback.format_tb(sys.exc_info()[2])))
         modelXbrl.error("IOerror",
                 _("%(fileName)s: file error: %(error)s"),
                 modelObject=referringElement, fileName=os.path.basename(uri), error=str(err))
-        if modelXbrl.modelManager.abortOnMajorError and (isEntry or isDiscovered):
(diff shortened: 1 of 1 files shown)


aviary2-wf:
## Security Insights

## Action Items
---

arelle/ModelDocument.py
```diff
@@ -1288,7 +1298,10 @@ def contextDiscover(self, modelContext) -> None:
                         if sElt.namespaceURI == XbrlConst.xbrldi and sElt.localName in ("explicitMember","typedMember"):
                             dimQn = sElt.dimensionQname
                             if dimQn: # may be null if schema error omits dimension element
-                                #xmlValidate(self.modelXbrl, sElt)
+                                if targetModelXbrl is not None: # ixds possibly-shared context
+                                    sElt.targetModelXbrl = targetModelXbrl
+                                    if hasattr(sElt, "_dimension") and sElt._dimension is None:
```
brettkail-wk, line 1303: Do I understand correctly that ModelContext/ModelDimensionValue are shared instances across all iXDS, and those objects are mutated as each iXDS is processed?  So something like ixbrl-viewer that wants to aggregate metadata across all iXDS needs to iteratively process a single iXDS, capture metadata, and then move to the next iXDS?  If so, that seems unfortunate.  I wonder if secondary iXDS ModelInlineFact should instead have an unshared context/values (and unit) that can always point to the correct target modelXbrl?

hermfischer-wf: Currently there are different code paths for different situations.  
   1) in EDGAR the documents are assigned to IXDSes in calling Java code and -f structure has independent IXDSes for primary doc and supporting exhibits
   2) in GUI mode or cmd line where docs are intermingled and passed together they are all loaded in one pass and sorted into separate IXDSes which partition the model facts to the right target instance
   3) in either case with multi-target IXDSes they facts are partitioned to the right target instance after the IXDS is loaded
   4) contexts and units which are not shared between facts of multiple targets are allocated to the right target instance after the IXDS is loaded
   5) unreferenced contexts and units are attempted to be shoved into the primary document default target
   6) ixviewer-plus loads the extracted instances to access facts and units, not the original source inline documents

brettkail-wk: Discussed offline, and we're aligned: the current approach is mostly holding together, but we probably need to create some shims (akin to the Prototype objects?) for units+contexts+dims+values of secondary iXDS's.

austinmatherne-wk (approved):

PR #1011, opened 2023-12-21
Fix exceptions with "save Xbrl Formula file" in GUI.

paulwarren-wk:
#### Reason for change

Exception thrown when trying to using `FormulaSaver` plugin.  Use "Tools->Save Xbrl Formula File" on any DTS and you get:

image

This error is in the exception handling code, and the exception that it's trying to report is passing a non-existent `mode` variable as a parameter to `GenerateXbrlFormula`.

#### Description of change

Fixes logging of exception, fixes underlying cause of exception.

This PR also renames the menu option to "Save Formula as XF..." because "Save as Xbrl Formula" is confusing (and badly capitalised).  We try to use the term "XF" to refer to the text-based XBRL Formula format.  Ellipsis also added, because it opens  a file chooser dialog.

#### Steps to Test

Open GUI with any XBRL and the "Formula Saver" plugin enabled. Use Tools->Save Xbrl Formula File and check that a file is produced instead of an exception.

**review**:
@Arelle/arelle

### DIFF
--- a/arelle/plugin/formulaSaver.py
+++ b/arelle/plugin/formulaSaver.py
@@ -470,7 +470,7 @@ class GenerateXbrlFormula:
 
 def saveXfMenuEntender(cntlr, menu, *args, **kwargs):
     # Extend menu with an item for the savedts plugin
-    menu.add_command(label="Save Xbrl Formula file",
+    menu.add_command(label="Save Formula as XF...",
                      underline=0,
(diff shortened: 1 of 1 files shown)


aviary3-wk:
## Security Insights

## Action Items
---

brettkail-wk (approved):

austinmatherne-wk (approved):

PR #1023, opened 2024-01-02
Python 3.12 Support

### LINKED ISSUE #818: Run CI tests against Python 3.12.0 release candidate

[Python 3.12.0rc1](https://www.python.org/downloads/release/python-3120rc1/) has been released. We'll need a supporting [numpy release](https://github.com/numpy/numpy/issues/23808#issuecomment-1670277265) and then we should run our CI tests against 3.12.

austinmatherne-wk:
#### Reason for change
Resolves #818

#### Description of change
Update Arelle to be compatible with Python 3.12 and workflows (besides frozen builds) to use it.
cx_Freeze doesn't support 3.12 yet, so frozen builds will continue to use 3.11 for the time being, but cx_Freeze was updated to the latest release to support newer versions of Matplotlib and NumPy that support Python 3.12.

Updating cx_Freeze broke the code signing step of our macOS build workflow as the library recently overhauled their macOS builds. Fortunately a large part of the overhaul was to make code signing work from within the library, so I was able to delete our code signing step and configure cx_Freeze to handle signing instead.

Python versions prior to 3.12 incorrectly mapped the `C` locale to `en_US` this necessitated overhauling local handling in Arelle. Previously locales and preferred label languages were intermingled (`lang.replace('-', '_')` and `lang.replace('_', '-')` used all over the place) when there's no reason that preferred label languages should be limited to system locales.

#### Steps to Test
* CI
* Frozen builds already tested for [Linux](https://github.com/austinmatherne-wk/Arelle/actions/runs/7394873206), [macOS](https://github.com/austinmatherne-wk/Arelle/actions/runs/7394873849), and [Windows](https://github.com/austinmatherne-wk/Arelle/actions/runs/7394874508).

**review**:
@Arelle/arelle

### DIFF
--- a/arelle/Locale.py
+++ b/arelle/Locale.py
@@ -1,18 +1,25 @@
-'''
+"""
 This module is a local copy of python locale in order to allow
 passing in localconv as an argument to functions without affecting
 system-wide settings.  (The system settings can remain in 'C' locale.)
 
 See COPYRIGHT.md for copyright information.
-'''
+"""
+
 from __future__ import annotations
-import sys, subprocess
-import regex as re
-from typing import Generator, cast, Any, Callable
-from fractions import Fraction
-from arelle.typing import TypeGetText, LocaleDict
-from collections.abc import Mapping
+
+import locale
+import subprocess
+import sys
 import unicodedata
+from collections.abc import Mapping
+from decimal import Decimal
+from fractions import Fraction
+from typing import Any, Callable, Generator, cast
+
+import regex as re
+
+from arelle.typing import LocaleDict, TypeGetText
 
 _: TypeGetText
 
@@ -42,7 +49,6 @@ def _getUserLocaleUnsafe(localeCode: str = '') -> tuple[LocaleDict, str | None]:
     :param localeCode: The locale code to use to retrieve conventions. Defaults to system default.
     :return: Tuple of local conventions dictionary and a user-directed setup message
     """
-    import locale
     conv = None
     localeSetupMessage = None
     localeCode = localeCode.replace('-', '_')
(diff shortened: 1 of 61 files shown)


aviary3-wk:
## Security Insights

## Action Items
---

derekgengenbacher-wf (dismissed):

arelle/Locale.py
```diff
@@ -98,23 +86,145 @@ def getUserLocale(localeCode: str = '') -> tuple[LocaleDict, str | None]:
...
+    Returns a list of language code formats that can be used with gettext to look up translation files.
+    [en_US, en-US, en]
+    [fr]
+    """
     if lang is None:
         lang = getLanguageCode()
-    # allow searching on the lang with country part, either python or standard form, or just language
-    return [lang, lang.replace("-","_"), lang.partition("-")[0]]
+    if LANGUAGE_CODE_LOCALE_SEPARATOR in lang:
+        return [lang, langCodeXBRLFormat(lang), lang.partition(LANGUAGE_CODE_LOCALE_SEPARATOR)[0]]
```
brettkail-wk, line 106: I'm a bit confused about the terminology here.  I think the relevant string formats are "POSIX locale" and BCP 47 "language tag".  Is the latter what you're calling `LANGUAGE_CODE_XML_SEPARATOR` / `langCodeXBRLFormat`?  If so, perhaps `LANGUAGE_TAG_SEPARATOR`, `langTagFormat`, etc. would be clearer names.

The `lang` param sounds like it's language tag, but the code checks for a locale string separator, and the caller is passing a var named `locale`, which seems to come from a user-specified string.  Maybe it's really just a fuzzy config string that might be lang or locale and `configLang` would be clearer, or perhaps just a small doc comment that the string is user-specified, so we're making a best effort.

austinmatherne-wk: I updated a bunch of the variable names and included comments. It should be much more clear now.

derekgengenbacher-wf (approved):

aaroncameron-wk (approved):

aaroncameron-wk:
QA +10
CI passes.

PR #1031, opened 2024-01-09
Use Arelle XML encoding detection instead of lxml encoding

austinmatherne-wk:
#### Reason for change
lxml 5.1.0 [fails the current nl_ntp conformance suites](https://github.com/Arelle/Arelle/actions/runs/7446084583/job/20261413737?pr=1028).

#### Description of change
Use stable ModelDocument encoding instead of lxml.

#### Steps to Test
* CI

**review**:
@Arelle/arelle

### DIFF
--- a/arelle/plugin/validate/NL/rules/fr_nl.py
+++ b/arelle/plugin/validate/NL/rules/fr_nl.py
@@ -227,7 +227,7 @@ def rule_fr_nl_1_05(
     """
     for doc in val.modelXbrl.urlDocs.values():
         if doc.type == ModelDocument.Type.INSTANCE:
-            if 'UTF-8' != doc.xmlDocument.docinfo.encoding:
+            if 'UTF-8' != doc.documentEncoding.upper():
                 yield Validation.error(
                     codes='NL.FR-NL-1.05',
                     msg=_('The XML character encoding \'UTF-8\' MUST be used in the filing instance document'),
--- a/arelle/ModelDocument.py
+++ b/arelle/ModelDocument.py
@@ -683,6 +683,8 @@ class ModelDocument:
         Qualifies as a discovered schema per XBRL 2.1
     """
 
+    # The document encoding. The XML declaration is stripped from the document
+    # before lxml parses the document making the lxml DocInfo encoding unreliable.
     documentEncoding: str
     xmlRootElement: Any
     targetXbrlRootElement: ModelObject
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -1,5 +1,5 @@
-flake8==6.1.0
-flake8-noqa==1.3.2
+flake8==7.0.0
+flake8-noqa==1.4.0
 
 mypy==1.8.0
 pytest==7.4.4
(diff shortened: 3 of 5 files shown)


aviary2-wf:
## Security Insights

## Action Items
---

arelle/plugin/validate/NL/rules/fr_nl.py
```diff
@@ -227,7 +227,7 @@ def rule_fr_nl_1_05(
     """
     for doc in val.modelXbrl.urlDocs.values():
         if doc.type == ModelDocument.Type.INSTANCE:
-            if 'UTF-8' != doc.xmlDocument.docinfo.encoding:
+            if 'UTF-8' != doc.documentEncoding.upper():
```
brettkail-wk, line 230: Out of curiosity, why doesn't `doc.xmlDocument.docinfo.encoding` work, and why is `doc.documentEncoding` the correct thing to use?

(As an aside, I don't know understand what [this logic](https://github.com/Arelle/Arelle/blob/0975615971ef93de11d19b59bc5cf8750bb1b9d1/arelle/XmlUtil.py#L160-L161) is doing.)

austinmatherne-wk: > Out of curiosity, why doesn't `doc.xmlDocument.docinfo.encoding` work, and why is `doc.documentEncoding` the correct thing to use?

Depending on the version of lxml it will detect the encoding as ascii if the file content is limited to ascii characters even if the XML declaration is utf-8.
 
> (As an aside, I don't know understand what [this logic](https://github.com/Arelle/Arelle/blob/0975615971ef93de11d19b59bc5cf8750bb1b9d1/arelle/XmlUtil.py#L160-L161) is doing.)

I believe that's a legacy pattern for declaring encoding in python scripts (not sure why Arelle needs that for loading files).

brettkail-wk: > Depending on the version of lxml it will detect the encoding as ascii if the file content is limited to ascii characters even if the XML declaration is utf-8.

I see.  Is there some lxml bug/release note, or how did we discover that?  A comment.. somewhere.. (maybe in `def encoding`? or on `documentEncoding`?) that explains why we're this should be using rather than`xmlDocument.docinfo.encoding` might be useful.

austinmatherne-wk: I added a comment. Arelle [strips the XML declaration](https://github.com/Arelle/Arelle/blob/0975615971ef93de11d19b59bc5cf8750bb1b9d1/arelle/ModelDocument.py#L157) before parsing the file contents with lxml, so lxml is detecting the encoding without the declaration. I tried changing that, but including the declaration (which lxml then requires the content to be bytes) breaks many many tests and appears to require a significant rewrite.

derekgengenbacher-wf (approved):

arelle/plugin/validate/NL/rules/fr_nl.py
brettkail-wk: Thanks.  That `stripDeclaration` seems to wrongly allow `<?xml encoding='utf-8'?>` (missing version) or even `<?xml invalid='invalid'?>`.  It seems like the code should be updated to always pass bytes to the XML parser.  I agree that looks non-trivial.

derekgengenbacher-wf:

PR #1042, opened 2024-01-16
Save deduplicated traditional XBRL instance

### LINKED ISSUE #1002: Add Option to Deduplicate Facts from Traditional XBRL

We should reuse the implementation in #1001 to offer the ability to deduplicate complete or consistent facts from traditional XBRL instance documents.

```bash
python arelleCmdLine.py --file instance.xml --deduplicateFacts=consistent --saveDeduplicateInstance=report-deduplicated.xml
```

`--deduplicateFacts` should default to `complete` and an exception should be raised if `--deduplicateFacts` is used without `--saveDeduplicateInstance`.

* [ ] complete duplicates.
* [ ] consistent duplicates (the more precise fact should be preserved).

aaroncameron-wk:
#### Reason for change
Resolves #1002 

#### Description of change
Reused parts of duplicate fact validation to transform duplicate fact sets into a deduplicated list based on the provided argument. Implemented `complete`, `consistent-set`, and `consistent-pairs` deduplication modes.
Added duplicate fact user guide to documentation.
Added fact deduplication integration test script.

#### Steps to Test
CI

**review**:
@Arelle/arelle

### DIFF
--- a/arelle/CntlrCmdLine.py
+++ b/arelle/CntlrCmdLine.py
@@ -134,6 +134,13 @@ def parseArgs(args):
                       choices=[a.value for a in ValidateDuplicateFacts.DUPLICATE_TYPE_ARG_MAP],
                       dest="validateDuplicateFacts",
                       help=_("Select which types of duplicates should trigger warnings."))
+    parser.add_option("--deduplicateFacts", "--deduplicatefacts",
+                      choices=[a.value for a in ValidateDuplicateFacts.DeduplicationType],
+                      dest="deduplicateFacts",
+                      help=_("When using '--saveDeduplicatedInstance' to save a deduplicated instance, check for duplicates of this type."))
+    parser.add_option("--saveDeduplicatedInstance", "--savededuplicatedinstance",
+                      dest="saveDeduplicatedInstance",
+                      help=_("Save an instance document with duplicates of the provided type ('--deduplicateFacts') deduplicated."))
     parser.add_option("--noValidateTestcaseSchema", "--novalidatetestcaseschema", action="store_false", dest="validateTestcaseSchema", default=True,
                       help=_("Validate testcases against their schemas."))
     betaGroup = OptionGroup(parser, "Beta Features",
@@ -1083,6 +1090,16 @@ class CntlrCmdLine(Cntlr.Cntlr):
                             ViewFileRoleTypes.viewRoleTypes(modelXbrl, options.roleTypesFile, "Role Types", isArcrole=False, lang=options.labelLang)
                         if options.arcroleTypesFile:
                             ViewFileRoleTypes.viewRoleTypes(modelXbrl, options.arcroleTypesFile, "Arcrole Types", isArcrole=True, lang=options.labelLang)
+                        if options.saveDeduplicatedInstance:
+                            if options.deduplicateFacts:
+                                deduplicateFactsArg = ValidateDuplicateFacts.DeduplicationType(options.deduplicateFacts)
+                            else:
+                                deduplicateFactsArg = ValidateDuplicateFacts.DeduplicationType.COMPLETE
+                            if modelXbrl.modelDocument.type == ModelDocument.Type.INSTANCE:
+                                ValidateDuplicateFacts.saveDeduplicatedInstance(modelXbrl, deduplicateFactsArg, options.saveDeduplicatedInstance)
+                        else:
+                            assert not options.deduplicateFacts, "'deduplicateFacts' can only be used with 'saveDeduplicatedInstance'"
+
                         for pluginXbrlMethod in pluginClassMethods("CntlrCmdLine.Xbrl.Run"):
                             pluginXbrlMethod(self, options, modelXbrl, _entrypoint, responseZipStream=responseZipStream)
 
--- a/arelle/CntlrCmdLine.py
+++ b/arelle/CntlrCmdLine.py
@@ -137,7 +137,8 @@ def parseArgs(args):
     parser.add_option("--deduplicateFacts", "--deduplicatefacts",
                       choices=[a.value for a in ValidateDuplicateFacts.DeduplicationType],
                       dest="deduplicateFacts",
-                      help=_("When using '--saveDeduplicatedInstance' to save a deduplicated instance, check for duplicates of this type."))
+                      help=_("When using '--saveDeduplicatedInstance' to save a deduplicated instance, check for duplicates of this type. "
+                             "Defaults to 'complete'."))
     parser.add_option("--saveDeduplicatedInstance", "--savededuplicatedinstance",
                       dest="saveDeduplicatedInstance",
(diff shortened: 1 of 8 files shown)


aviary2-wf:
## Security Insights

## Action Items
---

arelle/CntlrCmdLine.py
```diff
@@ -1083,6 +1090,16 @@ def monitorParentProcess():
                             ViewFileRoleTypes.viewRoleTypes(modelXbrl, options.roleTypesFile, "Role Types", isArcrole=False, lang=options.labelLang)
                         if options.arcroleTypesFile:
                             ViewFileRoleTypes.viewRoleTypes(modelXbrl, options.arcroleTypesFile, "Arcrole Types", isArcrole=True, lang=options.labelLang)
+                        if options.saveDeduplicatedInstance:
+                            if options.deduplicateFacts:
+                                deduplicateFactsArg = ValidateDuplicateFacts.DeduplicationType(options.deduplicateFacts)
```
austinmatherne-wk, line 1095: Does this need an assertion (and documentation) that this should only be used with traditional XBRL?

What happens if this runs on an inline doc? Does it remove the fact and value from the HTML? I guess a stretch goal could be to remove the fact tagging from iXBRL while preserving the rendered number (replace the XBRL element with its inner text), but I think exiting with an error is good enough until someone asks us to support iXBRL. iXBRL support has limited (no?) value while introducing many complexities (document sets, multiple targets, etc.)

Also, I think this should probably exit after saving the deduplicated instance instead of continuing on (log an error if the user tries to do other actions, validate, etc.) It appears that this puts the model into a strange state (we've modified the lxml tree to remove facts, but the model XBRL still includes them all).

aaroncameron-wk: Opted to just raise an error if this is run on any document that isn't a traditional XBRL instance. I do think iXBRL deduplication would be possible, but it makes sense to wait for demand.

I moved deduplication to run as late as possible since it does put the model into an undefined state. For what it's worth, I do think there's a way to write a new deduplicated XBRL document without manipulating the in-memory ModelXbrl (probably involves walking the lxml tree and writing every element that *isn't* a duplicate fact), but doing it this way allowed reuse of `ModelXbrl.saveInstance`. I'm willing to explore this approach if there's a consensus that it is worthwhile.

It seems that the only case where the model would potentially be used after the model is manipulated is if `keepOpen` is active or plugins that make use of `CntlrCmdLine.Filing.Validate` or `CntlrCmdLine.Filing.End` are active.

austinmatherne-wk: `CntlrCmdLine.Filing.End` and `CntlrCmdLine.Filing.Validate` are pretty common hooks. Can you move this after those hooks and raise an exception if `keepOpen` is used along with `saveDeduplicatedInstance`?

aaroncameron-wk: I've updated this to raise an error after saving if `keepOpen` is used.

The `ModelXbrl` is closed immediately after deduplication occurs currently (unless `keepOpen` is enabled), so it can't be moved any later. Looking into it more, it would appear that if either `CntlrCmdLine.Filing.End` and `CntlrCmdLine.Filing.Validate` are used by a plugin, they won't have access to an open `ModelXbrl` unless they're also using `keepOpen`, so I don't think moving this is necessary.

arelle/CntlrCmdLine.py
```diff
@@ -1123,8 +1123,12 @@ def monitorParentProcess():
                         # Anything depending on the ModelXbrl that runs after this may encounter unexpected behavior,
                         # so we'll run it as a final step in the CLI controller flow.
                         ValidateDuplicateFacts.saveDeduplicatedInstance(modelXbrl, deduplicateFactsArg, options.saveDeduplicatedInstance)
+                        if options.keepOpen:
+                            raise RuntimeError(_("Attempted to keep model connection open after saving deduplicated instance. "
```
austinmatherne-wk, line 1127: I believe both this and the assert below should be `self.addToLog(..., level=logging.CRITICAL)` and set `success` to false.

arelle/ValidateDuplicateFacts.py
```diff
@@ -167,6 +172,105 @@ def areWithinRoundingError(self) -> bool:
...
+            return facts
+        decimalsMap = defaultdict(list)
+        for fact in facts:
+            decimals = self.getDecimals(fact)
+            assert decimals is not None
+            decimalsMap[decimals].append(fact)
+        sortedDecimals = sorted(decimalsMap.keys())
+        results = set(facts)
+
+        for a, decimalsA in enumerate(sortedDecimals[:len(sortedDecimals)]):
```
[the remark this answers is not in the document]
aaroncameron-wk: This implementation (`consistent-pairs` deduplication) delivers the end result of having no consistent duplicates in the output. I think "removing all consistent duplicate facts" is a reasonable interpretation of "consistent deduplication". Even though it's a higher time-complexity implementation, it seems fair to provide this option for anyone expecting this deduplication result.

The other implementation (`consistent-sets` deduplication) may retain consistent duplicates in the output since it takes a hands-off approach when inconsistent duplicates are encountered. We could rename the arg to `consistent` to nudge users in that direction if they're not sure.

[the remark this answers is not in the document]
aaroncameron-wk: The intent was that this loop did not iterate on the final group (the group of highest precision) since facts in that group will never be removed due to having a higher-precision duplicate. I see now that I incorrectly expected the index to be exclusive.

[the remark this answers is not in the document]
brettkail-wk: > This implementation (consistent-pairs deduplication) delivers the end result of having no consistent duplicates in the output

The only difference between `consistent-sets` and `consistent-pairs` is when the report has inconsistent duplicates, which means the report is already of "questionable" validity.  If a user asks to deduplicate, they probably consider inconsistent duplicates to be a fatal error, so I'm trying to understand in what scenario they would care about deduplicating amongst the inconsistent duplicates.  Giving an option that is equivalent to `consistent-sets` when the report is good but runs slower when the report needs more changes seems odd to me.

> We could rename the arg to consistent to nudge users in that direction if they're not sure.

Renaming `consistent-sets` to the shorter `consistent` seems good to me.  I'm not sure about `consistent-pairs` since a1,a2,a3 and b1,b2,b3 with values a,b at three varying precision levels still reduces to a3,b3, so at a high-level it's not about "pairs", it's that it tries hard to remove consistent duplicates even if there are inconsistent duplicates.  Perhaps `complete`, `consistent`, and `inconsistent` are reasonable names for what they do (particularly if `inconsistent` is documented as being potentially slow)?

Anyway, I'll let you all decide.

derekgengenbacher-wf (approved):

arelle/CntlrCmdLine.py
austinmatherne-wk: Thanks. I missed that the models were closed before those hooks executed.

austinmatherne-wk (approved):

PR #1064, opened 2024-01-29
Added HMRC validations for abridged reports and audited small and medium entities.

allenbross-wf:
#### Description of change
Added HMRC validations:
Audited Company Abridged Account
Audited Abridged LLP Accounts
Audited Small Company
Audited Small LLP
Audited Medium Company
Audited Medium LLP

#### Steps to Test

CI.  Conformance suite was updated.

**review**:
@Arelle/arelle

### DIFF
--- a/arelle/plugin/validate/HMRC/ValidateHmrc.py
+++ b/arelle/plugin/validate/HMRC/ValidateHmrc.py
@@ -7,7 +7,8 @@ from dataclasses import dataclass
 from dataclasses import field
 from enum import Enum
 from functools import cached_property
-from typing import cast
+from regex.regex import Pattern
+from typing import Any, Dict, Tuple, cast
 
 import regex as re
 
@@ -16,14 +17,18 @@ from arelle.ModelXbrl import ModelXbrl
 
 # Error codes
 CO_ABRID = 'Co.Abrid'
+CO_AUDIT = 'Co.Audit'
 CO_AUDIT_NR = 'Co.AuditNR'
 CO_DIR_RESP = 'Co.DirResp'
+CO_MED_CO = 'Co.MedCo'
 CO_MICRO = 'Co.Micro'
 CO_MISSING_ELEMENT = 'Co.MissingElement'
 CO_SM_CO = 'Co.SmCo'
 CO_SEC_477 = 'Co.Sec477'
 CO_SEC_480 = 'Co.Sec480'
 LP_ABRID = 'Lp.Abrid'
+LP_AUDIT = 'Lp.Audit'
+LP_MED_LP = 'Lp.MedLp'
 LP_MEM_RESP = 'Lp.MemResp'
 LP_MICRO = 'Lp.Micro'
 LP_SEC_477 = 'Lp.Sec477'
@@ -33,6 +38,7 @@ LP_SM_LP = 'Lp.SmLp'
 # Concept local names
 CONCEPT_ABRIDGED_ACCOUNTS = 'AbridgedAccounts'
 CONCEPT_ABBREVIATED_ACCOUNTS = 'AbbreviatedAccounts'
+CONCEPT_AUDITED = 'Audited'
 CONCEPT_APPLICABLE_LEGISLATION = 'ApplicableLegislation'
 CONCEPT_APPLICABLE_LEGISLATION_DIMENSION = 'ApplicableLegislationDimension'
 CONCEPT_ACCOUNTING_STANDARDS_APPLIED = 'AccountingStandardsApplied'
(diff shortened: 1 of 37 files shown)


aviary-wf:
## Security Insights

## Action Items
---

derekgengenbacher-wf (dismissed):

arelle/plugin/validate/HMRC/ValidateHmrc.py
```diff
@@ -331,13 +347,54 @@ def _lang(self) -> HmrcLang:
                         return HmrcLang.WELSH
         return HmrcLang.ENGLISH
 
+    def _verifyAuditFacts(self, prefix: str) -> None:
```
aaroncameron-wk, line 350: This probably wasn't clear but my intent for `_evaluateCode` was that codes would be passed in to that and the logic within it would determine how the code was evaluated. The benefits are that (a) `_evaluateCode` caches the the result so it won't be evaluated twice should that scenario ever occur, and (b) it separates the logic of the rule evaluation from actually triggering an error/warning.

Take this, for example:
https://github.com/Arelle/Arelle/blob/094f2b0c965bcabb7f27f01f3358e7492dc482e3/arelle/plugin/validate/HMRC/ValidateHmrc.py#L537-L542
This only trigger errors if the evaluation of two different codes has an unsuccessful result. If we were triggering errors during unsuccessful evaluations we would have unwanted errors in this case. It may not come up, but imagine some situation where *either* `_verifyAuditFacts` OR `_verifySomethingElse` passing should result in no errors.

So ultimately my suggestion would be to add a step to `_evaluateCode` that checks these conditions and returns an unsuccessful `CodeResult` if it fails (you can add fields to `CodeResult` to hold whatever information you need to pass into the error). Then the calls below might look like this:
```
result = self._evaluateCode(LP_AUDIT)
if not result.success:
	self._errorOnMissingAuditFacts(LP_AUDIT, result)
```

I'd agree that this is more repetitive in some ways, so I'm open to not being strict about that pattern if others feel that way.

arelle/plugin/validate/HMRC/ValidateHmrc.py
```diff
@@ -331,13 +347,54 @@ def _lang(self) -> HmrcLang:
...
+        """
+        Logs an error when an audited report does not facts tagged with the concepts of "DateAuditorsReport",  "OpinionAuditorsOnEntity"
+        as well as either "NameIndividualAuditor" or the combination of "NameSeniorStatutoryAuditor" and "NameEntityAuditors".
+        :param prefix:
+        :return:
+        """
+        missingConcepts = []
+        dateFacts = self._getFacts(CONCEPT_DATE_AUDITOR_REPORT)
+        if not any(self._checkValidFact(x) for x in dateFacts):
```
aaroncameron-wk, line 359: The `_checkValidFact` helper is a nice addition. I'd even go a step further and add `_checkValidFacts` as will, which retrieves the facts and performs the `any(...)` logic, since it seems to be repeated several times.

arelle/plugin/validate/HMRC/ValidateHmrc.py
```diff
@@ -331,13 +347,54 @@ def _lang(self) -> HmrcLang:
...
+        if len(missingConcepts) > 0:
+            self.modelXbrl.error(
+                prefix + ".Audit",
+                "An audited report must contain facts tagged with the concepts of DateAuditorsReport, OpinionAuditorsOnEntity"
+                "as well as either NameIndividualAuditor or the combination of NameSeniorStatutoryAuditor and NameEntityAuditors"
+                "There are no facts tagged with the concepts: %(concepts)s",
+                concepts=missingConcepts,
+            )
+
+    def _verifyMedFact(self, circumfix: str) -> None:
```
aaroncameron-wk, line 380: In line with my `_verifyAuditFacts` comment above, this logic could be added to `_evaluateCode`. The logic seems pretty generic (evaluating given code should check if given fact exists and is non-nil) so would probably eventually be duplicated for similar errors.

You could establish some kind of `CODE_REQUIRED_CONCEPTS` dict that maps codes to concept local names (i.e. `CODE_REQUIRED_CONCEPTS = { CO_MED_CO: (CONCEPT_MEDIUM_COMPANY,) }`).
Then in the validation you'd have 
```
result = self._evaluateCode(CO_MED_CO)
if not result.success:
  self._errorOnMissingFact(CO_MED_CO, result)
```

derekgengenbacher-wf (approved):

derekgengenbacher-wf (dismissed):

austinmatherne-wk (dismissed):

derekgengenbacher-wf (approved):

arelle/plugin/validate/HMRC/ValidateHmrc.py
```diff
@@ -314,9 +330,96 @@ def _evaluateCode(self, code: str) -> CodeResult:
...
+            result = self._evaluateTextPattern(code, textValidationPatterns)
+        if code == CO_AUDIT:
+            result = self._evaluateAuditFacts(COMPANY)
+        if code == CO_MED_CO:
+            result = self._evaluateMedFact(COMPANY)
+        if code == LP_AUDIT:
+            result = self._evaluateAuditFacts(LLP)
+        if code == LP_MED_LP:
+            result = self._evaluateMedFact(LLP)
+        return result
```
aaroncameron-wk, line 411: Instead of calling `_setCode` in the evaluate sub-functions, I think it's better to have it in only one place at the end of this function. Along with this, the above chain of `if` statements should be `elif` since we would not expect one code to need multiple of these evaluations

```suggestion
        return self._setCode(code, result)
```

arelle/plugin/validate/HMRC/ValidateHmrc.py
```diff
@@ -314,9 +330,96 @@ def _evaluateCode(self, code: str) -> CodeResult:
...
+            missingConcepts.append(CONCEPT_NAME_ENTITY_AUDITORS)
+        if len(missingConcepts) > 0:
+            return self._setCode(
+                code,
+                CodeResult(
+                    success=False,
+                    conceptList=missingConcepts,
+                    message="An audited report must contain facts tagged with the concepts of DateAuditorsReport, OpinionAuditorsOnEntity"
+                            "as well as either NameIndividualAuditor or the combination of NameSeniorStatutoryAuditor and NameEntityAuditors"
+                            "There are no facts tagged with the concepts: %(conceptLocalName)s"
```
aaroncameron-wk, line 364: It doesn't look like `conceptList` gets used. I'm guessing you want it here instead of `conceptLocalName`.

derekgengenbacher-wf (approved):

aaroncameron-wk (approved):

austinmatherne-wk (approved):

derekgengenbacher-wf:

PR #1107, opened 2024-03-04
Update insignificantDigits to handle large numbers

### LINKED ISSUE #1029: Update insignificantDigits to handle large numbers

Hi, Hope all is well! 

We ran into a numeric fact value that was too large for ValidateXbrlCalcs.py (388163667900000000000000000). This caused an exception (please see below) in Arelle. This number was proforma and not actual. I recommend either using a data type that can handle larger numbers or add a test that will alleviate exceptions and provide feedback to the user to easily identify what needs to be investigated and/or fixed. Further details below. Thank you :)

Regarding the file below:

Arelle/Arelle/ValidateXbrlCalcs.py

Line 741, which is below:

insignificantDigits = abs(vDecimal) % divisor

can we make this line of code more versatile please? 

One of our employees tagged the fact below:

    <us-gaap:IncomeLossFromDiscontinuedOperationsNetOfTaxPerBasicShare
      contextRef="C_a32cec8c-3b4f-4875-801f-2f8ff0368484"
      decimals="2"
      id="F_9e4fa528-f075-4f16-9685-e505a0ddfbb6"
      unitRef="U_UnitedStatesOfAmericaDollarsShare">388163667900000000000000000</us-gaap:IncomeLossFromDiscontinuedOperationsNetOfTaxPerBasicShare>

, which results in the following error below:

[exception:InvalidOperation] Instance validation exception: [<class 'decimal.DivisionImpossible'>], instance: mlm-20231231.htm - mlm-20231231.htm 
  File "D:\a\Arelle\Arelle\arelle\Validate.py", line 109, in validate
  File "D:\a\Arelle\Arelle\arelle\ValidateXbrl.py", line 401, in validate
  File "C:\Program Files\Arelle\plugin\validate\EFM\__init__.py", line 349, in validateXbrlFinally
    validateFiling(val, modelXbrl, isEFM=True)
  File "C:\Program Files\Arelle\plugin\validate\EFM\Filing.py", line 858, in validateFiling
    insignificance = insignificantDigits(f1.xValue, decimals=f1.decimals)
                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "D:\a\Arelle\Arelle\arelle\ValidateXbrlCalcs.py", line 741, in insignificantDigits
decimal.InvalidOperation: [<class 'decimal.DivisionImpossible'>]

It took me several hours to find the cause of the error.

aaroncameron-wk:
_Review by commit_

#### Reason for change
Resolves #1029 

#### Description of change
Refactor `insignificantDigits` and implement fix for `InvalidOperation`

#### Steps to Test
CI

**review**:
@Arelle/arelle


aviary-wf:
## Security Insights

## Action Items
---

derekgengenbacher-wf (approved):

arelle/ValidateXbrlCalcs.py
```diff
@@ -694,58 +695,133 @@ def rangeValue(value, decimals=None, truncate=False) -> tuple[decimal.Decimal, d
...
-        if scale:
-            iScale = int(scale)
-            vDecimal = vDecimal.scaleb(iScale)
-        if precision is not None:
-            vFloat = float(value)
-            if scale:
-                vFloat = pow(vFloat, iScale)
-    except (decimal.InvalidOperation, ValueError): # would have been a schema error reported earlier
+        return int(decimals)
+    except ValueError:  # would be a schema error
```
brettkail-wk, line 714: FWIW, I know these comments are pre-existing, but I don't know what they mean.  `decimals` is `decimalsType`, which includes `xs:integer`, which has no bound.  Of course, we can't/don't support arbitrarily large values...

arelle/ValidateXbrlCalcs.py
```diff
@@ -694,58 +695,133 @@ def rangeValue(value, decimals=None, truncate=False) -> tuple[decimal.Decimal, d
...
+    """
+    :param value: Fact value
+    :param precision: Fact precision value
+    :param scale: Fact scale value
+    :return: Tuple of significant and insignificant digit portions of the given value based on precision, if possible. Otherwise None.
+    """
+    if isinf(precision) or precision == 0 or isnan(precision):
+        return None
+    try:
+        valueFloat = float(value)
```
brettkail-wk, line 812: Pre-existing, I don't understand why this logic uses `float` rather than `Decimal`, it seems wrong.  The `pow`/`scaleb` are conceptually redundant, and it seems like it's adding this artificial split to the `insignificantDigits` method (`FromPrecision` and `FromDecimals`), which makes it harder to understand.  It seems like the overall algorithm `insignificantDigits` should be:
```
convert to Decimal
scale # for backwards compat
if precision
  validate precision and convert to decimals
else
  validate decimals
compute truncated/remainder
```
If you do this simpler logic, is there some specific testcase that fails?

arelle/ValidateXbrlCalcs.py
aaroncameron-wk: `ModelFact.decimals` is a `str`, and in most cases it is passed directly into `insignificantDigits`. I think the try/except here is intended to safely handle str values that may not be valid integers at all (which would have already caused a schema error), rather than concern over arbitrarily large values. I'd agree with the suggestion that we should probably be handling that case sooner in the process, but this is where I've drawn the line for the scope of this PR.

derekgengenbacher-wf (approved):

brettkail-wk (dismissed):

allenbross-wf (dismissed):

arelle/ValidateXbrlCalcs.py
```diff
@@ -694,55 +696,39 @@ def rangeValue(value, decimals=None, truncate=False) -> tuple[decimal.Decimal, d
...
-            divisor = ONE.scaleb(-decimals) # fractional scaling doesn't produce scientific notation
-        else:  # extra quantize step to prevent scientific notation for decimal number
-            divisor = ONE.scaleb(-decimals).quantize(ONE, decimal.ROUND_HALF_UP) # should never round
-        insignificantDigits = abs(vDecimal) % divisor
-        if insignificantDigits:
-            return (vDecimal // divisor * divisor,  # truncated portion of number
-                    insignificantDigits)   # nsignificant digits portion of number
+    if insignificant:
+        significant = quotient * divisor
+        return significant, insignificant
```
brettkail-wk, line 731: Do we need to `abs` the divisor for negative values?

aaroncameron-wk: To match previous behavior, yes. The values are only output in error messages as far as I can tell but negative insignificant digit values do seem strange. I've also added negative value test cases.

brettkail-wk (approved):

derekgengenbacher-wf (approved):

derekgengenbacher-wf:

PR #1134, opened 2024-03-21
Run individual conformance suites via `workflow_dispatch`

aaroncameron-wk:
#### Reason for change
Preparation for long-running conformance suites that we don't need to run against every PR

#### Description of change
Move core conformance suite workflow steps to composite action. `conformance-suite` and `conformance-suites` then call the same action.

Also fixed some tangentially related errors I encountered:

1. The new filter-based approach to shards had a flaw where the IDs being compared were URL-encoded and thus getting errantly filtered out. Made a fix and added a check to confirm the number of filter args being passed matches the number of non-skipped results from the shard.
1. The testcase discovery process for shards was only grabbing the first `testcases` element found in testcase index files, meaning testcase index files where testcases are grouped into multiple `testcases` elements were not being ran completely. This may have only recently effected `xbrl_formula_1_0`.

#### Steps to Test
CI. May not be able to run via `workflow_dispatch` until after merge.

**review**:
@Arelle/arelle

### DIFF
--- a/.github/actions/test_conformance_suite/action.yml
+++ b/.github/actions/test_conformance_suite/action.yml
@@ -1,0 +1,58 @@
+name: Test Conformance Suite
+description: Runs pytest against conformance suite runner using given inputs. Adds logs to workflow run.
+inputs:
+  name:
+    description: Name of the conformance suite to run
+    required: true
+  python_version:
+    description: Python version to use
+    required: true
+  shard:
+    description: Optional selection of shards to run (e.g. "0", "0-3")
+    required: false
+  short_name:
+    description: Optional shortened version of name for compact display (defaults to `name`)
+    required: false
+
+runs:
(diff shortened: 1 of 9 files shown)


aviary-wf:
## Security Insights

## Action Items
---

derekgengenbacher-wf (approved):

austinmatherne-wk (approved):

.github/actions/test_conformance_suite/action.yml
```diff
@@ -0,0 +1,58 @@
+name: Test Conformance Suite
```
austinmatherne-wk, line 1: arbitrary line:

Testing for this PR should include making a conformance suite fail and confirming error reporting continues to work properly in our CI.

austinmatherne-wk (approved):

PR #1144, opened 2024-03-27
[FIX] Avoid using id() for the fact aspect cache

RobinetDenisAcsone:
#### Reason for change

The `id()` function used in the fact aspect cache can return the same value for different objects, when the first object is freed.
It can be tested with this code:
```
class Test:
    pass

know_ids = []
new_id = id(Test())
while new_id not in know_ids:
    know_ids.append(new_id)
    new_id = id(Test())

print(know_ids)
```
Since the test object is freed as soon as the `id()` function is finished, the second time that a test object is initialized, it will usually take the same memory address, which will give the same output for `id()` 

It might be linked to the OS, I am on Ubuntu 22.04

#### Description of change

The change removed the usage of `id()` in favor of an uuid stored in the ModelFact and FactPrototype.
There might be a better way to do it, but I didn't find an existing attribute that seems to be usable as a unique identifier.

#### Steps to Test

The random behavior is really visible in the table linkbase conformance pr (https://github.com/Arelle/Arelle/pull/1111) where there is a few conformances that will fail randomly. Running the conformance multiple time will make it vary from 6 to 10 from my tests. (6 tests really fails the conformance suite at the moment)
The reason that it's more visible in that branch is because a lot of FactPrototype are created than freed in the new tableLayout. 

**review**:
@Arelle/arelle

### DIFF
--- a/arelle/ModelInstanceObject.py
+++ b/arelle/ModelInstanceObject.py
@@ -32,6 +32,8 @@
     measure sets.
 """
 from __future__ import annotations
+
+import uuid
 from collections import defaultdict
 from typing import Any, TYPE_CHECKING
 from lxml import etree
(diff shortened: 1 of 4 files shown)


aviary-wf:
## Security Insights

## Action Items
---

austinmatherne-wk (approved):

PR #1167, opened 2024-04-15
DBA: FR41 testcase and validation

aaroncameron-wk:
Based on research with publicly filed documents, I think this rule (and several others) refer specifically to contexts that meet certain "reporting period" criteria. For example, many (assumed valid) filings have qualifying `ProfitLoss` facts with no `TaxExpense` counterparts, but they're associated with contexts that may not be considered an "annual report" or a "reporting period". Because of this, the implementation of this rule only considers the current and previous year reporting period contexts, though that is very much liable to change as our understanding of these rules evolve.

_Note: There appears to be criteria that exempt some filings from this rule. The criteria is based on the legal form of the company, which may or may not be something we can determine from the identification number and/or other instance data. Once we determine if/how that criteria can be implemented, we can add it here._

#### Steps to Test
CI

**review**:
@Arelle/arelle


aviary-wf:
## Security Insights

## Action Items
---

allenbross-wf (approved):

austinmatherne-wk (approved):

PR #1172, opened 2024-04-18
Add check for pattern escapes in schema files

hermfischer-wf:
base spec conf test v401-05 check schema pattern element value for legal escapes

#### Reason for change
not passing v401-05 

#### Description of change
test pattern value escapes

#### Steps to Test
v401-05 

**review**:
@Arelle/arelle

### DIFF
--- a/arelle/ValidateXbrlDTS.py
+++ b/arelle/ValidateXbrlDTS.py
@@ -46,7 +46,8 @@ standard_roles_definitions = {
 standard_roles_other = ("xbrl.5.1.3", ())
 
 inlineDisplayNonePattern = re.compile(r"display\s*:\s*none")
-illegalXsdPatternEscapeChar = re.compile(r"\\[^nrt\\|.^?*+{}()[\]pPsSiIcCdDwW-]")
+# lookbehind below is to ignore even numbers of \ before illegal escape character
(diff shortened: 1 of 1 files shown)


aviary2-wf:
## Security Insights

## Action Items
---

arelle/ValidateXbrlDTS.py
```diff
@@ -46,6 +46,7 @@
 standard_roles_other = ("xbrl.5.1.3", ())
 
 inlineDisplayNonePattern = re.compile(r"display\s*:\s*none")
+illegalXsdPatternEscapeChar = re.compile(r"\\[^nrt\\.^?*+{}()[\]pPsSiIcCdDwW-]")
```
brettkail-wk, line 49: This matches `\|` but shouldn't (i.e., `<pattern value='\|'> matching literal `|` should be allowed).

This matches `\\z` but shouldn't (i.e., `<pattern value='\\z'>` matching the literal `\z` should be allowed).

hermfischer-wf: See https://www.w3.org/TR/2011/CR-xmlschema11-2-20110721/datatypes.diff-wd.html#cces
As I read the spec:
  * \\z  is not allowed
  * \\| is not allowed

brettkail-wk: @hermfischer-wf I agree `\z` is not allowed, but `\\z` should be (but this code incorrectly rejects it).

austinmatherne-wk (approved):

PR #1177, opened 2024-04-22
Python API: Session

### LINKED ISSUE #789: Python Interface Documentation

Issue [#1097](https://github.com/Arelle/Arelle/issues/1097) is simplifying the process of utilizing Arelle through a streamlined Python interface. It would be beneficial to incorporate several practical examples into the Read the Docs project, demonstrating the API's application in performing typical tasks:
1. Validate an ESEF submission and inspect the results.
2. Create an ixbrl-viewer.
3. Load and query an XBRL model.

These examples should be written as tests that are verified by our CI and published as documentation using [`literalinclude`s](https://myst-parser.readthedocs.io/en/latest/syntax/code_and_apis.html#including-code-from-files).

### LINKED ISSUE #1097: Make RuntimeOptions Usable From Python API

In our previous work (see PR [#828](https://github.com/Arelle/Arelle/pull/828)), we introduced the [`RuntimeOptions`](https://github.com/Arelle/Arelle/blob/42dcb76e2d8f9563df13ee8e5a357f2d2cb0b80c/arelle/RuntimeOptions.py) class. This development was aimed at decoupling command-line argument parsing from controller construction. The `RuntimeOptions` class is instantiated and utilized to initialize the [`CntlrCmdLine`](https://github.com/Arelle/Arelle/blob/42dcb76e2d8f9563df13ee8e5a357f2d2cb0b80c/arelle/CntlrCmdLine.py#L571) class directly within the `CntlrCmdLine` module.

The current implementation limits the `RuntimeOptions` class's utility to only the `CntlrCmdLine` class. This limitation prevents its use with the Python API with the underlying base [`Cntlr`](https://github.com/Arelle/Arelle/blob/42dcb76e2d8f9563df13ee8e5a357f2d2cb0b80c/arelle/Cntlr.py#L66) class.

To address this, we should allow both the `Cntlr` and `CntlrCmdLine` constructors to accept the `RuntimeOptions` class as an optional named field. If it's provided, the other parameters must either not be set or match the values in `RuntimeOptions` (if they don't, a ValueError should be raised).

There's [additional handling in CntlrCmdLine.py](https://github.com/Arelle/Arelle/blob/42dcb76e2d8f9563df13ee8e5a357f2d2cb0b80c/arelle/CntlrCmdLine.py#L476C5-L518) for preloading plugins and configuring controller logging. It should be possible to do this for the base `Cntlr` class from a single function that takes `RuntimeOptions`.

With this work a user should be able to:
1. Create and configure an instance of `RuntimeOptions`.
2. Call the function to create and setup a `Cntlr` from the `RuntimeOptions` instance.
3. Call the run method of the `Cntlr` instance to execute validation, generate renderers, or whatever else they configured in the options class.

aaroncameron-wk:
#### Reason for change
Resolves #1097
Resolves #789 

#### Description of change
Allow `RuntimeOptions` to be passed to `CntlrCmdLine`.
Add `Session` Python API entry point.
Add three integration test scripts using Python API and include in documentation.

#### Steps to Test
CI, review docs
Screenshot 2024-04-26 at 2 41 16 PM

**review**:
@Arelle/arelle

### DIFF
--- a/arelle/CntlrCmdLine.py
+++ b/arelle/CntlrCmdLine.py
@@ -477,7 +477,7 @@ def parseArgs(args):
     return runtimeOptions, arellePluginModules
 
 
-def createCntlrAndPreloadPlugins(uiLang, disablePersistentConfig, arellePluginModules):
+def createCntlrAndPreloadPlugins(uiLang, disablePersistentConfig, arellePluginModules) -> CntlrCmdLine:
     """
     This function creates a cntlr and preloads all the necessary plugins.
     :param uiLang: The UI Language
@@ -583,7 +583,7 @@ class CntlrCmdLine(Cntlr.Cntlr):
         super(CntlrCmdLine, self).__init__(hasGui=False, uiLang=uiLang, disable_persistent_config=disable_persistent_config)
         self.preloadedPlugins =  {}
 
-    def run(self, options, sourceZipStream=None, responseZipStream=None) -> bool:
+    def run(self, options: RuntimeOptions, sourceZipStream=None, responseZipStream=None) -> bool:
         """Process command line arguments or web service request, such as to load and validate an XBRL document, or start web server.
 
         When a web server has been requested, this method may be called multiple times, once for each web service (REST) request that requires processing.
--- a/arelle/CntlrCmdLine.py
+++ b/arelle/CntlrCmdLine.py
@@ -91,7 +91,7 @@ def parseArgs(args):
             break
     # Check if the config cache needs to be disabled prior to initializing the cntlr
     disable_persistent_config = bool({DISABLE_PERSISTENT_CONFIG_OPTION, DISABLE_PERSISTENT_CONFIG_OPTION.lower()} & set(args))
-    cntlr = createCntlrAndPreloadPlugins(uiLang, disable_persistent_config, {})  # This Cntlr is needed for translations and to enable the web cache.  The cntlr is not used outside the parse function
+    cntlr = CntlrCmdLine(uiLang=uiLang, disable_persistent_config=disable_persistent_config)  # This Cntlr is needed for translations and to enable the web cache.  The cntlr is not used outside the parse function
     usage = "usage: %prog [options]"
     parser = OptionParser(usage,
                           version="Arelle(r) {0} ({1}bit)".format(Version.__version__, getSystemWordSize()),
--- a/arelle/RuntimeOptions.py
+++ b/arelle/RuntimeOptions.py
@@ -3,16 +3,12 @@ See COPYRIGHT.md for copyright information.
 '''
 from __future__ import annotations
 
-import gettext
 from dataclasses import InitVar, dataclass
 from typing import Any, Optional, Union
 
 from arelle.FileSource import FileNamedStringIO
 from arelle.SystemInfo import hasWebServer
-from arelle.typing import TypeGetText
 from arelle.ValidateDuplicateFacts import DeduplicationType
-
-_: TypeGetText = gettext.gettext
 
 RuntimeOptionValue = Union[bool, int, str, None]
 
@@ -197,7 +193,7 @@ class RuntimeOptions:
                 if hasattr(self, optionName)
             )
             if existingBaseOptions:
-                raise RuntimeOptionsException(_('Provided plugin options already exist as base options {}').format(existingBaseOptions))
+                raise RuntimeOptionsException('Provided plugin options already exist as base options {}'.format(existingBaseOptions))
             for optionName, optionValue in pluginOptions.items():
                 setattr(self, optionName, optionValue)
         if (self.entrypointFile is None and
@@ -205,9 +201,9 @@ class RuntimeOptions:
                 not self.plugins and
                 not pluginOptions and
                 not self.webserver):
-            raise RuntimeOptionsException(_('Incorrect arguments'))
+            raise RuntimeOptionsException('Incorrect arguments')
         if self.webserver and not hasWebServer():
-            raise RuntimeOptionsException(_("Webserver option requires webserver module"))
+            raise RuntimeOptionsException("Webserver option requires webserver module")
         if self.webserver and any((
                 self.entrypointFile, self.importFiles, self.diffFile, self.versReportFile,
                 self.factsFile, self.factListCols, self.factTableFile, self.factTableCols,
@@ -215,5 +211,5 @@ class RuntimeOptions:
                 self.dimFile, self.anchFile, self.formulaeFile, self.viewArcrole, self.viewFile,
                 self.roleTypesFile, self.arcroleTypesFile
         )):
-            raise RuntimeOptionsException(_('Incorrect arguments with webserver'))
+            raise RuntimeOptionsException('Incorrect arguments with webserver')
         self._initialized = True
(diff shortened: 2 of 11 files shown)


aviary2-wf:
## Security Insights

## Action Items
---

arelle/RuntimeOptions.py
```diff
@@ -144,12 +146,35 @@ class RuntimeOptions:
...
+    def __getattr__(self, name: str) -> Any:
+        """
+        If an attribute isn't found, it may be an unspecified plugin option.
+        Return None for any attributes not found on the class.
+        :param name:
+        :return: None
+        """
+        if self._initialized and not self.strictOptions:
+            return None
+        raise AttributeError(name)
```
aaroncameron-wk, line 176: Ideally users of the Python API would not have to specify default values or `None` for every attribute that a plugin might use, and plugins themselves wouldn't have to make use of `hasattr` to determine if an attribute was provided. 

Setting `strictOptions=False` allows attributes like `viewerURL` in `ixbrl-viewer`, for example, to be unset in `RuntimeOptions` while allowing the plugin to reference the attribute and know that it will either contain a value or be `None` and will never raise an `AttributeError`.

tests/integration_tests/scripts/tests/python_api_query_model.py
```diff
@@ -0,0 +1,82 @@
...
+
+assert len(model_xbrls) == 1
+assert len(equity_facts) == 17, len(equity_facts)
+
+print("Cleaning up")
+try:
+    os.unlink(working_directory / 'python_api_query_model' / 'TC2_invalid.zip')
+except PermissionError as exc:
+    print(f"Failed to cleanup test files: {exc}")
+    pass
```
aaroncameron-wk, line 82: Applies to other `python_api_*` test scripts as well.

Windows appears to retain a lock on some files referenced during the Session API run. Something is likely not being closed/disposed properly.

A solution was not immediately clear so I'm choosing to just swallow the exception for now. If this merges as-is, I'll create an issue for it.

arelle/CntlrCmdLine.py
```diff
@@ -579,11 +578,27 @@ class CntlrCmdLine(Cntlr.Cntlr):
     Initialization sets up for platform via Cntlr.Cntlr.
     """
 
-    def __init__(self, logFileName=None, uiLang=None, disable_persistent_config=False):
-        super(CntlrCmdLine, self).__init__(hasGui=False, uiLang=uiLang, disable_persistent_config=disable_persistent_config)
-        self.preloadedPlugins =  {}
+    def __init__(
+        self,
+        options: RuntimeOptions | None = None,
```
austinmatherne-wk, line 585: It would be better to append options as the last parameter so the prior constructor signature is still valid.

arelle/CntlrCmdLine.py
```diff
@@ -592,6 +607,11 @@ def run(self, options, sourceZipStream=None, responseZipStream=None) -> bool:
         :param options: OptionParser options from parse_args of main argv arguments (when called from command line) or corresponding arguments from web service (REST) request.
         :type options: optparse.Values
         """
+        if options is None:
+            options = self.options
+        elif self.options is not None:
+            raise ValueError('Options may only be provided in either the constructor or in call to `run`, not both.')
```
austinmatherne-wk, line 615: Why do we support passing both in the constructor and in run?

arelle/api/Session.py
```diff
@@ -0,0 +1,55 @@
...
+from typing import Any, Iterator
+
+from arelle.CntlrCmdLine import createCntlrAndPreloadPlugins
+from arelle.RuntimeOptions import RuntimeOptions
+from arelle.api.SessionResults import SessionResults
+
+
+class Session:
+    options: RuntimeOptions
+    plugins: dict[str, Any]
```
austinmatherne-wk, line 16: These type hints are already covered by the constructor?

arelle/api/SessionResults.py
```diff
@@ -0,0 +1,35 @@
...
+            return str(handler.getXml(clearLogBuffer=False, includeDeclaration=False))
+        if hasattr(handler, 'getText'):
+            return str(handler.getText(clearLogBuffer=False))
+        return ""
+
+    def get_models(self) -> list[ModelXbrl]:
+        return self._cntlr.modelManager.loadedModelXbrls
+
+    def close(self) -> None:
+        self._cntlr.close()
```
austinmatherne-wk, line 35: SessionResults closing the Cntlr seems problematic. What happens if you call run more than once and have multiple SessionResults objects?

arelle/RuntimeOptions.py
```diff
@@ -190,3 +190,10 @@ def __post_init__(self, pluginOptions: Optional[dict[str, RuntimeOptionValue]])
                 self.roleTypesFile, self.arcroleTypesFile
         )):
             raise RuntimeOptionsException(_('Incorrect arguments with webserver'))
+
+    @staticmethod
+    def assertOptionsEqual(name: str, optionsValue: Any, value: Any) -> Any:
```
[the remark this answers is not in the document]
aaroncameron-wk: Agreed on the function name.

As far as the static implementation, doing something like this:
```
def requireNonConflictingOption(self, name: str, value: Any) -> Any:
        currentValue = getattr(self, name, None)
```
and
```
options.requireNonConflictingOption('logFile', logFileName)
```
means this code will not appear when someone is searching for references to `RuntimeOptions.logFile`. I wanted the implementation to force a direct reference to the property rather than using something like `getattr`, which led me to the implementation that had no use of `self` and thus a static method was appropriate. If there's some kind of trick I'm missing to get around this, or if that isn't a principle we care about, I'm fine with making the above change.

derekgengenbacher-wf (dismissed):

arelle/RuntimeOptions.py
```diff
@@ -11,7 +12,7 @@
 from arelle.typing import TypeGetText
 from arelle.ValidateDuplicateFacts import DeduplicationType
 
-_: TypeGetText
+_: TypeGetText = gettext.gettext
```
austinmatherne-wk, line 15: This overrides the gettext method for this module that's installed in the environment with the arelle domain and translations directory by the CLI and GUI. I'm assuming this is needed because the Python API doesn't run the install code paths for the [CLI](https://github.com/Arelle/Arelle/blob/78d321946eb754bda8651e4679d2002622290d79/arelle/CntlrCmdLine.py#L57) or [GUI](https://github.com/Arelle/Arelle/blob/78d321946eb754bda8651e4679d2002622290d79/arelle/Cntlr.py#L315)?

derekgengenbacher-wf (dismissed):

arelle/RuntimeOptions.py
aaroncameron-wk: Discussed separately. Some planning is required to get the API to a place where translations can be used prior to a `Cntlr` setting up translations *without* a `gettext.install('arelle')` call. For now, I'm proposing simply not attempting translations in `RuntimeOptions` since it's potentially the only place that might actually try to execute a translation prior to the `Cntlr` setting up translations.

derekgengenbacher-wf (approved):

austinmatherne-wk (approved):

tests/integration_tests/scripts/tests/python_api_query_model.py
aaroncameron-wk: https://github.com/Arelle/Arelle/issues/new?assignees=&labels=bug&projects=&template=bug.yml

PR #1203, opened 2024-05-17
[FIX] Fixes table linkbase (EBA)

RobinetDenisAcsone:
* Fix the opening of EBA .xsd files
* Fix the position of the combobox
* Fix the save of table linkbase instance

#### Reason for change
Opening some table linkbase taxonomies (such as EBA) cause a stacktrace. 

#### Description of change
Loading fix:
While testing the "arelle:tableParameterRequiresInstance" error, arelle tries to access the "name" attribute from the elements of the modelParameter.selectProg list. This list is composed of object from [formula/XPathParser.py](https://github.com/Arelle/Arelle/blob/master/arelle/formula/XPathParser.py) like ProgHeader, OperationDef... In the EBA case, there is also QNameDef, which doesn't have the name attribute. 

Combobox position:
Combo boxes were misaligned by one row and one column. It was probably an oversight from the table linkbase update

Save Table linkbase.
The save was never done with tables linkbase, as the filename was never asked.

#### Steps to Test
Open an EBA entrypoint (I tested with corep_LE).
Without this PR, the entrypoint won't load, and if you correct the loading, combobox are misaligned and the save doesn't work.

#### Reason of draft PR
There is still one issue in table linkbase, related to open axes. When there are 2 breakdowns with open axis, only one is handled.
It seems that the issue is located in the RenderingResolution, so it's trickier to fix, I am still looking for the correction.

Old (and correct) behavior:
image

New behavior (missing an open typed dimention):
image

**review**:
@Arelle/arelle

### DIFF
--- a/arelle/CntlrWinMain.py
+++ b/arelle/CntlrWinMain.py
@@ -596,16 +596,6 @@ class CntlrWinMain (Cntlr.Cntlr):
                                 initialdir=initialdir,
                                 filetypes=[(_("HTML file .html"), "*.html"), (_("HTML file .htm"), "*.htm"), (_("XML file .xml"), "*.xml"), (_("JSON file .xml"), "*.json")],
                                 defaultextension=f".{fileType}")
-                    else: # ask file type
-                        if filename is None:
-                            filename = self.uiFileDialog("save",
-                                    title=_caption,
-                                    initialdir=initialdir,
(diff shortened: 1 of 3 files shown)


aviary2-wf:
## Security Insights

## Action Items
---

derekgengenbacher-wf (approved):

Issue #1187, opened 2024-04-30
Make XML Argument Dictionary Logging Truncation Configurable


austinmatherne-wk:
### What should we change and why?

Currently argument values passed as a dictionary to the logger are truncated [to a length of 128 or 4096000](https://github.com/Arelle/Arelle/blob/7b2cde9b5db66f85afb9a114184e66b86184cf2b/arelle/Cntlr.py#L747) if the dictionary key is `json` when writing out xml message logs. We should make this truncation user configurable.

Given:
```python
from arelle.api.Session import Session
from arelle.RuntimeOptions import RuntimeOptions

options = RuntimeOptions(
    entrypointFile='https://www.sec.gov/Archives/edgar/data/1445305/000144530523000166/0001445305-23-000166-xbrl.zip',
    plugins='transforms/SEC',
    logFile='log.xml',
)
with Session() as session:
    session.run(options)
    session._cntlr.addToLog(
        "Document Error",
        level="ERROR",
        messageCode="My.Error",
        messageArgs={
            "myKey": "x" * 129,
        }
    )

```
Will produce:
```xml
<entry code="My.Error" level="error">
 <message myKey="xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx...">[My.Error] Document Error - </message>
 <ref href=""/>
</entry>
```

In this example the long string of `x` chars is truncated at 128.

PR #1224, opened 2024-06-13
Making XML dictionary logging truncation configurable

allenbross-wf:
#### Reason for change

[GitHub issue](https://github.com/Arelle/Arelle/issues/1187)

#### Description of change

Added logXmlMaxLength as a user configurable option

#### Steps to Test

Testing steps outlined in the issue

**review**:
@Arelle/arelle

### DIFF
--- a/arelle/Cntlr.py
+++ b/arelle/Cntlr.py
@@ -338,7 +338,8 @@ class Cntlr:
         logHandler: logging.Handler | None = None,
         logToBuffer: bool = False,
         logTextMaxLength: int | None = None,
-        logRefObjectProperties: bool = True
+        logRefObjectProperties: bool = True,
+        logXmlMaxAttributeLength: int | None = None
     ) -> None:
         # add additional logging levels (for python 2.7, all of these are ints)
         logging.addLevelName(logging.INFO - 1, "INFO-RESULT") # result data, has @name, @value, optional href to source and readable message
@@ -364,7 +365,11 @@ class Cntlr:
                 self.logHandler = StructuredMessageLogHandler()
                 setattr(self.logger, "logRefObjectProperties", logRefObjectProperties)
             elif logFileName.endswith(".xml") or logFileName.endswith(".json") or logToBuffer:
-                self.logHandler = LogToXmlHandler(filename=logFileName, mode=logFileMode or "a")  # should this be "w" mode??
+                self.logHandler = LogToXmlHandler(
+                    filename=logFileName,
+                    mode=logFileMode or "a",
+                    logXmlMaxAttributeLength=logXmlMaxAttributeLength
+                )  # should this be "w" mode??
                 setattr(self.logger, "logRefObjectProperties", logRefObjectProperties)
                 if not logFormat:
                     logFormat = "%(message)s"
--- a/arelle/CntlrCmdLine.py
+++ b/arelle/CntlrCmdLine.py
@@ -256,6 +256,8 @@ def parseArgs(args):
                       help=_("Log reference object properties (default)."), default=True)
     parser.add_option("--logNoRefObjectProperties", "--lognorefobjectproperties", action="store_false", dest="logRefObjectProperties",
                       help=_("Do not log reference object properties."))
+    parser.add_option("--logXmlMaxAttributeLength", "--logxmlmaxattributelength", action="store", dest="logXmlMaxAttributeLength", type="int",
+                      help=_("Truncate XML log file attribute values at length. The default is 4096000 for JSON content and 128 for everything else."))
     parser.add_option("--statusPipe", action="store", dest="statusPipe", help=SUPPRESS_HELP)
     parser.add_option("--monitorParentProcess", action="store", dest="monitorParentProcess", help=SUPPRESS_HELP)
     parser.add_option("--outputAttribution", "--outputattribution", action="store", dest="outputAttribution", help=SUPPRESS_HELP)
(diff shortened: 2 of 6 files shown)


aviary3-wk:
## Security Insights

## Action Items
---

arelle/Cntlr.py
```diff
@@ -364,7 +365,7 @@ def startLogging(
                 self.logHandler = StructuredMessageLogHandler()
                 setattr(self.logger, "logRefObjectProperties", logRefObjectProperties)
             elif logFileName.endswith(".xml") or logFileName.endswith(".json") or logToBuffer:
-                self.logHandler = LogToXmlHandler(filename=logFileName, mode=logFileMode or "a")  # should this be "w" mode??
+                self.logHandler = LogToXmlHandler(filename=logFileName, mode=logFileMode or "a", )  # should this be "w" mode??
```
derekgengenbacher-wf, line 368: What was this change for?

arelle/logging/handlers/LogHandlerWithXml.py
```diff
@@ -39,8 +44,9 @@ def propElts(properties: list[tuple[Any, Any, Any]], indent: str, truncateAt: in
 
         msg = self.format(logRec)
         if logRec.args and isinstance(logRec.args, Mapping):
-            args = "".join([' {0}="{1}"'.format(ncNameEncode(n), entityEncode(v,
-                                                                              truncateAt=(4096000 if n in ("json",) else 128)))
+            args = "".join([' {0}="{1}"'.format(
+                ncNameEncode(n), entityEncode(v, truncateAt=(JSON_MAX_LENGTH if n in ("json",) else (self.logXmlMaxLength or DEFAULT_LOG_XML_MAX_LENGTH)))
```
austinmatherne-wk, line 48: This will raise an AttributeError if `logXmlMaxLength` hasn't been set on self.

arelle/CntlrCmdLine.py
```diff
@@ -256,6 +256,8 @@ def parseArgs(args):
                       help=_("Log reference object properties (default)."), default=True)
     parser.add_option("--logNoRefObjectProperties", "--lognorefobjectproperties", action="store_false", dest="logRefObjectProperties",
                       help=_("Do not log reference object properties."))
+    parser.add_option("--logXmlMaxLength", "--logxmlmaxlength", action="store", dest="logXmlMaxLength", type="int",
+                      help=_("Log file XML field max length override."))
```
austinmatherne-wk, line 260: It's not clear to me what this does based only on the name and help info. Perhaps `--logXmlMaxAttributeLength` and `help=_("Truncate XML log file attribute values at length."))`?  Also set this to the default of `128`?

arelle/Cntlr.py
allenbross-wf: Sorry.  Inadvertent.

arelle/CntlrCmdLine.py
aaroncameron-wk: > Also set this to the default of 128

The pattern we've established thus far with `RuntimeOptions` is that all arguments are optional, and the default behavior for when the value is absent is defined wherever the argument is used. If we define default values in the parser, any direct use of `RuntimeOptions` would not have access to that. We're creating a potential difference between CLI and API usage.

austinmatherne-wk: Perhaps just update the help text to make it clear that the unset value is 128 then?

arelle/logging/handlers/LogHandlerWithXml.py
```diff
@@ -38,9 +43,14 @@ def propElts(properties: list[tuple[Any, Any, Any]], indent: str, truncateAt: in
...
         msg = self.format(logRec)
+        if self.logXmlMaxAttributeLength is not None:
+            xmlTruncation = self.logXmlMaxAttributeLength
+        else:
+            xmlTruncation = DEFAULT_LOG_XML_MAX_LENGTH
         if logRec.args and isinstance(logRec.args, Mapping):
-            args = "".join([' {0}="{1}"'.format(ncNameEncode(n), entityEncode(v,
-                                                                              truncateAt=(4096000 if n in ("json",) else 128)))
+            args = "".join([' {0}="{1}"'.format(
+                ncNameEncode(n), entityEncode(v, truncateAt=(JSON_MAX_LENGTH if n in ("json",) else (xmlTruncation)))
```
austinmatherne-wk, line 52: The parentheses around xmlTruncation are unnecessary and preexisting, but the in tuple check is also unnecessary.
```suggestion
            args = "".join([' {0}="{1}"'.format(
                ncNameEncode(n), entityEncode(v, truncateAt=(self.maxJsonAttributeLength if n == "json" else self.maxXmlAttributeLength))
```

arelle/CntlrCmdLine.py
```diff
@@ -242,6 +242,8 @@ def parseArgs(args):
                       help=_("Write log file mode, a=append if existing (default), w=overwrite if existing. "))
     parser.add_option("--logFormat", "--logformat", action="store", dest="logFormat",
                       help=_("Logging format for messages capture, otherwise default is \"[%(messageCode)s] %(message)s - %(file)s\"."))
+    parser.add_option("--logJsonMaxAttributeLength", "--logjsonmaxattributelength", action="store", dest="logJsonMaxAttributeLength", type="int",
+                      help=_("Truncate JSON log file attribute values at length. The default is 4096000."))
```
austinmatherne-wk, line 246: This isn't quite right. It's still an XML file that's being logged to. The difference is that the content of the message is json. I don't think it's worth cluttering the CLI with two options. If a user needs to increase this they can use the other option.

I think the best course of action is to use the defaults (128/4096000) if the user doesn't provide their own value, but if they do, use it for all messages.

austinmatherne-wk (approved):

PR #1233, opened 2024-06-20
OIM optimization: prevent unnecessary root element replacement

aaroncameron-wk:
#### Reason for change
The approach Arelle [currently uses](https://github.com/Arelle/Arelle/blob/cc8d600ab14a1ce0b8ca17f145ed4290f4c1615e/arelle/XmlUtil.py#L909) to add individual namespaces to a document's `nsmap` is to essentially create a new root element and copy over all child elements from the original root to the new root, and then use that new root as the document going forward. This method is understandably slower the larger the document is.

OIM parsing creates an empty document and progressively writes data to it as it reads source data. It will trigger the above process whenever a new prefix/namespace is encountered, which is quick when the document is small but can take more than a minute once the document has reached considerable size.

#### Description of change
A perfect solution would involve improving the efficiency of `nsmap` updates, but I did not find any potential solutions in my investigation. Instead, an effective solution may involve simply avoiding the need for `nsmap` updates on large documents.

`loadFromOIM` is aware of many prefix/namespace combinations very early on in the parsing process. If we assign these namespaces to the empty initialized document, we won't need to do the costly transferring between roots as described above since that process doesn't run if the prefix/namespace already exists.

#### Steps to Test
I used a 5,000 line CSV recently provided to our support email inbox. Initially, it was taking over 11 minutes to run. With this change it runs in just under a minute, a 92% reduction.

**review**:
@Arelle/arelle

### DIFF
--- a/arelle/oim/Load.py
+++ b/arelle/oim/Load.py
@@ -1759,6 +1759,22 @@ def _loadFromOIM(cntlr, error, warning, modelXbrl, oimFile, mappedUri):
                 initialComment="extracted from OIM {}".format(mappedUri),
                 base=documentBase)
             _return = modelXbrl.modelDocument
+
+        # Updating the namespace map on the document is expensive,
+        # and increases cost as the size of the document grows.
+        # To minimize the effect of this, we can add any known and valid
+        # prefix/namespace combinations to the document preemptively.
+        for prefix, namespace in namespaces.items():
+            if not prefix or not namespace:
+                continue
+            # First, test if the prefix/namespace combination is valid
+            try:
+                etree.Element('nsmap', nsmap={prefix: namespace})
+            except ValueError:
+                # Skip if not valid
+                continue
+            XmlUtil.setXmlns(modelXbrl.modelDocument, prefix, namespace)
+
         if len(modelXbrl.errors) > prevErrLen:
             error("oime:invalidTaxonomy",
                   _("Unable to obtain a valid taxonomy from URLs provided"),


aviary2-wf:
## Security Insights

## Action Items
---

arelle/oim/Load.py
```diff
@@ -1759,6 +1760,22 @@ def dimChecks():
...
             _return = modelXbrl.modelDocument
+
+        # Updating the namespace map on the document is expensive,
+        # and increases cost as the size of the document grows.
+        # To minimize the effect of this, we can add any known and valid
+        # prefix/namespace combinations to the document preemptively.
+        for prefix, namespace in namespaces.items():
+            if not prefix or not namespace:
+                continue
+            # First, test if the prefix/namespace combination is valid
```
brettkail-wk, line 1771: Out of curiosity, what combinations would be invalid?  Are invalid prefixes checked elsewhere, and if so, can we make this code only run on the validated ones?

austinmatherne-wk: I think we should be able to do that [here](https://github.com/Arelle/Arelle/blob/173c273051ab65dbda0794584d2e8ae01fb90f6f/arelle/oim/Load.py#L747-L766). Though I suspect the regex validations aren't as robust as they'd need to be.

austinmatherne-wk (approved):

arelle/oim/Load.py
brettkail-wk: That seems like a gap in the OIM test suite (should be using such a URI) and a gap in Arelle (should be encoding the strings so that they are valid URIs, per [OIM](https://www.xbrl.org/Specification/oim-common/PR-2021-08-04/oim-common-PR-2021-08-04.html#sec-uri-map) -> https://www.w3.org/TR/xmlschema-2/#anyURI -> [xlink](https://www.w3.org/TR/2001/REC-xlink-20010627/#link-locators)) that should be fixed rather than just ignoring the prefixes?

aaroncameron-wk: The testcases in question are around the `oimce:multipleAliasesForURI` error, which it seems is less concerned about the validity of URIs being used and more about whether or not they are duplicate URIs. It introduces the invalid URI above and confirms that we don't consider it a duplicate of "http://xbrl.org/entity/identification/sch%20eme" since uniqueness is determined using the raw values.

This is the testcase description: 
> Two different aliases are bound to URIs that are the same after applying the XLink URI encoding. We use string comparison of raw values when determining uniqueness.

Even then, the prefixes aren't necessarily *ignored*, they're just not included in this optimization. If the prefix/namespace in question is actually used and Arelle attempts to add it to the document later on, then it should likely trigger `oimce:invalidURI`.

allenbross-wf:
Observed similar results following the steps to test.
QA +10

PR #1262, opened 2024-07-03
Edgar validation update 1

hermfischer-wf:
* fix arelle edgar cmd line operation with flat zip of multi-doc single-IXDS with no json file argument, was raising exception, found by Ram

#### Reason for change
 * exception with flat zip of multi-doc single-IXDS without json file argument

#### Description of change
 * assume no attachmentDocumentType

#### Steps to Test
 * try with -f of multi-doc zip

**review**:
@Arelle/arelle

### DIFF
--- a/arelle/plugin/validate/EFM/Filing.py
+++ b/arelle/plugin/validate/EFM/Filing.py
@@ -159,7 +159,7 @@ def validateFiling(val, modelXbrl, isEFM=False, isGFM=False):
     # efmSubmissionType and efmIxdsType are already set when re-validating after redaction/redline removal
     submissionType = getattr(modelXbrl,'efmSubmissionType', val.params.get("submissionType", ""))
     attachmentDocumentType = getattr(modelXbrl,'efmIxdsType', val.params.get("attachmentDocumentType", "")) # this is different from dei:documentType
-    isFeeTagging = feeTaggingAttachmentDocumentTypePattern.match(attachmentDocumentType)
+    isFeeTagging = feeTaggingAttachmentDocumentTypePattern.match(attachmentDocumentType or "")
     requiredFactLang = disclosureSystem.defaultXmlLang.lower() if disclosureSystem.defaultXmlLang else disclosureSystem.defaultXmlLang
     hasSubmissionType = bool(submissionType)
     hasAttachmentDocumentType = bool(attachmentDocumentType)


aviary3-wk:
## Security Insights

## Action Items
---

austinmatherne-wk (approved):

arelle/plugin/validate/EFM/Filing.py
```diff
@@ -159,7 +159,7 @@ def validateFiling(val, modelXbrl, isEFM=False, isGFM=False):
     # efmSubmissionType and efmIxdsType are already set when re-validating after redaction/redline removal
     submissionType = getattr(modelXbrl,'efmSubmissionType', val.params.get("submissionType", ""))
     attachmentDocumentType = getattr(modelXbrl,'efmIxdsType', val.params.get("attachmentDocumentType", "")) # this is different from dei:documentType
```
austinmatherne-wk, line 161: It seems like the intent is that attachmentDocumentType is always a string and should be ensured at assignment?
```suggestion
    attachmentDocumentType = getattr(modelXbrl,'efmIxdsType', val.params.get("attachmentDocumentType")) or "" # this is different from dei:documentType
```

hermfischer-wf: Not sure I'm comfortable always making it a string, but for this one case on the pattern match it can't be None.  This will probably evolve a bit because attachmentDocumentType might be inferable from the instance in later processing in Filing.py, so would prefer it remain None when not specified.

austinmatherne-wk: In that case, we can skip the regex match if we know it won't match.

```python
isFeeTagging = attachmentDocumentType and feeTaggingAttachmentDocumentTypePattern.match(attachmentDocumentType)
```

michaelchadfleming: @austinmatherne-wk @hermfischer-wf Hi Guys, Hope all is well. Is it possible to get this fix merged to master today please? I spoke with Herm on the 3rd and he said that the change Austin recommended was fine for now. I see there was a merged commit yesterday but this file was not updated. Thank you as always for your valuable time :)

carolusalarcon: Hi all, 
Out team is also waiting for this fix. This is solving our current issue.

Thanks

austinmatherne-wk: This was merged and released yesterday ([2.30.5](https://github.com/Arelle/Arelle/releases/tag/2.30.5)).

carolusalarcon: I'm missing something?

this line still the same
https://github.com/Arelle/Arelle/blob/18922e707834b6f30d7ca2e379d7c325fcbcd8db/arelle/plugin/validate/EFM/Filing.py#L161

austinmatherne-wk: My apologies, I thought you were referring to Herm's change in this PR, not my suggestion. Herm is testing the suggestion now.

PR #1280, opened 2024-07-11
DBA: FR22 and FR23 validation rules and related test cases

allenbross-wf:
The control only looks at instances without dimensions or instances that only have the dimension (ConsolidatedSoloDimension with ConsolidatedMember). Instances with other dimensions do not lift the reporting obligation.

> DBA.FR23: Liabilities (fsa:LiabilitiesAndEquity) must be entered and must not be negative.
The control only looks at instances without dimensions or instances that only have the dimension (ConsolidatedSoloDimension with ConsolidatedMember). Instances with other dimensions do not lift the reporting obligation.

#### Description of change

Added FR22 and FR23 as well as related test cases.

#### Steps to Test

CI

**review**:
@Arelle/arelle


aviary3-wk:
## Security Insights

## Action Items
---

arelle/plugin/validate/DBA/rules/__init__.py
```diff
@@ -73,12 +73,38 @@ def errorOnRequiredFact(
...
+    """
+    Yields an error if a fact with the given QName is not tagged with a valid date and a non-nil value.
+    :return: Yields validation errors.
+    """
+    errorModelObjects = []
+    if not facts:
+        errorModelObjects.append(modelXbrl.modelDocument)
+    else:
+        for fact in facts:
+            if fact.xValue < 0:
```
derekgengenbacher-wf, line 91: I think that this will throw a Type error if a fact is tagged without a value such that xValue is None: `TypeError: '<' not supported between instances of 'NoneType' and 'int'`. So I think you also need to add an xValid check. So something like the following:

`if fact.xValid >= VALID and fact.xValue < 0: `

allenbross-wf: I added the check to the file but I don't think that adding a conformance suite test is of any benefit as I can't envision a test that would be testing the validation and not core Arelle.

aaroncameron-wk (dismissed):

derekgengenbacher-wf (approved):

aaroncameron-wk (approved):

derekgengenbacher-wf:

PR #1296, opened 2024-07-30
Added ROS.18 Validation and conformance suite.

allenbross-wf:
Reason for change
>Rule 18: DPLTurnoverRevenue cannot be a negative value.

Description of change

Added ROS.18
Added conformance suite.

Steps to Test
CI

### DIFF
--- a/arelle/plugin/validate/ROS/rules/__init__.py
+++ b/arelle/plugin/validate/ROS/rules/__init__.py
@@ -3,21 +3,22 @@ See COPYRIGHT.md for copyright information.
 """
 from __future__ import annotations
 
-from typing import Iterable, Callable, cast, Union
+from typing import Iterable, cast
 
 from arelle.ModelInstanceObject import ModelFact
 from arelle.ModelXbrl import ModelXbrl
+from arelle.XmlValidateConst import VALID
 from arelle.utils.validate.Validation import Validation
 
 
-def errorOnRequiredFact(
+def errorOnMissingRequiredFact(
         modelXbrl: ModelXbrl,
         conceptLn: str,
         code: str,
         message: str,
 ) -> Iterable[Validation]:
     """
-    Yields an error if a fact with the given QName is not tagged with a non-nil value.
+    Yields an error if a fact with the given localName is not tagged with a non-nil value.
     :return: Yields validation errors.
     """
     facts: set[ModelFact] = modelXbrl.factsByLocalName.get(conceptLn, set())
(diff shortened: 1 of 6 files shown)


aviary3-wk:
## Security Insights

## Action Items
---

arelle/plugin/validate/ROS/rules/__init__.py
```diff
@@ -17,16 +18,41 @@ def errorOnRequiredFact(
...
+        modelXbrl: ModelXbrl,
+        conceptLn: str,
+        code: str,
+        message: str,
+) -> Iterable[Validation]:
+    """
+    Yields an error if a fact with the given localName is tagged with a negative value.
+    :return: Yields validation errors.
+    """
+    errorModelFacts: list[ModelFact | None] = []
```
derekgengenbacher-wf, line 47: I dont believe that you need a None here for typing. The list will either be a list of of model facts or an empty list of model facts.

arelle/plugin/validate/ROS/rules/__init__.py
```diff
@@ -17,16 +18,41 @@ def errorOnRequiredFact(
...
 ) -> Iterable[Validation]:
     """
-    Yields an error if a fact with the given QName is not tagged with a non-nil value.
+    Yields an error if a fact with the given localName is not tagged with a non-nil value.
     :return: Yields validation errors.
     """
     facts: set[ModelFact] = modelXbrl.factsByLocalName.get(conceptLn, set())
     for fact in facts:
         if not fact.isNil:
             return
```
derekgengenbacher-wf, line 27: If a list of facts gets run through this we do not indicate which facts are actually in error. It seems like this should be appending fats to a list and then returning that list as the model objects of the error?

aaroncameron-wk: I think the function name may be misleading. The intent is to error on a _missing_ required fact. There are no facts to include in the error, just the concept for which a fact is missing. Suggest `errorOnMissingRequiredFact`?

arelle/plugin/validate/ROS/rules/__init__.py
```diff
@@ -17,16 +18,41 @@ def errorOnRequiredFact(
...
+        conceptLn=conceptLn,
         codes=code,
         msg=message,
-        conceptLn=conceptLn,
+
 
     )
+
+
+def errorOnPositiveOnlyFact(
```
aaroncameron-wk, line 37: I think the name here is the opposite of the implementation. It errors on a negative fact. Suggest `errorOnNegativeFact`.

arelle/plugin/validate/ROS/rules/__init__.py
allenbross-wf: I don't think that's necessarily wrong to have a fact with a nil-value , the rule seems to indicate that one fact needs to be non-nil.  Not all of them.

derekgengenbacher-wf (approved):

aaroncameron-wk (approved):

PR #1299, opened 2024-08-01
Allow self closed nil ix elements for EDGAR validation

hermfischer-wf:
#### Reason for change
EdgarRenderer rewritten inline documents (for addition of fact ID and removal of redline/redaction) rewrites nil ix elements as self closed (for compatibility with other validation tools), but round tripping gets caught with an input validation disallowing self-closed (nil) elements, which isn't in the specification.

#### Description of change
Allow self closed nil ix elements for reciprocity/compatibility with EdgarRenderer method uncloseSelfClosedTags

#### Steps to Test
Any nonNumeric or nonFraction with xsi:nil and self closed should be accepted.

**review**:
@Arelle/arelle

### DIFF
--- a/arelle/ValidateFilingText.py
+++ b/arelle/ValidateFilingText.py
@@ -30,7 +30,8 @@ namedEntityPattern = re.compile("&[_A-Za-z\xC0-\xD6\xD8-\xF6\xF8-\xFF\u0100-\u02
 #                                "\xB7A-Za-z0-9\xC0-\xD6\xD8-\xF6\xF8-\xFF\u0100-\u02FF\u0370-\u037D\u037F-\u1FFF\u200C-\u200D\u2070-\u218F\u2C00-\u2FEF\u3001-\uD7FF\uF900-\uFDCF\uFDF0-\uFFFD\u0300-\u036F\u203F-\u2040]*;")
 
 inlinePattern = re.compile(r"xmlns:[\w.-]+=['\"]http://www.xbrl.org/2013/inlineXBRL['\"]")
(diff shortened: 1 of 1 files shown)


aviary2-wf:
## Security Insights

## Action Items
---

arelle/ValidateFilingText.py
```diff
@@ -497,7 +501,8 @@ def close(self): pass
             if isInline and '/>' in line:
                 for match in inlineSelfClosedElementPattern.finditer(line):
                     selfClosedLocalName = match.group('localName')
-                    if selfClosedLocalName not in elementsWithNoContent:
+                    if (selfClosedLocalName not in elementsWithNoContent and
+                        not (selfClosedLocalName in ixElementsAllowingNil and xsiNilPattern.match(match.group('attributes') or ''))):
```
brettkail-wk, line 505: I don't understand.  The point of this warning is that browsers in non-XHTML mode will interpret this:
```
<ix:nonNumeric/>
<div>content</div>
```
...as this:
```
<ix:nonNumeric><div>content</div></ix:nonNumeric>
```
It doesn't matter whether the element has `xsi:nil` or not, it's wrong to output self-closing.

For the specific case we discussed, the SEC has logic that rewrites the iXBRL, and that logic is wrongly introducing self-closing elements.  I think that logic needs to be fixed, not the warning.

hamscher: Completely agree with brettkail-wk.
The reserializer in arelle/EDGAR/render needs to be fixed, not this.

PR #1300, opened 2024-08-01
Add ModelXbrl ixdsTarget property/setter that handles (default) value

aaroncameron-wk:
#### Reason for change
Arelle inconsistently handles the "(default)" IXDS target value.
For example:
In the IXDS plugin, the "(default)" value [translates](https://github.com/Arelle/Arelle/blob/948d3ea67940b3bd9b0d12714e50f9d8685e0e11/arelle/plugin/inlineXbrlDocumentSet.py#L155) to `None`.
But where the `arelle:ixdsTargetNotDefined` error code [fires](https://github.com/Arelle/Arelle/blob/948d3ea67940b3bd9b0d12714e50f9d8685e0e11/arelle/ModelDocument.py#L1995), there is no consideration for the "(default)" value.

#### Description of change
Ensure that whenever `modelXbrl.ixdsTarget` is retrieved, it returns `None` if the underlying value is "(default)".

#### Steps to Test
CI

**review**:
@Arelle/arelle

### DIFF
--- a/arelle/ModelXbrl.py
+++ b/arelle/ModelXbrl.py
@@ -309,6 +309,7 @@ class ModelXbrl:
     _startedProfiledActivity: float
     _startedTimeStat: float
     _qnameUtrUnits: dict[QName, UtrEntry]
+    _ixdsTarget: str | None
 
     def __init__(self,  modelManager: ModelManager, errorCaptureLevel: str | None = None) -> None:
         self.modelManager = modelManager
@@ -1000,6 +1001,16 @@ class ModelXbrl:
             self.modelManager.addToLog(_("Exception viewing properties {0} {1} at {2}").format(
                             modelObject,
                             err, traceback.format_tb(sys.exc_info()[2])))
+
+    @property
+    def ixdsTarget(self) -> str | None:
+        if self._ixdsTarget == '(default)':
+            return None
+        return self._ixdsTarget
+
+    @ixdsTarget.setter
+    def ixdsTarget(self, value: str | None) -> None:
+        self._ixdsTarget = value
 
     def effectiveMessageCode(self, messageCodes: tuple[Any] | str) -> str | None:
         """


aviary-wf:
## Security Insights

## Action Items
---

arelle/ModelXbrl.py
```diff
@@ -1001,6 +1002,16 @@ def viewModelObject(self, objectId: str | int) -> None:
...
                             err, traceback.format_tb(sys.exc_info()[2])))
 
+    @property
+    def ixdsTarget(self) -> str | None:
+        if self._ixdsTarget == '(default)':
+            return None
+        return self._ixdsTarget
+
+    @ixdsTarget.setter
+    def ixdsTarget(self, value: str | None) -> None:
```
brettkail-wk, line 1012: Is it safe to change the getter, or are there callers that will rely on reading the `(default)`?

If it's safe to change the getter, why not change the caller to set None instead?

If it's not safe to change the setter (perhaps there are many or they're unknown), why not replace `(default)` on set rather than on get?

If we need to normalize in this class, can we add a constant and import it in inlineXbrlDocumentSet.py?  If not, can we add a comment explaining where it comes from?

austinmatherne-wk: It's safe to change the getter/setter. All Arelle code that uses the getter expects `None`, not `(default)` (the setting code in the ixds plugin currently transforms it to `None`. The issue is code using Arelle as a library that bypasses the ixds plugin and manually sets the target.

I think it would be better to perform the transform in the setter and also throw a better exception message if the getter is accessed before the target is set.

brettkail-wk: Out of curiosity, why is it better to change the `@ixdsTarget.setter` rather than removing `DEFAULT_TARGET` altogether?

austinmatherne-wk: `(default)` comes from the XBRL spec and is what's used in the CLI to explicitly specify the default target, so the translation needs to happen somewhere unless we want to make a breaking change. Performing it within the setter ensures that all interfaces (cli, gui, python library) behave the same.

brettkail-wk: > `(default)` comes from the XBRL spec

Where does it say that?  The iXBRL spec says:
> The Target Document identified by the actual value of the target attribute or, if absent, the default Target Document

> what's used in the CLI to explicitly specify the default target, so the translation needs to happen somewhere unless we want to make a breaking change. Performing it within the setter ensures that all interfaces (cli, gui, python library) behave the same.

Agreed.  `(default)` seems like a CLI-specific string, so it seems like that should be converted in the CLI code rather than core Arelle.

austinmatherne-wk: > Where does it say that? The iXBRL spec says:

It doesn't. My memory failed me. I was conflating Arelle's behavior with the spec 🤦‍♂️ 

Agreed about converting in the CLI code and expecting other interfaces (GUI and python lib) to not use it.

PR #1303, opened 2024-08-03
Added DBA.FR21 and related conformance test.

allenbross-wf:
#### Reason for change

> DBA.FR21: Assets (BalanceSheet:Assets) must equal liabilities (BalanceSheet:LiabilitiesAndEquity). The check only looks at instances without dimensions and instances that only have the dimension group (ConsolidatedMember) in context. Occurrences of assets and liabilities with other dimensions do not lift the reporting obligation. The comparison takes place within pairs of numbers for currency and group/parent, respectively.

#### Description of change

Added validation DBA.FR21 and related conformance suite test case.

#### Steps to Test

CI

**review**:
@Arelle/arelle


aviary-wf:
## Security Insights

## Action Items
---

tests/resources/conformance_suites/dba/fr/fr22-testcase.xml
```diff
@@ -18,6 +18,7 @@
             <instance readMeFirst="true">fr22-invalid-negative.xbrl</instance>
         </data>
         <result>
+            <error>DBA.FR21</error>
```
aaroncameron-wk, line 21: Is there a challenge in getting these docs not to fire DBA.FR21? I'm fine with this but just wanting to confirm. If it's as easy as changing a couple of fact values or adding a fact then I think it's worth doing for this to be a clean testcase, otherwise it's fine to keep as-is.

arelle/plugin/validate/DBA/rules/fr.py
```diff
@@ -177,6 +179,51 @@ def rule_fr10(
...
+    for contextID, facts in groupedFacts.items():
+        assetFact = None
+        liabilityFact = None
+        for fact in facts:
+            if fact.qname == pluginData.assetsQn:
+                assetFact = fact
+            elif fact.qname == pluginData.liabilitiesQn:
+                liabilityFact = fact
+        if assetFact is not None and liabilityFact is not None:
+            if assetFact.xValue != liabilityFact.xValue:
```
aaroncameron-wk, line 219: An alternative would be to collect each `fact.xValue` from `facts` into a set and check if the length is 1. This would have the benefit of handling a situation where there are potentially two liabilities facts with different values, but if we know that situation would fire an error elsewhere then we don't need to handle it here.

aaroncameron-wk (approved):

tests/resources/conformance_suites/dba/fr/fr22-testcase.xml
allenbross-wf: There is.  Because the fr22-invalid-negative and fr23-invalid-negative test cases have negative values for Assets and Liabilities   Resultantly when the comparison is made between the two values they aren't equal.  Making them both negative. would fix the fr21 error but then both the fr22-invalid-negative and fr23-invalid-negative would fire fr22 and fr23 errors.

arelle/plugin/validate/DBA/rules/fr.py
allenbross-wf: Two different values for the facts would have different contextIDs or they would be duplicate facts, wouldn't they?

aaroncameron-wk: I think so, but I don't think duplicate facts raise errors unless that feature is specifically invoked. See `validateDuplicateFacts`.

derekgengenbacher-wf:
This is duplicated by a formula

PR #1310, opened 2024-08-07
Add Option For Expected Test Case Errors

austinmatherne-wk:
#### Reason for change
The current approach of expected failing conformance suite tests doesn't tell us if we further break a test by introducing additional error codes.

#### Description of change
* Log the original URI and not the web cache path when a URI cannot be mapped to a file in the cache.
* Add option to explicitly list errors that cause a test case to fail.
* Don't attempt to open a file that has previously failed to open (limit duplicate error messages).

#### Steps to Test
* CI

**review**:
@Arelle/arelle


aviary3-wk:
## Security Insights

## Action Items
---

aaroncameron-wk (approved):

PR #1335, opened 2024-08-28
ROS.19 fires if principle currency is not majority of monetary facts

derekgengenbacher-wf:
#### Reason for change
ROS.19 was firing if even one fact did not match the principle currency.

#### Description of change
ROS.19 was updated to only fire if principle currency is not the majority of monetary facts.

#### Steps to Test
CI

**review**:
@Arelle/arelle

### DIFF
--- a/arelle/plugin/validate/ROS/rules/ros.py
+++ b/arelle/plugin/validate/ROS/rules/ros.py
@@ -6,6 +6,8 @@ from __future__ import annotations
 import os
 from typing import Any, Iterable
 
+from collections import Counter
+
 from arelle.typing import TypeGetText
 from arelle.ValidateXbrl import ValidateXbrl
 from collections import defaultdict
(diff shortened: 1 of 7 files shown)


aviary3-wk:
## Security Insights

## Action Items
---

arelle/plugin/validate/ROS/rules/ros.py
```diff
@@ -347,30 +349,35 @@ def rule_ros19(
...
+        principal_currency_value = principal_currency_values.pop()
+        monetary_facts = list(val.modelXbrl.factsByDatatype(False, qnXbrliMonetaryItemType))
+        monetary_units = [list(fact.utrEntries)[0].unitName for fact in monetary_facts if fact.unit is not None]
+        unit_counts = Counter(monetary_units)
+        principal_currency_value_count = unit_counts[principal_currency_value]
+        for unit, count in unit_counts.items():
+            if count > principal_currency_value_count:
+                modelObjects = pricipal_currency_facts
+                message = _("'%(concept_name)s' must exist and its value must match the functional unit of the financial statement. "
+                            "Units used for monetary facts: %(monetary_units)s, '%(concept_name)s' currency values: %(principal_currency_values)s.")
```
brettkail-wk, line 373: I'm not sure how to reconcile this change with the docstring or error message.  Is the thinking that the authority also only requires "majority", or that they don't check it at all, or that it's really a warning?

derekgengenbacher-wf: The fact tagged with `PrincipalCurrencyUsedInBusinessReport` denotes what unit a majority of the monetary facts must use. If there is a tie for the number of facts and their usage of different units then we do not fire the error. If `PrincipalCurrencyUsedInBusinessReport` is not tagged or is tagged multiple times then we fire an error.

brettkail-wk: I see that's what the code does, but I don't understand why the change is OK per the PDF.  It looks like the PDF only requires the `PrincipalCurrencyUsedInBusinessReport` fact to exist and that it "match the functional currency of the financial statements".  Checking the latter seems like a heuristic, which seems like a warning or info rather than an error?

aaroncameron-wk (approved):

arelle/plugin/validate/ROS/rules/ros.py
derekgengenbacher-wf: This is the information that I see:

> "The PrincipalCurrencyUsedInBusinessReport tag is required in all iXBRL submissions and the currency declared MUST match the functional currency of the financial statements. If this error is returned in the course of submitting iXBRL documents, filers should either
> (i)        Check to make sure that the tag is present and that the currency declared for it matches the functional currency of the financial statements
> OR
> (ii)        Take note of the error and contact your software provider to request that the file is amended to satisfy this rule."

The `MUST` part makes it an error.

brettkail-wk: I understand it's an error according to the manual, but if Arelle can't know for sure whether a fact must have the principal currency or not, it seems wrong to give an error.  Alternatively, if a user wants to use a consistent currency for all facts but mistakenly uses the wrong currency for only one, it seems wrong that Arelle will no longer report an issue for that.

brettkail-wk: Discussed internally.  It's unclear whether the rule is actually enforceable in an automated way (only the user knows whether the narrative is using the primary currency), but our analysts believe that because the secondary currency(ies) are derivative of the values in the primary currency, that it should always be the case that there are at least as many values in the primary currency as the secondary currency.

PR #1348, opened 2024-09-06
Fix KeyError in ROS.19

derekgengenbacher-wf:
#### Reason for change
ROS.19 can fail with a KeyError

#### Description of change
Updated usages of `principal_currency_values`

#### Steps to Test
CI

**review**:
@Arelle/arelle

### DIFF
--- a/arelle/plugin/validate/ROS/rules/ros.py
+++ b/arelle/plugin/validate/ROS/rules/ros.py
@@ -351,14 +351,15 @@ def rule_ros19(
     ROS: Rule 19:PrincipalCurrencyUsedInBusinessReport must exist and its value must match the unit
     used for the majority of monetary facts.
     """
-    message: str | None = None
-    principal_currency_value: str | None = None
-    modelObjects: set | None = None
     principal_currency_facts = val.modelXbrl.factsByLocalName.get(PRINCIPAL_CURRENCY, set())
     principal_currency_values = set(fact.text for fact in principal_currency_facts)
     if len(principal_currency_values) != 1:
-        modelObjects = principal_currency_facts
-        message = _("'PrincipalCurrencyUsedInBusinessReport' must exist and have a single value.  Values found: %(principal_currency_values)s.")
+        yield Validation.error(
+            "ROS.19",
+            _("'PrincipalCurrencyUsedInBusinessReport' must exist and have a single value.  Values found: %(principal_currency_values)s."),
+            modelObject=principal_currency_facts,
+            principal_currency_values=principal_currency_values,
+        )
     else:
         principal_currency_value = principal_currency_values.pop()
         monetary_facts = list(val.modelXbrl.factsByDatatype(False, qnXbrliMonetaryItemType))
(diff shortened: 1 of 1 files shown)


aviary-wf:
## Security Insights

## Action Items
---

arelle/plugin/validate/ROS/rules/ros.py
```diff
@@ -368,13 +368,13 @@ def rule_ros19(
         for unit, count in unit_counts.items():
             if count > principal_currency_value_count:
                 modelObjects = principal_currency_facts
-                message = _("'PrincipalCurrencyUsedInBusinessReport' has a value of %(principal_currency_value)s, "
+                message = _("'PrincipalCurrencyUsedInBusinessReport' has a value of %(principal_currency_values)s, "
```
brettkail-wk, line 371: This output seems worse ("has a value of {USD}" vs "has a value of USD").  Why this change?

arelle/plugin/validate/ROS/rules/ros.py
```diff
@@ -368,13 +368,13 @@ def rule_ros19(
...
                             "which must match the functional(majority) unit of the financial statement.")
                 continue
     if message:
         yield Validation.warning(
             "ROS.19",
             message,
             modelObject=modelObjects,
-            principal_currency_value=principal_currency_value
+            principal_currency_values=principal_currency_values if principal_currency_values else None
```
brettkail-wk, line 379: Is the intent to turn empty set into `None`?  If yes, out of curiosity, why?  If no, why not simply `principal_currency_values=principal_currency_values`?

austinmatherne-wk (dismissed):

arelle/plugin/validate/ROS/rules/ros.py
```diff
@@ -376,5 +376,6 @@ def rule_ros19(
             "ROS.19",
             message,
             modelObject=modelObjects,
+            principal_currency_values=principal_currency_values,
             principal_currency_value=principal_currency_value
```
brettkail-wk, line 380: I think the `principal_currency_value` line still needs to be removed (this was presumably the cause of the error since thatvar is only defined if there's exactly one).

derekgengenbacher-wf: `principal_currency_value` will show up as a None if it is not set.

austinmatherne-wk: It'll throw a 'principal_currency_value' is not defined exception if the conditional check on line 359 is True because it's only assigned in the else branch.

brettkail-wk: Can you update the PR description with the error you're trying to fix?  I assumed it was `NameError: name 'principal_currency_value' is not defined` (and that `KeyError` in the summary was a mistake), but perhaps I misunderstand the problem.

austinmatherne-wk: I haven't seen the original exception, but I believe it does indeed throw a KeyError because of how this gets passed through and handled by the logger (it looks for an arg matching the name of the variable within the message).

brettkail-wk: The `principal_currency_value` isn't defined in the "then" branch, so shouldn't the code fail (before and after the PR) with `NameError` on the `principal_currency_value=principal_currency_value` line before `Validation.warning` fails to resolve `%(principal_currency_values)s`?  It seems like we're missing a test.

derekgengenbacher-wf: I restructured the validation to yield at the time of validation rather than the end of the validation.

arelle/plugin/validate/ROS/rules/ros.py
```diff
@@ -367,14 +371,10 @@ def rule_ros19(
...
-                            "which must match the functional(majority) unit of the financial statement.")
-                continue
-    if message:
-        yield Validation.warning(
-            "ROS.19",
-            message,
-            modelObject=modelObjects,
-            principal_currency_value=principal_currency_value
-        )
+                yield Validation.warning(
```
austinmatherne-wk, line 374: Can this warning potentially fire multiple times within the containing loop now?

brettkail-wk (approved):

PR #1362, opened 2024-09-24
Fix conflict between isEnumeration / isEnumeration2

paulwarren-wk:
#### Reason for change

Extensible enumeration validation is broken because `isEnumeration` and `isEnumeration2` cache to the same attribute. This means that these two methods will always return the same answer, and what that answer is depends on which one gets called first.

#### Description of change

Use separate attributes for caching.

#### Steps to Test

I encountered this because an EE1 validation error was not being detected.  I'm confused about how this can pass the conformance suite.

**review**:
@Arelle/arelle


aviary2-wf:
## Security Insights

## Action Items
---

brettkail-wk (approved):

austinmatherne-wk (approved):

PR #1391, opened 2024-10-27
Validate Report Packages

### LINKED ISSUE #1242: Make Report Packages 0xx Tests Pass

Following up on #1241, make updates to Arelle in order to get the 0xx (basic zip structure and package identification tests) set of tests to pass in CI.

### LINKED ISSUE #1243: Make Report Packages 1xx Tests Pass

Following up on #1241, make updates to Arelle in order to get the 1xx (structural JSON constraints for reportPackage.json) set of tests to pass in CI.

### LINKED ISSUE #1244: Make Report Packages 2xx Tests Pass

Following up on #1241, make updates to Arelle in order to get the 2xx (co-constraints on documentType and package file extension) set of tests to pass in CI.

### LINKED ISSUE #1246: Make Report Packages 4xx Tests Pass

Following up on #1241, make updates to Arelle in order to get the 4xx (invalid .xbri packages) set of tests to pass in CI.

### LINKED ISSUE #1248: Make Report Packages 6xx Tests Pass

Following up on #1241, make updates to Arelle in order to get the 6xx (invalid .xbr packages) set of tests to pass in CI.

### LINKED ISSUE #1250: Make Report Packages 8xx Tests Pass

Following up on #1241, make updates to Arelle in order to get the 8xx (invalid .zip packages) set of tests to pass in CI.

### LINKED ISSUE #1251: Make Report Packages 9xx Tests Pass

Following up on #1241, make updates to Arelle in order to get the 9xx (future report packages) set of tests to pass in CI.

austinmatherne-wk:
#### Reason for change
Resolves #1242, resolves #1243, resolves #1244, resolves #1246, resolves #1248, resolves #1250, resolves #1251.

#### Description of change
Adds report package validation support.

#### Steps to Test
* CI

**review**:
@Arelle/arelle


aviary2-wf:
## Security Insights

## Action Items
---

arelle/packages/report/ReportPackage.py
```diff
@@ -30,28 +31,75 @@ def _getReportPackageTopLevelDirectory(filesource: FileSource) -> str | None:
...
+    fullJsonFilePath = cast(str, filesource.url)
+    content = None
+    encodings = ["utf-8", "utf-8-sig"]
+    for encoding in encodings:
+        try:
+            with filesource.file(fullJsonFilePath, encoding=encoding)[0] as f:
+                content = cast(dict[str, Any], json.load(f, object_pairs_hook=forbidDuplicateKeys))
+                break
+        except (OSError, ValueError, zipfile.BadZipFile, json.JSONDecodeError):
+            continue
```
brettkail-wk, line 78: If we catch+ignore all errors, where do we report errors in the file?  Is there another code path where we do validations rather than loading the data?  If so, that's a little unfortunate, but maybe just a comment would be helpful.

arelle/plugin/validate/ESEF/__init__.py
```diff
@@ -131,14 +131,16 @@ def modelDocumentPullLoader(
                     if documentType in {ModelDocument.Type.TESTCASESINDEX, ModelDocument.Type.TESTCASE}:
                         return None  # allow zipped test case to load normally
 
-                if disclosureSystemYear >= 2023 and not modelXbrl.fileSource.isZip:
+                isZipFormat = modelXbrl.fileSource.isZip
+                hasZipFileExtension = modelXbrl.fileSource.type.lower() == ".zip"
+                if disclosureSystemYear >= 2023 and not (isZipFormat and hasZipFileExtension):
```
brettkail-wk, line 136: Do I understand that `.isZip` previously only returned true if the suffix was `.zip`, but we're changing it to return true for any zip archive, so we're adjusting this code to maintain the same semantics as before?  I wonder if that change in behavior would have any impact on plugins(?) and whether a different `.isZipArchive` would be preferable to changing the meaning of `.isZip`.

To be sure, I suppose some subsequent PR would change the logic here to allow `.xbri` here too, right?

austinmatherne-wk: > Do I understand that .isZip previously only returned true if the suffix was .zip, but we're changing it to return true for any zip archive, so we're adjusting this code to maintain the same semantics as before?

Correct.

> I wonder if that change in behavior would have any impact on plugins(?) and whether a different .isZipArchive would be preferable to changing the meaning of .isZip.

I conducted a search across Arelle and all known plugins, both within the Arelle repository and externally. The direct use of `isZip` is quite limited. Most other uses are actually checking if the file source is in a zip archive format. The more common usage is the `isArchive` property, which includes `isZip`, or `isTarGz`, etc. These uses focus on the file format rather than the file extension.

> To be sure, I suppose some subsequent PR would change the logic here to allow .xbri here too, right?

Yes, this change, along with others, will be necessary to complete support for ESEF 2024.

arelle/packages/report/ReportPackage.py
austinmatherne-wk: I pushed an update to log the error (same pattern is used in the `filesource.open()` method), but validation is deferred to validation classes that have the full context of which specifications are being validated against and which errors should be raised.

derekgengenbacher-wf (approved):

derekgengenbacher-wf:

PR #1421, opened 2024-11-25
Don't crash on missing units in ModelFact.isDuplicateOf.

stevenbronson-wk:
#### Steps to Test
Validate [s.zip](https://github.com/user-attachments/files/17909627/s.zip) and expect not to see
```
[DQC.US.0015.2688] Validation was unable to complete rule DQC.US.0015 due to an internal error.
```

**review**:
@Arelle/arelle

### DIFF
--- a/arelle/ModelInstanceObject.py
+++ b/arelle/ModelInstanceObject.py
@@ -512,7 +512,7 @@ class ModelFact(ModelObject):
             # parent test can only be done if in same instauce
             if self.modelXbrl == other.modelXbrl and self.parentElement != other.parentElement:
                 return False
-            if not (self.context.isEqualTo(other.context,dimensionalAspectModel=False) and
+            if not (self.context is not None and self.context.isEqualTo(other.context,dimensionalAspectModel=False) and
                     (not self.isNumeric or (self.unit is not None and self.unit.isEqualTo(other.unit)))):
                 return False
         elif self.isTuple:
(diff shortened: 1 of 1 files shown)


arelle/ModelInstanceObject.py
```diff
@@ -513,7 +513,7 @@ def isDuplicateOf(self, other, topLevel=True, deemP0Equal=False, unmatchedFactsS
             if self.modelXbrl == other.modelXbrl and self.parentElement != other.parentElement:
                 return False
             if not (self.context.isEqualTo(other.context,dimensionalAspectModel=False) and
-                    (not self.isNumeric or self.unit.isEqualTo(other.unit))):
+                    (not self.isNumeric or (self.unit is not None and self.unit.isEqualTo(other.unit)))):
```
austinmatherne-wk, line 516: What if both `self.unit` and `other.unit` are `None`?

brettkail-wk, line 516: I wonder whether a numeric fact should be considered duplicate if both of them are missing a unit.

brettkail-wk (approved):

arelle/ModelInstanceObject.py
brettkail-wk: Austin was first :-).

austinmatherne-wk: The use of xValid checks makes sense for specific validations where we know certain conditions can’t be met. However, in this case, we’re dealing with a generic function. Since this code is also used by formula, and Arelle doesn’t and shouldn't restrict formula processing to xValid facts, I don’t think we can expect callers to pre-filter.

Additionally, I noticed a similar issue with context usage in the line above; it can also be None.

austinmatherne-wk: That's reasonable. Can you add a not None check for self.context on the line above so that it behaves the same?

brettkail-wk (approved):

austinmatherne-wk (approved):

PR #1490, opened 2025-01-13
Adding DBA.FR89

allenbross-wf:
#### Description of change

Adding DBA.FR89

#### Steps to Test

CI

**review**:
@Arelle/arelle


aviary2-wf:
## Security Insights

## Action Items
---

arelle/plugin/validate/DBA/rules/fr.py
```diff
@@ -797,6 +797,55 @@ def rule_fr81(
...
+                    pluginData.reportingClassDEnglish,
+                ]:
+                    auditorFacts = modelXbrl.factsByQname.get(pluginData.typeOfAuditorAssistanceQn)
+                    if auditorFacts is not None:
+                        for auditorFact in auditorFacts:
+                            if auditorFact.xValid >= VALID and auditorFact.xValue not in [
+                                pluginData.auditedFinancialStatementsDanish,
+                                pluginData.auditedFinancialStatementsEnglish
+                            ]:
+                                yield Validation.error(
```
austinmatherne-wk, line 837: This error could potentially be raised multiple times if there are multiple classOfReportingEntity or typeOfAuditorAssistance facts. Is that the intended behavior?

Additionally, no error is raised when there are matching classOfReportingEntity values but no valid auditor facts. That seems like an oversight.

Note that some work within each iteration of the loop, such as building a list of reporting class values and performing auditor fact lookups, is not dependent on the iteration and could be done once outside the loop for efficiency.

arelle/plugin/validate/DBA/rules/fr.py
```diff
@@ -797,6 +797,59 @@ def rule_fr81(
...
+                    pluginData.reportingClassDEnglish,
+                ]:
+                    classFacts.append(fact)
+                    if auditorFacts is not None:
+                        for auditorFact in auditorFacts:
+                            if auditorFact.xValid >= VALID and auditorFact.xValue in [
+                                pluginData.auditedFinancialStatementsDanish,
+                                pluginData.auditedFinancialStatementsEnglish
+                            ]:
+                                return
```
austinmatherne-wk, line 839: This auditor fact check isn't dependent on the state of the reporting facts loop. It should be performed once before the iteration to avoid redundant computation.

austinmatherne-wk (dismissed):

arelle/plugin/validate/DBA/rules/fr.py
```diff
@@ -797,6 +797,59 @@ def rule_fr81(
...
+                if fact.xValue in [
+                    pluginData.reportingClassCLargeDanish,
+                    pluginData.reportingClassCLargeEnglish,
+                    pluginData.reportingClassCMediumDanish,
+                    pluginData.reportingClassCMediumEnglish,
+                    pluginData.reportingClassDDanish,
+                    pluginData.reportingClassDEnglish,
+                ]:
+                    classFacts.append(fact)
+        if len(classFacts) > 0:
```
derekgengenbacher-wf, line 840: I believe that this logic is wrong. This should be checking to see if `classOfReportingEntityQn` is one of the specified values and if it is then it should look at the value of `typeOfAuditorAssistanceQn`. If `typeOfAuditorAssistanceQn` is not in the specified values then we fire an error.

arelle/plugin/validate/DBA/rules/fr.py
```diff
@@ -797,6 +797,59 @@ def rule_fr81(
...
+    """
+    DBA.FR89: If ClassOfReportingEntity is one of the following values:
+
+    Regnskabsklasse C, mellemstor virksomhed // Reporting class C, medium-size enterprise
+    Regnskabsklasse C, stor virksomhed // Reporting class C, large enterprise
+    Regnskabsklasse D // Reporting class D
+    Then TypeOfAuditorAssistance should be: Revisionspåtegning // Auditor's report on audited financial statements
+    """
+    modelXbrl = val.modelXbrl
+    facts = modelXbrl.factsByQname.get(pluginData.classOfReportingEntityQn)
```
derekgengenbacher-wf, line 818: Since we are not using facts until line 827 and there is a possibility that we may short circuit out of the validation before using facts I think it makes sense to push this line down to 827ish.

austinmatherne-wk (approved):

derekgengenbacher-wf (approved):

derekgengenbacher-wf:

PR #1498, opened 2025-01-15
Add DBA.TR19

derekgengenbacher-wf:
#### Reason for change
Add DBA.TR19

#### Steps to Test
CI

**review**:
@Arelle/arelle


aviary-wf:
## Security Insights

## Action Items
---

arelle/plugin/validate/DBA/rules/tr.py
```diff
@@ -0,0 +1,55 @@
...
+        *args: Any,
+        **kwargs: Any,
+) -> Iterable[Validation]:
+    """
+    DBA.TR19: Duplicate facts must not have different content. "Duplicate" is defined by the following criteria:
+    - same concept name
+    - same period
+    - same dimensions (including same identifier in typed dimension)
+    - same language (xml:lang)
+    - same unit
```
austinmatherne-wk, line 40: Do we know if this rule is trying to prevent reporting equivalent facts with different scale values? The description makes it seem like that's the case, but I'm not sure if "content" is the literal value in the HTML, or the parsed and transformed XML value.

derekgengenbacher-wf: Good question. Im not sure wither. Ill ask Peter.

derekgengenbacher-wf: Peter says its the raw HTML/XML value and not the scaled value.

allenbross-wf (approved):

allenbross-wf:
QA +10

arelle/plugin/validate/DBA/rules/tr.py
brettkail-wk: @austinmatherne-wk Rather than facts with different scales, I think you're asking about different _accuracies_ (where differences might or might not occur due to different scales: {value=123,scale=3,decimals=0} and {value=123000,scale=0,decimals=0} are equivalent)?  Using a "raw" value (visible HTML text?) doesn't make sense to me, but I could imagine requiring all facts have the same accuracy+value, so we would just need to check precision in addition to `xValue`.

austinmatherne-wk: Yes, exactly. The rule isn't very clear (I'm assuming it's translated). But enforcing literal value equality seems strange to me.

PR #1526, opened 2025-02-04
Update disclosure systems for Danish DBA validations

derekgengenbacher-wf:
#### Reason for change
- Create new multi-target disclosure system
 -Define appropriate validations for each disclosure system

#### Steps to Test
CI

**review**:
@Arelle/arelle


arelle/plugin/validate/DBA/rules/tr.py
```diff
@@ -17,12 +17,22 @@
...
 _: TypeGetText
 
 
 @validation(
     hook=ValidationHook.XBRL_FINALLY,
+    disclosureSystems=[
+        ARL_2022_PREVIEW,
+        ARL_2024_PREVIEW,
+        ARL_2024_MULTI_TARGET_PREVIEW
+    ],
```
austinmatherne-wk, line 35: Are the only validations that need to specify their disclosure systems the ones that don’t run for multi-target? To simplify maintenance, it would be better to avoid explicitly specifying validations that apply to all disclosure systems, so we don’t have to update them annually. Additionally, we could define a list variable containing all single-target disclosure systems and reference it throughout for validations that don't run for multi-target, reducing the need to update individual rules next year if we expect them to stick around.

arelle/plugin/validate/DBA/rules/fr.py
```diff
@@ -50,6 +52,7 @@ def rule_fr1(
 
 @validation(
     hook=ValidationHook.XBRL_FINALLY,
+    disclosureSystems=ALL_DISCLOSURE_SYSTEMS,
```
austinmatherne-wk, line 55: I guess I wasn't very clear with my previous comment. `ALL_DISCLOSURE_SYSTEMS` is the default behavior, so you don't need to specify this. Just the rules which use `STAND_ALONE_DISCLOSURE_SYSTEMS` or `MULTI_TARGET_DISCLOSURE_SYSTEMS`.

austinmatherne-wk (approved):

arelle/plugin/validate/DBA/rules/fr.py
derekgengenbacher-wf: Ahhh, I knew that and for some reason completely forgot. Fixing now.

austinmatherne-wk (approved):

PR #1533, opened 2025-02-07
Adding DBA.TR_15 and related multi-target test case suite

allenbross-wf:
#### Description of change

Adding DBA.TR_15 and related multi-target test case suite

#### Steps to Test

CI

**review**:
@Arelle/arelle


arelle/plugin/validate/DBA/rules/tr.py
```diff
@@ -192,6 +192,33 @@ def rule_tr09(
...
+        **kwargs: Any,
+) -> Iterable[Validation]:
+    """
+    DBA.TR15: If the context identifier scheme is http://standards.iso.org/iso/17442, all context identifiers must be equal to the value of the field gsd:LegalEntityIdentifierOfReportingEntity.
+    """
+    contextEntityIdentifiers = {context.entityIdentifier for context in val.modelXbrl.contexts.values()}
+    leiFacts = val.modelXbrl.factsByQname.get(pluginData.legalEntityIdentifierOfReportingEntityQn, set())
+    if len(leiFacts) > 0:
+        leiFact = next(iter(leiFacts))
+        if leiFact.context is not None and leiFact.context.entityIdentifier[0] == 'http://standards.iso.org/iso/17442' and len(contextEntityIdentifiers) >0:
```
derekgengenbacher-wf, line 212: This is checking all of the context entity identifiers. So this duplicates TR09. I would just check that the context on the lei fact matches the fact value. And then that lets TR09 does its thing.

derekgengenbacher-wf: This line should be something like:
```suggestion
        if leiFact.context is not None and leiFact.context.entityIdentifier[0] == 'http://standards.iso.org/iso/17442' and leiFact.xValid >= VALID and leiFact.context.entityIdentifier[1] != leiFact.xValue:
```

This allows you to remove all of the `contextEnitytIdentifier` variable stuff.

derekgengenbacher-wf (approved):

derekgengenbacher-wf:

PR #1550, opened 2025-02-20
ESEF.1.4.1 doesn't check datatype namespace URIs

austinmatherne-wk:
#### Reason for change
Arelle’s QName string representation follows a non-standard prefix:namespace format, making it unsuitable for equality checks. Because of this, the ESEF 2.4.1 validation checks that data type prefixes and local names match instead of the namespace URI and local name.

#### Description of change
Perform direct QName comparison between data types in the ESEF 1.4.1 validation.

#### Steps to Test
* CI

**review**:
@Arelle/arelle


brettkail-wk (approved):

derekgengenbacher-wf (approved):

derekgengenbacher-wf:

PR #1558, opened 2025-03-03
Consolidate and fix code for handling schema whitespace normalization replace and collapse

strr:
#### Reason for change
There are a variety of implementations of schema whitespace normalization, in particular replace and collapse, through the Arelle codebase. Some of them accidentally do too much, too little or both. For example, `str.strip()` (the no argument variant) both does too much by removing all leading and trailing whitespace characters and does too little by not replacing and/or collapsing the relevant whitespace characters when they are inside the string body.

References:
- https://www.w3.org/TR/2004/REC-xmlschema-2-20041028/datatypes.html#rf-whiteSpace
- https://www.w3.org/TR/xmlschema-1/#d0e1654

#### Description of change
Update `XmlUtil.collapseWhitespace` and add `XmlUtil.replaceWhitespace` and use them in place of various other implementations in the codebase. With `XmlUtil.replaceWhitespace`, we make use of `str.translate`  as it's about 10x faster than a regex. With `XmlUtil.collapseWhitespace`, we don't need to find every space, just every run of two or more spaces, so we avoid having to split & join any run of already collapsed text.

Change `ModelFact.effectiveValue` to prefer using the PSVI value (`xValue`) for non-numeric facts so that the normalized fact values are displayed.

#### Steps to Test

1. Download [sample_report_and_taxonomy.zip](https://github.com/user-attachments/files/19060800/sample_report_and_taxonomy.zip)
2. Open it in a published release of Arelle in the GUI. You can see the first three fact values all look the same despite the underlying data types meaning they should look different post schema validation. The values are not actually the raw values either due to the `effectiveValue.strip()` calls. If you don't like the GUI, just use `--facts` and dump the facts to CSV to see the same thing.
3. It will look a bit like this arelle-pre-whitespace-fixes
4. Try doing the same thing in a build with this set of changes in. It will look a bit like this arelle-post-whitespace-fixes

**review**:
@Arelle/arelle

### DIFF
--- a/arelle/XmlUtil.py
+++ b/arelle/XmlUtil.py
@@ -202,9 +202,9 @@ def innerText(
     try:
         text = "".join(text for text in innerTextNodes(element, ixExclude, ixEscape, ixContinuation, ixResolveUris))
         if strip:
-            # Defined as removing leading and trailing [ \t\r\n], the same
-            # characters as xs:whiteSpace operates on. Note, this is a different
-            # normlization to any of whiteSpace = preserve/replace/collapse
+            # For legacy reasons, strip=True removes leading/trailing
+            # whitespace even though most callers probably want
+            # collapseWhitespace semantics instead.
             return text.strip(" \t\r\n")
         return text
     except (AttributeError, TypeError):
--- a/arelle/XmlUtil.py
+++ b/arelle/XmlUtil.py
@@ -202,7 +202,10 @@ def innerText(
     try:
         text = "".join(text for text in innerTextNodes(element, ixExclude, ixEscape, ixContinuation, ixResolveUris))
         if strip:
-            return text.strip(" \t\r\n") # strip follows xml whitespace collapse, only blank, tab and newlines
+            # Defined as removing leading and trailing [ \t\r\n], the same
+            # characters as xs:whiteSpace operates on. Note, this is a different
+            # normlization to any of whiteSpace = preserve/replace/collapse
+            return text.strip(" \t\r\n")
         return text
     except (AttributeError, TypeError):
         return ""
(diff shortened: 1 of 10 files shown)


arelle/XmlUtil.py
```diff
@@ -199,7 +202,10 @@ def innerText(
     try:
         text = "".join(text for text in innerTextNodes(element, ixExclude, ixEscape, ixContinuation, ixResolveUris))
         if strip:
-            return text.strip(" \t\r\n") # strip follows xml whitespace collapse, only blank, tab and newlines
+            # Defined as removing leading and trailing [ \t\r\n], the same
+            # characters as xs:whiteSpace operates on. Note, this is a different
+            # normlization to any of whiteSpace = preserve/replace/collapse
+            return text.strip(" \t\r\n")
```
brettkail-wk, line 208: The comment says "defined", but do you know what defines this?  It seems like some of the callsites are using it for UI display purposes where it doesn't really matter, but it seems like others maybe want collapse semantics (e.g., `ModelInlineValueObject.rawValue` even has a comment saying so).

brettkail-wk: > I meant Arelle has defined strip = True to mean this as I was confused by the previous comment.
> [...] I saw that comment in `ModelInlineValueObject.rawValue` too

I guess I was asking more broadly: how many callsites with `strip=True` (implicit or explicit) actually want `collapseWhitespace` (or should be updated to `strip=False` + `collapseWhitespace` if that's too invasive)?

Perhaps fixing all that is too much for this PR.  In that case, perhaps the comment should say "For legacy reasons, strip=True only removes leading/trailing whitespace only even though most callers probably want collapseWhitespace instead."

> some (all?) of the current entries [...] What do you think?

It seems like an incremental improvement, but the "some (all?)" makes me think the collapse should be in the transforms themselves, not `rawValue`.  If that's too invasive for now, it seems the comment should admit something like "Most transforms are whitespace-collapse, so do that now. TODO: this collapsing should be moved to the individual transformation functions in order to support non-collapsing transformations."

brettkail-wk (approved):

austinmatherne-wk (approved):

PR #1560, opened 2025-03-04
Update FR35 and FR59 to check ConsolidatedSoloDimension when it exists

derekgengenbacher-wf:
#### Reason for change
Update FR35 and FR59 to check ConsolidatedSoloDimension when it exists

#### Steps to Test
CI

**review**:
@Arelle/arelle

### DIFF
--- a/arelle/plugin/validate/DBA/rules/__init__.py
+++ b/arelle/plugin/validate/DBA/rules/__init__.py
@@ -308,3 +308,15 @@ def minimumRequiredFactsFound(
         if count >= requiredCount:
             return True
     return False
+
+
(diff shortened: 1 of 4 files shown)


arelle/plugin/validate/DBA/rules/__init__.py
```diff
@@ -308,3 +308,15 @@ def minimumRequiredFactsFound(
         if count >= requiredCount:
             return True
     return False
+
+
+def consolidated_dimension_exists(modelXbrl: ModelXbrl, consolidatedSoloQn: QName) -> bool:
```
allenbross-wf, line 313: Arbitrary line.  The functions and variables you added should be camel case to be consistent with the rest of Arelle.

allenbross-wf (approved):

austinmatherne-wk (approved):

PR #1563, opened 2025-03-05
Add missing validation that if ix:nonFraction has a text node child, it must be a non-empty string.

### LINKED ISSUE #1562: InlineXBRL 10.1.1 nonFraction with empty string not detected as error
### What happened?

An InlineXBRL nonFraction-tag who has no value (empty string), and is not nil, should generate a validation error. But this error was not reported.

Example:

```
<ix:nonFraction contextRef="Context_11" unitRef="U-EUR" name="ifrs-full:IssueOfEquity" id="f2183m1744p" decimals="-6" scale="6" format="ixt:fixed-zero"></ix:nonFraction>
```

InlineXBRL specification 10.1.1:

> [...]
> 
> The ix:nonFraction element MUST have exactly one child which SHALL be either an ix:nonFraction element or a text node, unless it has an xsi:nil attribute with the value true.
> 
> [...]
> 
> The text node child, if present, MUST be a non-empty string.
> 

### Documents

_No response_

_No response_

### Interface

CLI (arelleCmdLine)

### Version

2.36.18

### Download

arelle.org (bundled Python)

### Operating System

Windows

strr:
#### Reason for change
Increase Inline XBRL 1.1 Specification Conformance by adding missing validation. Fixes #1562 

#### Description of change
Check that the `ix:nonFraction` element is not nil, has no children and that the text node child is empty … if so raise `[ix11.10.1.1:nonFraction]`

#### Steps to Test

1. Open [sample_report_and_taxonomy_nonfraction.zip](https://github.com/user-attachments/files/19096420/sample_report_and_taxonomy_nonfraction.zip)
2. Arelle should raise two `[ix11.10.1.1:nonFraction]` but only one `[xmlSchema:valueError]` as the `fixed-zero` transform on one of the facts prevents the value error

**review**:
@Arelle/arelle

### DIFF
--- a/arelle/ValidateXbrlDTS.py
+++ b/arelle/ValidateXbrlDTS.py
@@ -975,6 +975,10 @@ def checkElements(val, modelDocument, parent):
                                 modelObject=(elt,e2))
                     else:
                         c = XmlUtil.children(elt, '*', '*')
+                        if not c and elt.text is None:
+                            val.modelXbrl.error(ixMsgCode("nonFraction", elt, sect="constraint"),
+                                _("Inline XBRL ix:nonFraction The text node child, if present, MUST be a non-empty string."),
+                                modelObject=elt)
                         if c and (len(c) != 1 or c[0].namespaceURI != elt.namespaceURI or c[0].localName != "nonFraction"):
                             val.modelXbrl.error(ixMsgCode("nonFractionChildren", elt, sect="validation"),
                                 _("Inline XBRL nil ix:nonFraction may only have one child ix:nonFraction"),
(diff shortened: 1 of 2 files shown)


austinmatherne-wk (approved):

arelle/ValidateXbrlDTS.py
```diff
@@ -975,6 +975,10 @@ def linkbaseTopElts():
                                 modelObject=(elt,e2))
                     else:
                         c = XmlUtil.children(elt, '*', '*')
+                        if not c and XmlUtil.innerText(elt, strip = False) == '':
+                            val.modelXbrl.error(ixMsgCode("nonFraction", elt, sect="constraint"),
+                                _("Inline XBRL ix:nonFraction The text node child, if present, MUST be a non-empty string."),
```
brettkail-wk, line 980: It's somewhat worrying that there is no conformance test for this.

brettkail-wk (approved):

PR #1564, opened 2025-03-06
Format Boolean fact values as lowercase

austinmatherne-wk:
#### Reason for change
A recent change to effectiveValue used `str(xValue)`. However, while in Python `str(bool)` returns `True/False`, XML Boolean values are lowercase `true/false`.

effectiveValue is used for rendering fact values within the GUI and for formatting values in the XML log.

#### Description of change
Render boolean values using their XML valid format.

#### Steps to Test
* CI

**review**:
@Arelle/arelle


derekgengenbacher-wf (approved):

derekgengenbacher-wf:

PR #1568, opened 2025-03-10
Check either/or on consolidatedSolo dimension rather than both

derekgengenbacher-wf:
#### Reason for change
Check the either/or case instead of both around consolidatedSolo dimension and non dimension cases

#### Steps to Test
CI

**review**:
@Arelle/arelle


allenbross-wf (approved):

allenbross-wf:
QA +10
All tests passed.

PR #1569, opened 2025-03-11
Validate Base Taxonomies

austinmatherne-wk:
#### Reason for change

Arelle currently skips full DTS validation for documents that aren’t considered part of an extension taxonomy. However, this can result in undetected errors when a report references a broken taxonomy.

This update ensures full validation for all DTS documents while maintaining the existing behavior of not invoking the plugin `Validate.XBRL.DTS.document` hook on base taxonomy documents, preventing extension taxonomy rules from being incorrectly applied to them.

If Arelle had implemented this earlier, it would have flagged errors in all NT filings due to the use of the standard role on [a linkbaseRef](https://www.nltaxonomie.nl/2016/xbrl/sbr-labelroles.xsd) within the base taxonomy. Last year, a couple of test cases were added to the XBRL 2.1 conformance suite to enforce this validation. However, they were later removed due to real-world usage in NT taxonomies. An upcoming conformance suite will now explicitly require validators to allow this behavior.

#### Description of change

- Updated the XBRL 2.1 conformance suite.
- Allowed linkbase references to use standard roles (a practice used in NT taxonomies and soon to be required in an upcoming 2.1 test suite).
- Enabled full validation for all DTS documents.
- Added performance optimization option to disable full base taxonomy validation (prior behavior).

#### Steps to Test

- CI

**review**:
@Arelle/arelle


brettkail-wk (approved):

derekgengenbacher-wf (approved):

derekgengenbacher-wf:

PR #1573, opened 2025-03-13
Taxonomy Package autoloader

hermfischer-wf:
#### Reason for change
It's a PIA to specify taxonomy packages manually with the GUI, or to list them out in the cmd line, when they are loadable from the XBRL Taxonomy Registry.

#### Description of change
This plugin grabs the XBRL Taxonomy Packages Registry and checks all references to http[s] URLs to see if any appear as a taxonomy package entry point.  If found to be an entry point and the defining taxonomy package isn't loaded, calls the TaxonomyPackage loader for the package.

If the taxonomy package is malformed or has mappings conflicting with prior-loaded taxonomy packages this will raise an 
Arelle warning.

(Suggest that there should be a CI script at XII which fires up arelle to load every taxonomy package in the registry in a single run, and give black stars to those raising mapping conflicts.)

(If operating with a preview unpublished taxonomy package not yet in Registry, it still has to be manually specified by GUI or cmd line.)

Please see if you can come up with a shorter or better name for the plugin and its methods.

#### Steps to Test
Suggest:
  1. Clean out cache (GUI tools menu).
  2. Try an instance with entry points which are in a package in the registry (us-gaap, srt, ifrs, etc).
  3. Verify that package was inserted into cache and entry point file was not inserted into cache (and that nothing else broke).

**review**:
@Arelle/arelle

### DIFF
--- a/arelle/plugin/txmyPkgAutoLoad.py
+++ b/arelle/plugin/txmyPkgAutoLoad.py
@@ -1,0 +1,47 @@
+"""
+
+Auto load taxonomy package for entry points in XBRL Taxonomy Registry (via STANDARD_PACKAGES_URL).
+
+See COPYRIGHT.md for copyright information.
+"""
+
+import json
+
+from arelle import PackageManager
+from arelle.DialogPackageManager import STANDARD_PACKAGES_URL
+from arelle.Version import authorLabel, copyrightLabel
+
+entryPointsPackageUrl = None
+
+def taxonomyPackageAutoLoad(_cntlr, packagesConfig, url):
+    global entryPointsPackageUrl
+    if packagesConfig is not None and url is not None and url.startswith("http"):
+        if entryPointsPackageUrl is None: # load standard taxonomy packages
+            entryPointsPackageUrl = {}
+            with open(_cntlr.webCache.getfilename(STANDARD_PACKAGES_URL, reload=True), 'r', errors='replace') as fh:
+                regPkgs = json.load(fh) # always reload
+            for pkgTxmy in regPkgs.get("taxonomies", []):
+                _url = pkgTxmy.get("Links",{}).get("AuthoritativeURL")
+                if _url.endswith(".zip"):
+                    # Ignore taxonomies that lack a direct download link.
+                    # Although future valid taxonomy download links from the registry might not have a ".zip" extension,
+                    # as of October 27th, 2024, all links that without a zip file extension have been invalid.
+                    for _ep in pkgTxmy.get("EntryPoints") or ():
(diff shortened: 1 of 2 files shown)


arelle/plugin/txmyPkgAutoLoad.py
```diff
@@ -0,0 +1,47 @@
...
+                regPkgs = json.load(fh) # always reload
+            for pkgTxmy in regPkgs.get("taxonomies", []):
+                _url = pkgTxmy.get("Links",{}).get("AuthoritativeURL")
+                if _url.endswith(".zip"):
+                    # Ignore taxonomies that lack a direct download link.
+                    # Although future valid taxonomy download links from the registry might not have a ".zip" extension,
+                    # as of October 27th, 2024, all links that without a zip file extension have been invalid.
+                    for _ep in pkgTxmy.get("EntryPoints") or ():
+                        for _doc in _ep.get("EntryPointDocuments") or ():
+                            entryPointsPackageUrl[_doc] = _url
```
austinmatherne-wk, line 31: Are there any packages in the registry which cover the same URLs that this would silently treat as last one wins?

arelle/plugin/txmyPkgAutoLoad.py
```diff
@@ -0,0 +1,47 @@
...
+        if url in entryPointsPackageUrl and not any(
+            url.startswith(mapFrom) and not url.startswith(mapTo)
```
austinmatherne-wk, line 33: It seems like this check (does the package manager already have a matching package) would be better performed before the download and parsing steps above?

arelle/plugin/txmyPkgAutoLoad.py
hermfischer-wf: I'll add a validation function (for this):
  (a) same Entry Point URLs
  (b) overlapping redirects

arelle/plugin/txmyPkgAutoLoad.py
hermfischer-wf: The preceding "if entryPointsPackageUrl is None" loads entryPointsPackageUrl the first time (when used in a given invocation).  Otherwise "url in entryPointsPackageUrl" would be asking "url in None" before the registry is loaded.

austinmatherne-wk: I'm thinking of the case of the GUI, where someone has this plugin enabled as a catch all, but they frequently already have the required packages. It will redownload the registry once for each instance of the GUI regardless of if they need it.

Are there any downsides to performing this check earlier?

PR #1582, opened 2025-03-17
When displaying strings, prefer to use them after whitespace normalization

### LINKED ISSUE #1581: Facts against concepts with data-types with schema whitespace normalization are not being displayed correctly
### What happened?

If you grab a copy of Arelle 2.36.26 and open the attached report package, you get a fact list that looks like this:

Image

Whereas if the whitespace normalization specified for the data types has been applied, you should get:

Image

You get the same results with `--facts` on the command line.

Judging by 37a7ae98469a27c08a6fd44e28a51c3aae831bc9 , there are some requirements `ModelFact.effectiveValue()` needs to meet so putting this issue here to try and get to the bottom of it.

### Documents

[sample_report_and_taxonomy.zip](https://github.com/user-attachments/files/19292835/sample_report_and_taxonomy.zip)

_No response_

### Interface

GUI (arelleGUI)

### Version

2.36.26

### Download

GitHub release (bundled Python)

### Operating System

Windows

strr:
For anything that is still of string data type after schema validation, prefer to display the string after schema whitespace normalization.

#### Reason for change
Fixes #1581 

#### Description of change
Use the PSVI value as the `effectiveValue` for display if and only if the PSVI type is still string.

This is a more minimal change than previously attempted as part of 6568d5169d15a54ed31fd6ed7785cda56ab80d52 and reverted in 37a7ae98469a27c08a6fd44e28a51c3aae831bc9 so hopefully will be more successful in working with whatever downstream requirements exist.

#### Steps to Test

**review**:
@Arelle/arelle

### DIFF
--- a/arelle/ModelInstanceObject.py
+++ b/arelle/ModelInstanceObject.py
@@ -428,6 +428,9 @@ class ModelFact(ModelObject):
             # non-numeric fact at this point
             if len(val) == 0: # zero length string for HMRC fixed fact
                 return "(reported)"
+            if self.xValid >= VALID and isinstance(self.xValue, str):
+                # Prefer the xs:whiteSpace normalized version of strings
+                return self.xValue
             return val
         except Exception as ex:
             return str(ex)  # could be transform value of inline fact


austinmatherne-wk (approved):

austinmatherne-wk:
Hi @strr,

Thanks for the PR. The last change caused an issue with how values were formatted in XML logs which mostly came down to how non-string values were transforms by `str()`. This looks like a reasonable solution.

PR #1592, opened 2025-03-22
Fix Calculation GUI View

austinmatherne-wk:
#### Reason for change
Calc 1.1 support introduced more bugs to the GUI calc views.

* Weight and balance columns are no longer displayed.
* Saving calc table to csv throws an exception.

image

#### Description of change
* Correctly detect calculation arcroles.
* Fix thrown exception on saving calc table view to an excel file.

#### Steps to Test
* Open a report with calculations (either 1.1 or XBRL 2.1).
* Confirm that weight and balance columns are populated in the calculation table view.
* With the calculation tab selected use the file > save menu to save the table to an excel spreadsheet and confirm no exceptions are thrown.

**review**:
@Arelle/arelle


derekgengenbacher-wf (approved):

brettkail-wk (approved):

allenbross-wf:
QA +10

PR #1600, opened 2025-03-25
Split args from message in JSON log

austinmatherne-wk:
#### Reason for change
For the JSON log the same object is used for storing the message text (`text` key) along with the message arguments. If there is [an arg with the key name](https://github.com/Arelle/EDGAR/blob/4196818170991c916f6a40c73c0e115484f23f03/render/resources/arelleMessagesText.xml#L1681) `text` it overwrites the message text.

This update moves the arguments to a separate object in the json log so names don't collide.

#### Description of change
Move log message args to "args" object in JSON log.

#### Steps to Test
* CI

**review**:
@Arelle/arelle


arelle/logging/handlers/LogHandlerWithXml.py
```diff
@@ -68,14 +68,19 @@ def propElts(properties: list[tuple[Any, Any, Any]], indent: str, truncateAt: in
...
     def recordToJson(self, logRec: logging.LogRecord) -> dict[str, Any]:
-        message = { "text": self.format(logRec) }
-        if logRec.args and isinstance(logRec.args, Mapping):
-            for n, v in logRec.args.items():
-                message[n] = str(v)
         return {"code": getattr(logRec, "messageCode", ""),
                 "level": logRec.levelname.lower(),
                 "refs": getattr(logRec, "refs", []),
-                "message": message}
+                "message": self.format(logRec),
```
brettkail-wk, line 74: Do you know how the JSON format is used?  Is there a risk of this breaking someone by changing the meaning of the keys?  I wonder if `{"message": {"text": self.format(logRec), "args": self._buildJsonMessageArgs(logRec)}}` is marginally more compatible...

brettkail-wk (approved):

arelle/logging/handlers/LogHandlerWithXml.py
austinmatherne-wk: I don't. From folks that I've talked to the json format does not appear to be widely used, but we don't have concrete usage details. I agree, your suggestion is better. I've updated the message format.

brettkail-wk (approved):

arelle/logging/handlers/LogHandlerWithXml.py
austinmatherne-wk: It was identified by a service provider who sent a support email.

allenbross-wf:
QA +10

PR #1604, opened 2025-03-27
Workaround Python stdin bug when run as Windows service 

austinmatherne-wk:
#### Reason for change
Resolves google group issue [ktGBv3B-Do4](https://groups.google.com/g/arelle-users/c/ktGBv3B-Do4)

#### Description of change
When making a call to subprocess.run/Popen and:
1. Either stdout, or stderr is provided, or capture_output is True.
2. stdin is not provided.
3. Arelle is run as a Windows service.

An OSError is raised by a call to _winapi.GetStdHandle to set stdin. This can be avoided by explicitly specifying stdin.

#### Steps to Test
* CI

**review**:
@Arelle/arelle


derekgengenbacher-wf (approved):

derekgengenbacher-wf:

PR #1609, opened 2025-04-01
Plugin hook for transforming URLs within WebCache

aaroncameron-wk:
#### Reason for change
Some plugins may want to map URLs to their own file locations

#### Description of change
Add plugin hook to allow transformation of URLs before normal mapping to web cache filepaths, or to transform URLs into filepaths directly.

#### Steps to Test
CI

**review**:
@Arelle/arelle

### DIFF
--- a/arelle/utils/PluginHooks.py
+++ b/arelle/utils/PluginHooks.py
@@ -790,3 +790,26 @@ class PluginHooks(ABC):
         :return: None
         """
         raise NotImplementedError
+
+    @staticmethod
+    def webCacheTransformUrl(
+            cntlr: CntlrCmdLine,
+            url: str | None,
+            base: str | None,
+            *args: Any,
+            **kwargs: Any,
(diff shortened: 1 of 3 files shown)


arelle/utils/PluginHooks.py
```diff
@@ -790,3 +790,26 @@ def saveLoadableOimSave(
...
+            cntlr: CntlrCmdLine,
+            url: str | None,
+            base: str | None,
+            *args: Any,
+            **kwargs: Any,
+    ) -> tuple[str | None, bool]:
+        """
+        Plugin hook: `WebCache.TransformURL`
+
+        This hook is triggered after before the web cache maps a URL to a file path.
```
brettkail-wk, line 805:
```suggestion
        This hook is triggered before the web cache maps a URL to a file path.
```

austinmatherne-wk (approved):

PR #1625, opened 2025-04-17
Use qname to lookup formula param

### LINKED ISSUE #1386: ESEF Validation: RegEx pattern for IFRS namespace does not allow https
### What happened?

When we run ESEF validation we are getting an unexpected error message ESEF.RTS.ifrsRequired. Our investigation indicates that 2022 filings are getting validated by both ESEF_2021 and ESEF_Current:

1. Validation first run on exported data without target attribute
2. Validation run a second time on things that have target "UKFRS"

The ESEF.RTS.ifrsRequired false positive appears to be caused by this and RegEx that does not allow secure https 

image

------------------------------------------------------------------------------------------------------------------------------------

Did some extra investigation on ESEF.RTS.ifrsRequired warning and it seems this is related to the configuration of plugins that are used on different viewers.
As a baseline I took this filing [https://filings.xbrl.org/YF0Y5B0IB8SM0ZFG9G81/2023-12-31/ESEF/GB/0/YF0Y5B0IB8SM0ZFG9G81-2023-12-31/reports/ixbrlviewer.html#](https://filings.xbrl.org/YF0Y5B0IB8SM0ZFG9G81/2023-12-31/ESEF/GB/0/YF0Y5B0IB8SM0ZFG9G81-2023-12-31/reports/ixbrlviewer.html)

In the "official" viewer it only has a single warning:
Screenshot 2024-07-26 at 17.06.23.png

But when I validate the same project with Desktop Arelle and these plugins enabled
Screenshot 2024-07-26 at 17.12.21.png
I am getting a bit more validation messages:
Screenshot 2024-07-26 at 17.06.02.png
And these messages match what we are getting from Arelle Rest API when validating the same project:
Screenshot 2024-07-26 at 17.06.13.png

It seems this warning, a few others and the error are reported by validate/ESEF (Validate ESMA ESEF) plugin and the filings.xbrl.org are generating the viewers without using this plugin.
If I disable Validate ESMA ESEF plugin on Desktop Arelle I am also getting only a single message:
Screenshot 2024-07-26 at 17.08.36.png

### Documents

_No response_

_No response_

### Interface

CLI (arelleCmdLine)

### Version

Latest master

### Download

From source (git checkout)

### Operating System

Windows

austinmatherne-wk:
#### Reason for change
Resolves #1386

Not mentioned in the above ticket, but the issue was that passing the ixds target via a parameter didn't actually work because the param was looked up by string, while the keys [are QNames](https://github.com/Arelle/Arelle/blob/9b130be52b4f5987cf5d54c768b8458b683eec60/arelle/CntlrCmdLine.py#L953).

#### Description of change
Use a qname of `ixdsTarget` to check if an ixdsTarget was passed as a formula parameter.

#### Steps to Test
* Run `python arelleCmdLine.py --internetRecheck=never --plugins=validate/ESEF --disclosureSystem=esef --file https://github.com/user-attachments/files/19609180/529900T8BM49AURSDO55-2024-12-31.zip --validate --parameters="eps_threshold=.01,ixdsTarget=(default),authority=UKFRC-2023"`
* Verify that no `ESEF.RTS.ifrsRequired` errors are raised and only a single `[info] validated in x secs` message is logged.

**review**:
@Arelle/arelle

### DIFF
--- a/arelle/plugin/inlineXbrlDocumentSet.py
+++ b/arelle/plugin/inlineXbrlDocumentSet.py
@@ -208,7 +208,7 @@ def inlineXbrlDocumentSetLoader(modelXbrl, normalizedUri, filepath, isEntry=Fals
             elif "ixdsTarget" in kwargs: # passed from validate (multio test cases)
                 _target = kwargs["ixdsTarget"]
             else:
-                _target = modelXbrl.modelManager.formulaOptions.parameterValues["ixdsTarget"][1]
+                _target = modelXbrl.modelManager.formulaOptions.parameterValues[qname("ixdsTarget")][1]
             modelXbrl.ixdsTarget = None if _target == DEFAULT_TARGET else _target or None
         except (KeyError, AttributeError, IndexError, TypeError):
             pass # set later in selectTargetDocument plugin method


aaroncameron-wk (approved):

derekgengenbacher-wf (approved):

derekgengenbacher-wf:

PR #1631, opened 2025-04-23
Adding nl_kvk validation 3.1.3.1

allenbross-wf:
#### Description of change

Added nl_kvk validation 3.1.3.1

#### Steps to Test

Conformance suite will be added when available.
CI.

**review**:
@Arelle/arelle

### DIFF
--- a/arelle/plugin/validate/NL/rules/nl_kvk.py
+++ b/arelle/plugin/validate/NL/rules/nl_kvk.py
@@ -92,3 +92,25 @@ def rule_nl_kvk_3_1_1_2(
                 modelObject = val.modelXbrl
             )
             return
+
+
+@validation(
+    hook=ValidationHook.XBRL_FINALLY,
+)
+def rule_nl_kvk_3_1_3_1 (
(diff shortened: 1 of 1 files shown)


arelle/plugin/validate/NL/rules/nl_kvk.py
```diff
@@ -92,3 +92,25 @@ def rule_nl_kvk_3_1_1_2(
...
+        *args: Any,
+        **kwargs: Any,
+) -> Iterable[Validation]:
+    """
+    NL-KVK.3.1.3.1: xbrli:segment must not be used in contexts.
+    """
+    contexts = val.modelXbrl.contexts.values()
+    for context in contexts:
+        if context.hasSegment:
+            yield Validation.error(
```
austinmatherne-wk, line 112: Perhaps it would be better to raise a single error with references to all offending contexts? If the report was created using segments, it's likely that all of the contexts in the report were.

austinmatherne-wk (approved):

PR #1636, opened 2025-04-25
Update FileSource.py for XULE-referenced resource files redirection f…

freddym56:
…rom xbrl.us web files

#### Reason for change
Supports redirection request from http path (xbrl.us public file to EDGAR repo xule resources location)

#### Steps to Test

**review**:
@Arelle/arelle

### DIFF
--- a/arelle/DisclosureSystem.py
+++ b/arelle/DisclosureSystem.py
@@ -411,11 +411,9 @@ class DisclosureSystem:
         if url in self.mappedFiles:
             return True
         else:  # handle mapped paths
-            mappedUrl = url
             for mapFrom, mapTo in self.mappedPaths:
                 if url.startswith(mapFrom):
                     return True
-                    break
         return False
 
     def uriAuthorityValid(self, uri):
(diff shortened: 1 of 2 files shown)


arelle/DisclosureSystem.py
```diff
@@ -407,6 +407,17 @@ def mappedUrl(self, url):
...
 
+    def isMappedUrl(self, url):
+        if url in self.mappedFiles:
+            return True
+        else:  # handle mapped paths
+            mappedUrl = url
+            for mapFrom, mapTo in self.mappedPaths:
+                if url.startswith(mapFrom):
+                    return True
+                    break
```
austinmatherne-wk, line 418: Assuming this was carried over from copying the `mappedUrl` function, but both the unused `mappedUrl` variable and the unreachable `break` statement can be removed.

freddym56: indeed does should not be there. They have been removed.

austinmatherne-wk (approved):

PR #1657, opened 2025-05-14
Add NL.NL-KVK.3.2.7.1

allenbross-wf:
#### Description of change

Add NL.NL-KVK.3.2.7.1

#### Steps to Test

CI

**review**:
@Arelle/arelle

### DIFF
--- a/arelle/ModelInstanceObject.py
+++ b/arelle/ModelInstanceObject.py
@@ -795,6 +795,15 @@ class ModelInlineFact(ModelInlineValueObject, ModelFact):
             for inline root element, the xbrli:xbrl element is substituted for by the inline root element"""
         return getattr(self, "_ixFactParent") # set by ModelDocument locateFactInTuple for the inline target's root element
 
+    @property
+    def isEscaped(self):
+        """(bool) -- if true, the fact is escaped"""
+        try:
+            return self._isEscaped
+        except AttributeError:
+            self._isEscaped = self.get("escape") in ("true", "1")
+            return self._isEscaped
+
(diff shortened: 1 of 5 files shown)


arelle/ModelInstanceObject.py
```diff
@@ -795,6 +795,15 @@ def parentElement(self):
             for inline root element, the xbrli:xbrl element is substituted for by the inline root element"""
         return getattr(self, "_ixFactParent") # set by ModelDocument locateFactInTuple for the inline target's root element
 
+    @property
+    def isEscaped(self):
+        """(bool) -- if true, the fact is escaped"""
+        try:
+            return self._isEscaped
+        except AttributeError:
+            self._isEscaped = self.get("escape") in ("true", "1")
```
austinmatherne-wk, line 804: I think it would be good to represent the unset state with `None`. The ESEF and NL validation messages appear to need this, but aren't currently validating it `other data types (e.g., xbrli:stringItemType) are assigned @escape="false"`.

brettkail-wk: It is widely believed that ESMA intended PSVI false even though their text implies lexical false, so that shouldn't be changed without more discussion.  I don't know NL well enough to interpret their intent, but it broadly seems wrong to base validations on non-PSVI information, so it might be better to just always use a boolean, even if that means that if a validation for a jurisdiction needs to check lexical information that it has to use `isNil` + `get("escape")` itself.

derekgengenbacher-wf (approved):

derekgengenbacher-wf:

PR #1659, opened 2025-05-15
Validate all loaded models from GUI

austinmatherne-wk:
#### Reason for change
Resolves support email. If a user opens multiple reports in the GUI only the latest is validated. The CLI validates all loaded reports.

#### Description of change
* Validate all opened reports in the GUI.
* Reset the saved disclosure system if the user disables the plugin providing the disclosure system in the config.

#### Steps to Test
* Open multiple reports in the GUI.
* Validate.
* Confirm validated messages are logged for all opened reports.

**review**:
@Arelle/arelle


derekgengenbacher-wf (approved):

derekgengenbacher-wf:

PR #1660, opened 2025-05-15
Add NL.NL-KVK.3.5.2.2.taggedTextFactOnlyInLanguagesOtherThanLanguageOfAReport

derekgengenbacher-wf:
#### Reason for change
Add NL.NL-KVK.3.5.2.2.taggedTextFactOnlyInLanguagesOtherThanLanguageOfAReport

#### Steps to Test
CI 

**review**:
@Arelle/arelle

### DIFF
--- a/arelle/plugin/validate/NL/rules/nl_kvk.py
+++ b/arelle/plugin/validate/NL/rules/nl_kvk.py
@@ -427,3 +427,35 @@ def rule_nl_kvk_3_3_1_3 (
             msg=_('At least one footnote must have the same language as the report\'s language.'),
             modelObject=nonDefLangFtFacts
         )
+
+
+@validation(
+    hook=ValidationHook.XBRL_FINALLY,
+    disclosureSystems=[
+        DISCLOSURE_SYSTEM_NL_INLINE_2024
+    ],
+)
+def rule_nl_kvk_3_5_2_2(
+        pluginData: PluginValidationDataExtension,
+        val: ValidateXbrl,
+        *args: Any,
+        **kwargs: Any,
+) -> Iterable[Validation]:
+    """
+    NL-KVK.3.5.2.2: All tagged text facts MUST be provided in at least the language of the report.
+    """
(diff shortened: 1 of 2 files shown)


austinmatherne-wk:
# Verified

* ci

QA +10
@Workiva/release-management-p

arelle/plugin/validate/NL/rules/nl_kvk.py
```diff
@@ -427,3 +427,35 @@ def rule_nl_kvk_3_3_1_3 (
...
+        if (fact is not None and
+                fact.concept is not None and
+                fact.concept.type is not None and
+                fact.concept.type.isOimTextFactType and
+                fact.xmlLang != reportXmlLang):
+            factsWithWrongLang.add(fact)
+    if len(factsWithWrongLang) > 0:
+        yield Validation.error(
+            codes='NL.NL-KVK.3.5.2.2.taggedTextFactOnlyInLanguagesOtherThanLanguageOfAReport',
+            msg=_('Tagged text facts MUST be provided in the language of the report.'),
```
brettkail-wk, line 459: The docstring implies that it _is_ valid to include a fact in another language so long as you have a corresponding fact in the doc language.  i.e., it seems like we need something like `getAspectEqualFacts(textFacts,includeSingles=True,includeLang=False)`, and then ensure each set contains at least one fact that has the report lang.

Pre-existing, it looks like the context/unit logic in that method is O(n^2), so if you have the same concept used on very many facts with different contexts/units, it's going to run very slowly.  It seems like there should exist some hashable wrapper object for `ModelContext` that implements `__hash__` in terms of `periodHash`/`entityIdentifierHash`/`dimsHash`/`nonDimsHash`/`segmentHash`/`scenarioHash` and `__eq__` in terms of `isEqualTo`.  Is there something like that in formula, xule, etc. to avoid O(n^2)?

derekgengenbacher-wf: XT-4665

PR #1661, opened 2025-05-15
Add NL.NL-KVK.3.5.2.1.undefinedLanguageForTextFact

derekgengenbacher-wf:
#### Reason for change
Add NL.NL-KVK.3.5.2.1.undefinedLanguageForTextFact

#### Steps to Test
CI

**review**:
@Arelle/arelle

### DIFF
--- a/arelle/plugin/validate/NL/rules/nl_kvk.py
+++ b/arelle/plugin/validate/NL/rules/nl_kvk.py
@@ -435,6 +435,37 @@ def rule_nl_kvk_3_3_1_3 (
         DISCLOSURE_SYSTEM_NL_INLINE_2024
     ],
 )
+def rule_nl_kvk_3_5_2_1(
+        pluginData: PluginValidationDataExtension,
+        val: ValidateXbrl,
+        *args: Any,
+        **kwargs: Any,
+) -> Iterable[Validation]:
+    """
+    NL-KVK.3.5.2.1: Each tagged text fact MUST have the ‘xml:lang’ attribute assigned or inherited.
+    """
+    factsWithoutLang = []
+    for fact in val.modelXbrl.facts:
+        if (fact is not None and
+                fact.concept is not None and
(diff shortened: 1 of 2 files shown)


allenbross-wf (dismissed):

arelle/plugin/validate/NL/rules/nl_kvk.py
```diff
@@ -429,6 +429,37 @@ def rule_nl_kvk_3_3_1_3 (
...
+        val: ValidateXbrl,
+        *args: Any,
+        **kwargs: Any,
+) -> Iterable[Validation]:
+    """
+    NL-KVK.3.5.2.1: Each tagged text fact MUST have the ‘xml:lang’ attribute assigned or inherited.
+    """
+    factsWithoutLang = set()
+    for fact in val.modelXbrl.facts:
+        if (fact is not None and
```
brettkail-wk, line 449: Out of curiosity, why `set()` rather than `[]`?  I would not expect `modelXbrl.facts` to have duplicates.

Out of curiosity, can `modelXbrl.facts` actually contain `None`?

derekgengenbacher-wf: I updated to list rather than set. As far as `modelXbrl.facts` containing None, I dont have any examples to show you, but feel better about checking that prior just for safety sake.

allenbross-wf (approved):

arelle/plugin/validate/NL/rules/nl_kvk.py
austinmatherne-wk: I appreciate the need to always guard against None in Arelle. That said, the modelXbrl.facts collection should never contain None values. If it does, we’ll end up throwing exceptions throughout the codebase.

allenbross-wf:
QA +10
All tests passed.

PR #1667, opened 2025-05-19
Rework NL.NL-KVK.3.5.2.2.taggedTextFactOnlyInLanguagesOtherThanLanguageOfAReport

derekgengenbacher-wf:
#### Reason for change
Rework NL.NL-KVK.3.5.2.2.taggedTextFactOnlyInLanguagesOtherThanLanguageOfAReport to handle groups of consistent facts

#### Steps to Test
CI

**review**:
@Arelle/arelle

### DIFF
--- a/arelle/plugin/validate/NL/rules/nl_kvk.py
+++ b/arelle/plugin/validate/NL/rules/nl_kvk.py
@@ -17,7 +17,7 @@ from arelle.typing import TypeGetText
 from arelle.utils.PluginHooks import ValidationHook
 from arelle.utils.validate.Decorator import validation
 from arelle.utils.validate.Validation import Validation
-from arelle.ValidateDuplicateFacts import getHashEquivalentFactGroups
+from arelle.ValidateDuplicateFacts import getHashEquivalentFactGroups, getAspectEqualFacts
 from ..DisclosureSystems import DISCLOSURE_SYSTEM_NL_INLINE_2024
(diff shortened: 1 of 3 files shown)


arelle/plugin/validate/NL/rules/nl_kvk.py
```diff
@@ -505,17 +506,15 @@ def rule_nl_kvk_3_5_2_2(
...
-            codes='NL.NL-KVK.3.5.2.2.taggedTextFactOnlyInLanguagesOtherThanLanguageOfAReport',
-            msg=_('Tagged text facts MUST be provided in the language of the report.'),
-            modelObject=factsWithWrongLang
-        )
+    filtered_facts = [f for f in val.modelXbrl.facts if f.concept is not None and
+                      f.concept.type is not None and
+                      f.concept.type.isOimTextFactType and
+                      f.context is not None]
+    factGroups = getHashEquivalentFactGroups(filtered_facts)
+    for fgroup in factGroups:
```
brettkail-wk, line 514: This code can't use `getHashEquivalentFactGroups` like this.  Hash collisions need to be handled somehow.

Pre-existing, it seems like this API is easy to misuse, so it should probably be hidden.  It seems like it would be best to have a higher-level key object so that some method like this can return precise groups, or at least a method that returns `list[tuple[tuple[concept,context,unit],list[fact]]]`.  `getAspectEqualFacts` is almost that (somewhat confusingly named), but it would need an option to stop considering lang.

aaroncameron-wk (approved):

allenbross-wf (approved):

allenbross-wf:
QA +10

PR #1676, opened 2025-05-21
Update DBA.TM31 to only run on Annual Reports

derekgengenbacher-wf:
#### Reason for change
Update DBA.TM31 to only run on Annual Reports

#### Steps to Test
CI

**review**:
@Arelle/arelle

### DIFF
--- a/arelle/plugin/validate/DBA/rules/tm.py
+++ b/arelle/plugin/validate/DBA/rules/tm.py
@@ -281,7 +281,16 @@ def rule_tm31(
     reportTypeFacts = val.modelXbrl.factsByQname.get(pluginData.informationOnTypeOfSubmittedReportQn, set())
     filteredReportTypeFacts = [f for f in reportTypeFacts if f.xValid >= VALID and f.xValue in pluginData.annualReportTypes]
     if len(filteredReportTypeFacts) > 0:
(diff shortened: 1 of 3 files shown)


arelle/plugin/validate/DBA/rules/tm.py
```diff
@@ -278,7 +278,11 @@ def rule_tm31(
     """
     DBA.TM31: gsd:DateOfApprovalOfReport must only be tagged once if tagged
     """
-    return errorOnMultipleFacts(val.modelXbrl, pluginData.dateOfApprovalOfReportQn, 'DBA.TM31')
+    reportTypeFacts = val.modelXbrl.factsByQname.get(pluginData.informationOnTypeOfSubmittedReportQn, set())
```
aaroncameron-wk, line 281: Since this same logic is going to preclude at least a few validations from running at all, might be nice to write a helper function so we can just do:
```
if not pluginData.isAnnualReport(val.modelXbrl):
	return
```

derekgengenbacher-wf: I have one last PR to make and Ill simplify all the code in that one.

allenbross-wf (approved):

PR #1689, opened 2025-05-28
Add NL.NL-KVK.6.1.1.1.reportPackageMaximumSizeExceeded

derekgengenbacher-wf:
#### Reason for change
Add NL.NL-KVK.6.1.1.1.reportPackageMaximumSizeExceeded

#### Steps to Test
CI

**review**:
@Arelle/arelle

### DIFF
--- a/arelle/plugin/validate/NL/rules/nl_kvk.py
+++ b/arelle/plugin/validate/NL/rules/nl_kvk.py
@@ -4,6 +4,7 @@ See COPYRIGHT.md for copyright information.
 from __future__ import annotations
 
 from datetime import date
+import zipfile
 
 from arelle.ModelInstanceObject import ModelInlineFact
 from arelle.ValidateDuplicateFacts import getDuplicateFactSets
@@ -21,7 +22,8 @@ from arelle.ValidateDuplicateFacts import getHashEquivalentFactGroups, getAspect
 from arelle.utils.validate.ValidationUtil import etreeIterWithDepth
 from ..DisclosureSystems import DISCLOSURE_SYSTEM_NL_INLINE_2024
 from ..PluginValidationDataExtension import (PluginValidationDataExtension, XBRLI_IDENTIFIER_PATTERN,
-                                             XBRLI_IDENTIFIER_SCHEMA, DISALLOWED_IXT_NAMESPACES, ALLOWABLE_LANGUAGES)
+                                             XBRLI_IDENTIFIER_SCHEMA, DISALLOWED_IXT_NAMESPACES, ALLOWABLE_LANGUAGES,
+                                             MAX_REPORT_PACKAGE_SIZE_MBS)
 
 if TYPE_CHECKING:
     from arelle.ModelXbrl import ModelXbrl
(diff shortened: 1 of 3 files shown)


derekgengenbacher-wf:
Intermittently failing test in the DBA conformance suite is fixed in this [PR](https://github.com/Arelle/Arelle/pull/1690).

arelle/plugin/validate/NL/rules/nl_kvk.py
```diff
@@ -818,3 +820,30 @@ def rule_nl_kvk_3_6_3_3(
...
+        **kwargs: Any,
+) -> Iterable[Validation]:
+    """
+    NL-KVK.6.1.1.1: The size of the report package MUST NOT exceed 100 MB.
+    """
+    if val.modelXbrl.fileSource.fs: # must be a zip to be a report package
+        assert isinstance(val.modelXbrl.fileSource.fs, zipfile.ZipFile)
+        maxMB = float(MAX_REPORT_PACKAGE_SIZE_MBS)
+        _size = sum(zi.compress_size for zi in val.modelXbrl.fileSource.fs.infolist())
+        if _size > maxMB * 1048576:
```
brettkail-wk, line 844: Is the limit 100 MB or 100 MiB (~105MB)?  The code is checking the latter.

arelle/plugin/validate/NL/rules/nl_kvk.py
```diff
@@ -818,3 +820,30 @@ def rule_nl_kvk_3_6_3_3(
...
+        pluginData: PluginValidationDataExtension,
+        val: ValidateXbrl,
+        *args: Any,
+        **kwargs: Any,
+) -> Iterable[Validation]:
+    """
+    NL-KVK.6.1.1.1: The size of the report package MUST NOT exceed 100 MB.
+    """
+    if val.modelXbrl.fileSource.fs: # must be a zip to be a report package
+        assert isinstance(val.modelXbrl.fileSource.fs, zipfile.ZipFile)
```
brettkail-wk, line 841: `assert` seems like it will fail badly if it's not a zip, so it seems like the wrong thing to do.  Perhaps `Validation.warning` that the report package size cannot be determined?

arelle/plugin/validate/NL/rules/nl_kvk.py
```diff
@@ -818,3 +820,30 @@ def rule_nl_kvk_3_6_3_3(
...
+        maxMB = float(MAX_REPORT_PACKAGE_SIZE_MBS)
+        _size = sum(zi.compress_size for zi in val.modelXbrl.fileSource.fs.infolist())
```
brettkail-wk, line 843: It seems surprising that we're computing the size of the report package rather than using the size of the input file.

arelle/plugin/validate/NL/rules/nl_kvk.py
derekgengenbacher-wf: Documentation says MB, not MiB. I suppose it would be better to error on the side of smaller rather than the larger amount?

arelle/plugin/validate/NL/rules/nl_kvk.py
derekgengenbacher-wf: After spending some time researching how to best ascertain the size of a compressed zip in python, I can do it the way its already being done, or I can convert the zipfile into some IO file to do some seeking through the file. Doing some quick math on 50 zips included in the Netherlands iXBRL conformance suite, I found that the current way is over 99% accurate on average. That accuracy increases the larger the file is. For instance a 127 KB file was only accurately sized at 126 KB for a 98.2% accuracy. A 11.473 MB file was sized at 11.470 MB for an accuracy of 99.98% and a 163.178 MB files was found to be 163.176 MB for an accuracy of 99.998%. Given the short timeframe in which to complete these validations, the added complexity of converting the zipfile to IO file, the lack of a persuasive gain in accuracy and the increase in accuracy the larger the file gets, I am opting to stay with the current validation as is.

aaroncameron-wk (approved):

aaroncameron-wk:
QA +10

PR #1707, opened 2025-06-03
Add NL-KVK 4.4.3.1 validation

allenbross-wf:
#### Description of change

Add NL-KVK 4.4.3.1 validation

#### Steps to Test

CI

**review**:
@Arelle/arelle


arelle/plugin/validate/NL/rules/nl_kvk.py
```diff
@@ -1120,6 +1123,47 @@ def rule_nl_kvk_4_1_5_2(
...
+        pluginData: PluginValidationDataExtension,
+        val: ValidateXbrl,
+        *args: Any,
+        **kwargs: Any,
+) -> Iterable[Validation]:
+    """
+    NL-KVK.4.4.3.1: The extension taxonomy MUST not modify (prohibit and/or override) default members assigned to dimensions by the KVK taxonomy
+    """
+    for modelLink in cast(list[ModelLink], val.modelXbrl.baseSets[XbrlConst.dimensionDefault, None, None, None]):
+            if pluginData.isExtensionUri(modelLink.modelDocument.uri, val.modelXbrl):
```
aaroncameron-wk: I think you've got an extra indent on this whole block, but doing things like this can prevent a ton of nested indentation:
```suggestion
            if not pluginData.isExtensionUri(modelLink.modelDocument.uri, val.modelXbrl):
                continue
            ...
```

aaroncameron-wk (dismissed):

arelle/plugin/validate/NL/rules/nl_kvk.py
```diff
@@ -1120,6 +1123,47 @@ def rule_nl_kvk_4_1_5_2(
...
+    docsInDTS = pluginData.getDocumentsInDts(val.modelXbrl)
+    for doc in docsInDTS:
+        if doc.type == ModelDocument.Type.LINKBASE:
+            for arcElt in doc.xmlRootElement.iterdescendants(tag="{http://www.xbrl.org/2003/linkbase}definitionArc"):
+                if arcElt.get("use") == "prohibited" and arcElt.get("{http://www.w3.org/1999/xlink}arcrole")  == XbrlConst.dimensionDefault:
+                    yield Validation.error(
+                        codes='NL.NL-KVK.4.4.3.1.extensionTaxonomyOverridesDefaultMembers',
+                        msg=_('A default member is forbidden in the extension taxonomy.  Update the default member to taxonomy defaults.'),
+                        modelObject=val.modelXbrl.modelDocument
+                    )
```
aaroncameron-wk: With my branch merged, I think you can take advantage of `getExtensionData` and `iterArcsByType`, as well as some constants I added to `XbrlConst`. Other `http://www.w3.org/1999/xlink` QNames used above could probably be added as constants as well.

arelle/plugin/validate/NL/rules/nl_kvk.py
```diff
@@ -1322,6 +1325,54 @@ def rule_nl_kvk_4_4_2_3(
...
+                fromLabel = linkChild.get(XbrlConst.qnXlinkFrom.clarkNotation)
+                for fromResource in modelLink.labeledResources[fromLabel]:
+                    if not pluginData.isExtensionUri(fromResource.modelDocument.uri, val.modelXbrl):
+                        yield Validation.error(
+                            codes='NL.NL-KVK.4.4.3.1.extensionTaxonomyOverridesDefaultMembers',
+                             msg=_('A default member does not match the default member settings of the taxonomy. '
+                                   'Update the default member to taxonomy defaults.'
+                                   ),
+                            modelObject=val.modelXbrl.modelDocument
+                        )
```
aaroncameron-wk: If we're only going to pass the document as the ref, then I think we should ensure that we only fire this once per document. Alternatively, we could fire it per occurrence like we are here, but in that case we should pass the offending element.

Below as well.

allenbross-wf: For consistency with the message I change the modelObject to the element.

PR #1720, opened 2025-06-08
OIM Taxonomy Model change updates

hermfischer-wf:
#### Reason for change
 * Model and spec changes since original commit
 * GPU-based vectorized search facility

#### Description of change
Import taxonomy object renaming and refactoring of imported/exported profile and selections.

Add GPU assisted search to match cubes and factspaces by aspect properties
* On Windows/Linux with NVIDIA → CUDA kernels do the work.
* On macOS with Apple Silicon → MPS (Metal Performance Shaders) does the work.
* On other systems → CPU with vectorized math (SIMD) does the work.

saveOIMFactset.py plugin to create factset from inline XBRL (or other) instances.

#### Steps to Test
Example taxonomies

**review**:
@Arelle/arelle


hermfischer-wf:
Rebased and incorporated into #2418

PR #1724, opened 2025-06-09
NL.NL-KVK.4.4.5.1, NL.NL-KVK.4.4.5.2 validations

aaroncameron-wk:
#### Description of change
NL-KVK.4.4.5.1: Custom labels roles SHOULD NOT be used.
NL-KVK.4.4.5.2: Extension taxonomy elements SHOULD be assigned with at most one label for any combination of role and language. (Mixed information on whether or not this should be a warning or error, but the conformance suite and ESEF implementation suggest error)

#### Steps to Test
CI

**review**:
@Arelle/arelle

### DIFF
--- a/arelle/plugin/validate/NL/rules/nl_kvk.py
+++ b/arelle/plugin/validate/NL/rules/nl_kvk.py
@@ -1411,6 +1411,7 @@ def rule_nl_kvk_4_4_5_2(
 ) -> Iterable[Validation]:
     """
     NL-KVK.4.4.5.2: Extension taxonomy elements SHOULD be assigned with at most one label for any combination of role and language.
+    Additionally, extension taxonomies shall not override or replace standard labels of elements referenced in the KVK taxonomy.
     """
     labelsRelationshipSet = val.modelXbrl.relationshipSet(XbrlConst.conceptLabel)
     extensionData = pluginData.getExtensionData(val.modelXbrl)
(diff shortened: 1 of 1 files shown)


arelle/plugin/validate/NL/rules/nl_kvk.py
```diff
@@ -1363,6 +1364,90 @@ def rule_nl_kvk_4_4_3_2(
...
+            if concept in extensionConcepts and len(labels) > 1:
+                yield Validation.error(
+                    codes='NL.NL-KVK.4.4.5.2.taxonomyElementDuplicateLabels',
+                    msg=_('A concept was found with more than one label role for related language. '
+                          'Update to only one combination. Language: %(lang)s, Role: %(labelRole)s, Concept: %(concept)s.'),
+                    modelObject=[concept]+labels, concept=concept.qname, lang=lang, labelRole=labelRole,
+                )
+            elif labelRole == XbrlConst.standardLabel:
+                hasCoreLabel = False
+                hasExtensionLabel = False
```
brettkail-wk, line 1435: This doesn't seem to match the docstring, which seems to suggest that only extension concepts should be checked.  Is there more language in the manual that says this is disallowed, or is this only discoverable from the testsuite?  If the former, perhaps update the docstring, and if the latter, perhaps add a comment.

It seems weird that both base taxonomy and extension taxonomy can define a label.  Why is that allowed? It seems like concepts should only be allowed to have exactly one label per role, and if an extension taxonomy doesn't want the base taxonomy label, it should prohibit it first.

It seems weird that this checking is only done for the standard label and not other [standard-defined roles](https://www.xbrl.org/Specification/XBRL-2.1/REC-2003-12-31/XBRL-2.1-REC-2003-12-31+corrected-errata-2013-02-20.html#_5.2.2.2.2).  Why is that?

aaroncameron-wk: From the manual:
> At least one label defined in the standard label role, i.e. http://www.xbrl.org/2003/role/label,
must be applied for each taxonomy element. Moreover, legal entities shall not override or
replace standard labels (i.e. labels defined in the standard label role) of elements referenced
in the KVK taxonomy. This means that in cases where the standard labels are used, no
standard label for such taxonomy element should be presented in an extension taxonomy
label linkbase.

I've updated the docstring to reflect this "extra" condition. One of the invalid test cases in the suite tests this case explicitly with the description _"Extension taxonomy with core taxonomy elements assigned with 2 Dutch labels using standard label role"_.
I think the next paragraph answers your other questions:

> The above recommendation should not prevent legal entities from defining entity-specific
labels for elements referenced in the KVK taxonomy to better align with the human readable
layer, providing that they are defined in ‘xlink:role’ other than already defined labels in the KVK taxonomy (e.g. verboseLabel).

brettkail-wk: 👍 That's surprising, but ok.  It seems like the logic might be clearer if the `len(labels) > 1` were hoisted.
```
for (lang, labelRole), labels in conceptLangRoleLabels.items():
  if len(labels) > 1:
    if concept in extensionConcepts:
      yield Validation.error
    elif labelRole == XbrlConst.standardLabel:
      ... code as-is
```

PR #1749, opened 2025-06-18
Initial EDINET validation plugin and conformance suite

aaroncameron-wk:
Initial EDINET conformance suite and validation plugin

The EDINET scenario introduced a few new challenges:
- Test suite validation [does not support](https://github.com/Arelle/Arelle/blob/cf9640686a46fe8435974ba8a65b0e09bb90f5a5/arelle/Validate.py#L403) this multi-IXDS scenario.
- Testcase validation [re-uses](https://github.com/Arelle/Arelle/blob/cf9640686a46fe8435974ba8a65b0e09bb90f5a5/arelle/Validate.py#L351-L359) the `FileSource` that loaded the testcase in some cases. This makes using a `FileSource` unreliable for validations.
- The current handling of manifest files for Japan filings [does not support](https://github.com/Arelle/Arelle/blob/cf9640686a46fe8435974ba8a65b0e09bb90f5a5/arelle/plugin/inlineXbrlDocumentSet.py#L175) multiple distinct document sets within a single manifest, which is common for the `AuditDoc` portion of EDINET packages.

#### Steps to Test
CI

**review**:
@Arelle/arelle

### DIFF
--- a/arelle/plugin/validate/EDINET/rules/upload.py
+++ b/arelle/plugin/validate/EDINET/rules/upload.py
@@ -4,8 +4,9 @@ See COPYRIGHT.md for copyright information.
 from __future__ import annotations
 
 import re
+import zipfile
 from collections.abc import Iterable
-from typing import Any
+from typing import Any, cast, IO
 
 from arelle.ValidateXbrl import ValidateXbrl
 from arelle.typing import TypeGetText
@@ -140,10 +141,11 @@ def rule_EC0183E(
     """
     if not pluginData.shouldValidateUpload(val):
         return
-    zipFile = val.modelXbrl.fileSource.fs
-    zipFile.fp.seek(0, 2)  # Move to the end of the file
-    size = zipFile.fp.tell()
-    if size > 55 * 1024 * 1024:
+    zipFile = cast(zipfile.ZipFile, val.modelXbrl.fileSource.fs)
+    file = cast(IO[Any], zipFile.fp)
+    file.seek(0, 2)  # Move to the end of the file
+    size = file.tell()
+    if size > 55 * 1000 * 1000:  # Interpretting MB as megabytes (1,000,000 bytes)
         yield Validation.error(
             codes='EDINET.EC0183E',
             msg=_("The compressed file size exceeds 55MB. "
--- a/arelle/plugin/validate/EDINET/PluginValidationDataExtension.py
+++ b/arelle/plugin/validate/EDINET/PluginValidationDataExtension.py
@@ -10,7 +10,10 @@ from pathlib import Path
 
 from arelle.ModelXbrl import ModelXbrl
 from arelle.ValidateXbrl import ValidateXbrl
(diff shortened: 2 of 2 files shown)


arelle/plugin/validate/EDINET/rules/upload.py
```diff
@@ -0,0 +1,80 @@
...
+        **kwargs: Any,
+) -> Iterable[Validation]:
+    """
+    EDINET.EC0129E: Limit the number of subfolders to 3 or less from the XBRL directory.
+    """
+    startingDirectory = 'XBRL'
+    for path in (Path(file) for file in val.modelXbrl.fileSource.filesDir):
+        parents = [parent.name for parent in path.parents]
+        if startingDirectory in parents:
+            parents = parents[:parents.index(startingDirectory)]
```
brettkail-wk: (I see this is in draft.)  If `filesDir` contains fully-qualified paths, does that mean that `rule_EC0121E` will wrongly produce a warning if you expand the archive in a home directory with invalid characters (e.g., `~/tmp@`)?  It seems like the slicing will also prevent expanding to a directory like `~/XBRL/tmp@/archive/`.  Is there some way to iterate just the entry names rather than trying to derive the entry names from an absolute path?

aaroncameron-wk: I dropped this validation from this initial PR, but yes there are some challenges with the extracted scenario. The approach I've taken with the other validation is to only look at files beneath the directory that the manifest file is located within. I expect that everything within the package above that level is covered fairly strictly by other rules.

arelle/plugin/validate/EDINET/rules/upload.py
```diff
@@ -54,3 +56,96 @@ def rule_EC0121E(
...
+) -> Iterable[Validation]:
+    """
+    EDINET.EC0183E: The compressed file size exceeds 55MB.
+    """
+    if not pluginData.shouldValidateUpload(val):
+        return
+    zipFile = val.modelXbrl.fileSource.fs
+    zipFile.fp.seek(0, 2)  # Move to the end of the file
+    size = zipFile.fp.tell()
+    if size > 55 * 1024 * 1024:
```
brettkail-wk: Can this same `seek` technique be used to improve [this logic](https://github.com/Arelle/Arelle/pull/1689#discussion_r2113786796)?  Perhaps a helper should be added to `FileSource`.

Similar to [here](https://github.com/Arelle/Arelle/pull/1689#discussion_r2113768670), is the limit MB or MiB?  The code is checking the latter.

aaroncameron-wk: Updated to the smaller of the two to be safe, same as NL. We'll consolidate the ESEF/NL/EDINET file size checks with a future PR.

derekgengenbacher-wf (approved):

derekgengenbacher-wf:

PR #1758, opened 2025-06-30
Use macOS 14 for ARM builds

austinmatherne-wk:
#### Reason for change
Workaround for macOS ARM64 builds failing on macOS 15 runner: https://github.com/actions/runner-images/issues/12482

We also use macOS 14 for our test runners so it makes sense to be consistent anyway. Once the above issue is resolved we should switch both the build and test runners to macOS 15.

#### Description of change
Use macOS 14 runner for ARM64 macOS builds.

#### Steps to Test
* [x] Smoke test https://github.com/Arelle/Arelle/actions/runs/15981191746

**review**:
@Arelle/arelle

### DIFF
--- a/.github/workflows/build-macos.yml
+++ b/.github/workflows/build-macos.yml
@@ -13,7 +13,7 @@ on:
         required: false
         type: string
       macos_version:
-        default: '15'
+        default: '14'
         description: 'MacOS version to use (15, 14, 13, latest, etc.)'
         required: false
         type: string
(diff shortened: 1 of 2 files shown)


aaroncameron-wk (approved):

austinmatherne-wk:
The macOS 15 builds are working again thanks to https://github.com/actions/runner-images/issues/12482#issuecomment-3025283970

PR #1762, opened 2025-07-01
FileSource.getSizeBytes

aaroncameron-wk:
#### Reason for change
Consolidate and improve file size validations across ESEF, NL, and EDINET.

#### Steps to Test
CI (added webserver_validate_esef integration test script)

**review**:
@Arelle/arelle

### DIFF
--- a/arelle/plugin/validate/ESEF/ESEF_Current/ValidateXbrlFinally.py
+++ b/arelle/plugin/validate/ESEF/ESEF_Current/ValidateXbrlFinally.py
@@ -111,7 +111,7 @@ def validateXbrlFinally(val: ValidateXbrl, *args: Any, **kwargs: Any) -> None:
                        estimated=_("exact") if _sizeExact else _("estimated"),
                        reportPackageMeasurement=_(val.authParam["reportPackageMeasurement"]) or _("zipped"),
                        size=_size)
-        if _size > maxMB * 1048576:
+        if _size is not None and _size > maxMB * 1048576:
             modelXbrl.error("arelle.ESEF.maximumReportPackageSize",
                             _("The authority %(authority)s requires a report package size under %(maxSize)s MB, size is %(size)s."),
                             modelObject=modelXbrl, authority=val.authority, maxSize=reportPackageMaxMB, size=_size)
--- a/arelle/plugin/validate/ESEF/ESEF_2021/ValidateXbrlFinally.py
+++ b/arelle/plugin/validate/ESEF/ESEF_2021/ValidateXbrlFinally.py
@@ -106,7 +106,7 @@ def validateXbrlFinally(val: ValidateXbrl, *args: Any, **kwargs: Any) -> None:
                        estimated=_("exact") if _sizeExact else _("estimated"),
                        reportPackageMeasurement=_(val.authParam["reportPackageMeasurement"]) or _("zipped"),
                        size=_size)
-        if _size > maxMB * 1048576:
+        if _size is not None and _size > maxMB * 1048576:
             modelXbrl.error("arelle.ESEF.maximumReportPackageSize",
                             _("The authority %(authority)s requires a report package size under %(maxSize)s MB, size is %(size)s."),
                             modelObject=modelXbrl, authority=val.authority, maxSize=reportPackageMaxMB, size=_size)


arelle/FileSource.py
```diff
@@ -655,6 +655,20 @@ def file(
...
+    def getBytesSize(self) -> int | None:
+        """
+        Get the size of the zip file in bytes.
+        :return: Size of the zip file in bytes, or None if not applicable.
+        """
+        if isinstance(self.basefile, str) and os.path.isfile(self.basefile):
+            return os.path.getsize(self.basefile)
+        if isinstance(self.fs, zipfile.ZipFile):
+            zipFile = cast(zipfile.ZipFile, self.fs)
+            stream = cast(IO[Any], zipFile.fp)
```
brettkail-wk: It looks like `fp` is a private field, and it seems like it would be nullable if given a byte stream rather than a file stream.

Can we change `FileSource` creation to always assign the size if available, or can we reasonably assume that all `ZipFile` created from a file will have a `basefile` and we're never in this code path?

aaroncameron-wk: > it seems like it would be nullable if given a byte stream rather than a file stream.

I don't think that's the case. When opening a zip from a stream, it appears to be a `BytesIO` file-like object. I think it's only null if the file/stream has been closed. It doesn't hurt to have a check here anyway.

>  or can we reasonably assume that all ZipFile created from a file will have a basefile and we're never in this code path?

I think this a safe assumption

arelle/FileSource.py
```diff
@@ -655,6 +655,19 @@ def file(
...
             return openXmlFileStream(self.cntlr, filepath, stripDeclaration)
 
+    def getBytesSize(self) -> int | None:
+        """
+        Get the size of the zip file in bytes.
+        :return: Size of the zip file in bytes, or None if not applicable.
+        """
+        if isinstance(self.basefile, str) and os.path.isfile(self.basefile):
+            return os.path.getsize(self.basefile)
+        if isinstance(self.fs, zipfile.ZipFile) and self.fs.fp is not None:
```
brettkail-wk: If we're trying to be conservative, should it be `getattr(self.fs, 'fp')`, or do we want it to fail if the private `fp` attribute is renamed?  In any case, I wonder if it's worth a comment explaining that we're using a private field.

arelle/plugin/validate/ESEF/ESEF_Current/ValidateXbrlFinally.py
```diff
@@ -93,11 +93,24 @@ def validateXbrlFinally(val: ValidateXbrl, *args: Any, **kwargs: Any) -> None:
...
         else:
-            _size = sum(zi.compress_size for zi in modelXbrl.fileSource.fs.infolist())
-            # not usable because zip may be posted or streamed: _size = os.path.getsize(modelXbrl.fileSource.basefile)
+            try:
+                _size = modelXbrl.fileSource.getBytesSize()
+            except Exception:
+                pass
+            if _size is None:
+                _size = sum(zi.compress_size for zi in modelXbrl.fileSource.fs.infolist())
+                _sizeExact = False
```
brettkail-wk, line 108: This estimation logic could possibly be incorporated into a `getEstimatesBytesSize`.

brettkail-wk (approved):

derekgengenbacher-wf (dismissed):

arelle/FileSource.py
```diff
@@ -655,6 +655,30 @@ def file(
...
+            return os.path.getsize(self.basefile)
+        # ZipFile.fp is a private field, but is currently the simplest way for us to
+        # access the internal stream
+        if isinstance(self.fs, zipfile.ZipFile) and (fp := getattr(self.fs, 'fp')) is not None:
+            stream = cast(IO[Any], fp)
+            stream.seek(0, 2)  # Move to the end of the file
+            return stream.tell()  # Report the current position, which is the size of the file
+        return None
+
+    def getBytesSizeEstimate(self) -> int | None:
```
austinmatherne-wk, line 673: This function name should probably indicate that it's for zip file sources.

brettkail-wk: It doesn't seem like the function has to be specific to zip file sources (even if that's currently the only estimate).

austinmatherne-wk: Making it work for other source formats is certainly preferable. I just don't like that callers have to read the docs to realize they need to check their filesource type when calling this.

brettkail-wk: I don't think they do need to do that.  The intent is: call `getBytesSize`, if `None` then exact size is unknown, so call `getBytesSizeEstimate`, if `None` then not even an estimate is known.  Callers shouldn't have to know in what cases the exact or estimated sizes are available.

austinmatherne-wk: That's fair

PR #1779, opened 2025-07-10
EDINET: EC5804E, EC5805E, EC5806E

aaroncameron-wk:
#### Reason for change
EDINET.EC5804E: The preferredFilename attribute must be set on the instance element in the manifest file.

EDINET.EC5805E: The instance file extension is not ".xbrl". File name: xxx. Please change the extension of the instance name set in the preferredFilename attribute value of the instance element in the manifest file to ".xbrl".

EDINET.EC5806E: The same instance file name is set multiple times. File name: xxx. The preferredFilename attribute value of the instance element in the manifest file must be unique within the same file.

#### Steps to Test
CI

**review**:
@Arelle/arelle


arelle/ModelValue.py
```diff
@@ -35,11 +35,11 @@ def qname(value: ModelObject | str | QName, name: str | ModelObject | None = Non
 def qname(value: ModelObject, name: QName, noPrefixIsNoNamespace: bool) -> QName: ...
 
 @overload
-def qname(value: ModelObject | str | QName | Any | None, name: str | ModelObject | dict[str, str] | None) -> QName | None : ...
+def qname(value: ModelObject | str | QName | Any | None, name: str | ModelObject | dict[str, str] | dict[str | None, str] | None) -> QName | None : ...
```
austinmatherne-wk, line 38: `dict[str, str]` is made redundant by `dict[str | None, str]`?

aaroncameron-wk: For whatever reason, the typing engine does not consider `dict[str, str]` a subset of `dict[str | None, str]`.

```
Argument 2 to "qname" has incompatible type "dict[str, str]"; expected "str | ModelObject | dict[str | None, str] | None"
```

austinmatherne-wk (approved):

arelle/ModelValue.py
stevenbronson-wk: Mutable container types are considered invariant in most languages because, for example, `dict[str | None, str]` allows `d[None] = ''` which is incompatible with `dict[str, str]`.  Mypy talks a bit about it [here](https://mypy.readthedocs.io/en/stable/generics.html#variance-of-generic-types).  But even the read-only Mapping is invariant in its key: https://www.github.com/python/typing/issues/445.

PR #1792, opened 2025-07-16
Better handling of test case variations with only warnings

aaroncameron-wk:
#### Reason for change
Test case variations that expected *only* warnings were passing so long as no *errors* were fired. This results in false-positive test case results where an expected warning wasn't actually firing but the test case was passing regardless.

#### Description of change
- Update logic to not "pass" test case before checking expected warnings.
- Update formatting of conformance suite "expected" results output
- Update EDINET conformance suite runner to capture warnings

#### Steps to Test
CI

**review**:
@Arelle/arelle


derekgengenbacher-wf:

PR #1796, opened 2025-07-16
Adding EDINET.EC5700W.GFM.1.2.3

allenbross-wf:
#### Description of change

Adding EDINET.EC5700W.GFM.1.2.3
Moved entityIdentifiersInDocument function to be accessible by different plugins.

#### Steps to Test

CI

**review**:
@Arelle/arelle


arelle/utils/PluginData.py
```diff
@@ -10,3 +13,7 @@ class PluginData:
     Base dataclass which should be extended with fields by plugins that need to cache data while processing a report.
     """
     name: str
+
+    @lru_cache(1)
+    def entityIdentifiersInDocument(self, modelXbrl: ModelXbrl) -> set[tuple[str, str]]:
```
austinmatherne-wk: It feels a little strange to put this on the `PluginData` class when it doesn't use anything from `PluginData`. What about adding this as a method directly on `ModelXbrl`?

aaroncameron-wk: I suggested putting it on the `PluginData` class as a good shared location for plugins, but I guess I don't have a good reason for it not to be on `ModelXbrl`, especially since it's such a generally usable data structure.

PR #1803, opened 2025-07-17
EDINET.EC8033W

aaroncameron-wk:
#### Reason for change
The startDate of a context whose context ID starts with "CurrentYear" is not set to a date earlier than the endDate of a context whose context ID starts with "Prior1Year".

#### Steps to Test
CI

**review**:
@Arelle/arelle


derekgengenbacher-wf (approved):

derekgengenbacher-wf:

arelle/plugin/validate/EDINET/rules/edinet.py
```diff
@@ -59,3 +60,48 @@ def rule_EC5002E(
...
+           and context.isStartEndPeriod
+    ]
+    currentYearContexts = [
+        context
+        for contextId, context in val.modelXbrl.contexts.items()
+        if contextId.startswith('CurrentYear')
+           and context.startDatetime is not None
+           and context.isStartEndPeriod
+    ]
+    for priorYearContext, currentYearContext in itertools.product(priorYearContexts, currentYearContexts):
```
brettkail-wk, line 94: O(n^2) doesn't seem great.  It seems like we can find the minimum current startDate and the maximum current endDate and do a single comparison, and then we could report all current contexts and all prior contexts in one message.

PR #1805, opened 2025-07-18
Add ROS.19 from 2025 guidance

derekgengenbacher-wf:
#### Reason for change
Add ROS.19 from 2025 guidance

#### Steps to Test
CI

**review**:
@Arelle/arelle

### DIFF
--- a/arelle/plugin/validate/ROS/rules/ros.py
+++ b/arelle/plugin/validate/ROS/rules/ros.py
@@ -25,8 +25,7 @@ from arelle.utils.validate.Validation import Validation
 from arelle.ValidateXbrlCalcs import inferredDecimals, rangeValue
 from arelle.XbrlConst import qnXbrliMonetaryItemType, qnXbrliXbrl, xhtml
 from arelle.XmlValidateConst import VALID
-from . import errorOnNegativeFact
-from ..ValidationPluginExtension import IE_PROFIT_LOSS_ORDINARY, IE_PROFIT_LOSS, PRINCIPAL_CURRENCY, TURNOVER_REVENUE
+from ..ValidationPluginExtension import EQUITY, PRINCIPAL_CURRENCY, TURNOVER_REVENUE
 from ..PluginValidationDataExtension import MANDATORY_ELEMENTS,  SCHEMA_PATTERNS, TR_NAMESPACES, PluginValidationDataExtension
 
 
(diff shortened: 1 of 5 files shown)


arelle/plugin/validate/ROS/rules/ros.py
```diff
@@ -293,6 +292,29 @@ def rule_main(
...
+        pluginData: PluginValidationDataExtension,
+        val: ValidateXbrl,
+        *args: Any,
+        **kwargs: Any,
+) -> Iterable[Validation]:
+    """
+    ROS: Rule 19: DPLTurnoverRevenue should be tested to be less than or equal to 10x the absolute value of Equity
+    """
+    equity_fact = next(iter(val.modelXbrl.factsByLocalName.get(EQUITY, set())))
+    turnover_fact = next(iter(val.modelXbrl.factsByLocalName.get(TURNOVER_REVENUE, set())))
```
austinmatherne-wk: These will throw exceptions if there are no facts for these concepts.

If there is more than one, should we validate all of them?

derekgengenbacher-wf: Im not 100% sure. Im trying to get more info from Peter.

derekgengenbacher-wf: Both `DPLTurnoverRevenue` and `Equity` are mandatory concepts. So if either of them dont exist then we will fire that error.

arelle/plugin/validate/ROS/rules/ros.py
```diff
@@ -293,6 +292,42 @@ def rule_main(
...
+        **kwargs: Any,
+) -> Iterable[Validation]:
+    """
+    ROS: Rule 19: DPLTurnoverRevenue should be tested to be less than or equal to 10x the absolute value of Equity
+    """
+    equity_facts = val.modelXbrl.factsByLocalName.get(EQUITY, set())
+    largest_equity_fact = None
+    for e_fact in equity_facts:
+        if e_fact.xValid >= VALID:
+            if largest_equity_fact is None or abs(int(e_fact.value)) > abs(int(largest_equity_fact.value)):
```
austinmatherne-wk, line 311: Should this rule validate against the smallest equity fact?

derekgengenbacher-wf: Peter said to take the largest of both facts and compare them. This rule is a bit arbitrary in that they dont have a specific guideline for how large is too large. Peter selected 10x. So we will see how much noise we get on this.

PR #1806, opened 2025-07-19
Add InstanceInfo plugin to provide statistics on instances

hermfischer-wf:
#### Reason for change
Plugin provides statistics of instance which might affect performance (e.g. inline viewers)

#### Description of change
New feature

#### Steps to Test
Try on random poor performing filings (like N-CSR or REITs)

**review**:
@Arelle/arelle

### DIFF
--- a/arelle/examples/plugin/instanceInfo.py
+++ b/arelle/examples/plugin/instanceInfo.py
@@ -1,0 +1,340 @@
+'''
+instanceInfo.py provides information about an XBRL instance
+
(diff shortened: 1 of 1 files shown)


arelle/examples/plugin/instanceInfo.py
stevenbronson-wk, line 1: Looks like we have a version of this [here](https://github.com/Arelle/Arelle/blob/c05ec1347f525dcf8a28b655248a225784fa5fe2/arelle/archive/plugin/instanceInfo.py).

PR #1808, opened 2025-07-21
Adding EDINET.EC5700W.GFM.1.2.7

allenbross-wf:
#### Description of change

Adding EDINET.EC5700W.GFM.1.2.7

#### Steps to Test

CI

**review**:
@Arelle/arelle


arelle/plugin/validate/EDINET/rules/gfm.py
```diff
@@ -213,6 +214,36 @@ def rule_gfm_1_2_5(
...
+        **kwargs: Any,
+) -> Iterable[Validation]:
+    """
+    EDINET.EC5700W: [GFM 1.2.7] An instance must not contain duplicate xbrli:context elements.
+    """
+    contextHashes: dict[ModelContext, str] = {}
+    duplicateContexts = []
+    for context in val.modelXbrl.contexts.values():
+        contextHash = context.contextDimAwareHash
+        if contextHash in contextHashes and context.isEqualTo(contextHashes[contextHash]):
```
brettkail-wk: What if `contexts.values` contains A,B,A all with the same hash?  It looks like we're going to flip-flop between A->B->A and not diagnose the duplicate A.

It seems like we keep bumping into context hashing complexities.  It seems like we should create something like `ContextHashKey` that implements `__hash__`/`__eq__` in terms of `contextDimAwareHash`/`contextNonDimAwareHash` and `isEqualTo` `dimensionalAspectModel`.

derekgengenbacher-wf (approved):

austinmatherne-wk (approved):

aaroncameron-wk (approved):

PR #1811, opened 2025-07-23
Identify duplicate contexts and units more carefully.

stevenbronson-wk:
#### Description of change
Implement https://github.com/Arelle/Arelle/pull/1808#discussion_r2219717955.
Change some warnings about duplicate contexts/units from reporting duplicate pairs to reporting entire duplicate sets.

#### Steps to Test
CI.  Lightly tested EBA.

**review**:
@Arelle/arelle

### DIFF
--- a/arelle/utils/Contexts.py
+++ b/arelle/utils/Contexts.py
@@ -1,0 +1,41 @@
+from __future__ import annotations
+
+from typing import TYPE_CHECKING, Any
+
+from arelle.utils.Equivalence import (
+    partitionIntoEquivalenceClasses,
+    getDuplicateItemPairs,
+)
+
+if TYPE_CHECKING:
+    from collections.abc import Iterable
+    from arelle.ModelXbrl import ModelXbrl
+    from arelle.ModelInstanceObject import ModelContext
+
+class ContextHashKey:
+    __slots__ = ('context', 'dimensionalAspectModel', 'hash')
+    context: ModelContext
+    dimensionalAspectModel: bool
+    hash: int
+
+    def __init__(self, context: ModelContext, dimensionalAspectModel: bool) -> None:
+        self.context = context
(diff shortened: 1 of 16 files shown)


derekgengenbacher-wf (approved):

arelle/utils/Contexts.py
```diff
@@ -0,0 +1,38 @@
...
+        return NotImplemented
+
+    def __hash__(self) -> int:
+        return self.hash
+
+def partitionContexts(contexts: Iterable[ModelContext], dimensionalAspectModel: bool) -> dict[ContextHashKey, tuple[ModelContext, ...]]:
+    return partitionIntoEquivalenceClasses(contexts, lambda c: ContextHashKey(c, dimensionalAspectModel))
+
+def partitionModelXbrlContexts(modelXbrl: ModelXbrl) -> dict[ContextHashKey, tuple[ModelContext, ...]]:
+    return partitionContexts(modelXbrl.contexts.values(), dimensionalAspectModel=modelXbrl.hasXDT)
```
brettkail-wk, line 35: Hopefully `hasXDT` is accurate in all the code paths that were previously using `contextDimAwareHash` (I'm just not familiar enough with the `hasXDT` field, but presumably it's ok).

derekgengenbacher-wf:

arelle/utils/Contexts.py
stevenbronson-wk: The paths using ModelContext.isEqualTo didn't pass dimensionalAspectModel, so they were getting [hasXDT](https://github.com/Arelle/Arelle/blob/f63a36779e3836beaad972e1db58e9c9b72ce925/arelle/ModelInstanceObject.py#L1271-L1272).  At least now the hash and equality check match, which seems like an improvement.

PR #1818, opened 2025-07-29
Adding EDINET.EC5700W.GFM.1.2.13.

allenbross-wf:
#### Description of change

Adding EDINET.EC5700W.GFM.1.2.13.

#### Steps to Test

CI

**review**:
@Arelle/arelle


arelle/plugin/validate/EDINET/rules/gfm.py
```diff
@@ -289,6 +292,47 @@ def rule_gfm_1_2_10(
...
+)
+def rule_gfm_1_2_13(
+        pluginData: PluginValidationDataExtension,
+        val: ValidateXbrl,
+        *args: Any,
+        **kwargs: Any,
+) -> Iterable[Validation]:
+    """
+    EDINET.EC5700W: [GFM 1.2.13] An instance having a fact with non-nil content and the xml:lang
+    attribute of valuedifferent than default language must also contain a fact using the same element
```
aaroncameron-wk: Nit: typo

arelle/plugin/validate/EDINET/rules/gfm.py
```diff
@@ -289,6 +292,47 @@ def rule_gfm_1_2_10(
...
+    defaultLang = val.disclosureSystem.defaultXmlLang
+    languageFacts: dict[str,dict[QName, set[ModelFact]]] = defaultdict(lambda: defaultdict(set))
+    noMatchingFacts = set()
+    for fact in val.modelXbrl.facts:
+        if fact is not None and fact.xmlLang is not None and fact.xValid >= VALID:
+            languageFacts[fact.xmlLang][fact.qname].add(fact)
+    for language, qnames in languageFacts.items():
+        if language != defaultLang:
+            for qname, facts in qnames.items():
+
```
aaroncameron-wk: Nit: whitespace

arelle/plugin/validate/EDINET/rules/gfm.py
```diff
@@ -289,6 +292,47 @@ def rule_gfm_1_2_10(
...
+    """
+    EDINET.EC5700W: [GFM 1.2.13] An instance having a fact with non-nil content and the xml:lang
+    attribute of valuedifferent than default language must also contain a fact using the same element
+    and all other attributes with an xml:lang attribute that represents the default language.
+    """
+    defaultLang = val.disclosureSystem.defaultXmlLang
+    languageFacts: dict[str,dict[QName, set[ModelFact]]] = defaultdict(lambda: defaultdict(set))
+    noMatchingFacts = set()
+    for fact in val.modelXbrl.facts:
+        if fact is not None and fact.xmlLang is not None and fact.xValid >= VALID:
```
austinmatherne-wk: You don't need to check that fact is not None here. If modelXbrl.facts includes None we will have many other exceptions before this code is reached. Also, this is fine, but it's generally better to check xValid first.

arelle/plugin/validate/EDINET/rules/gfm.py
```diff
@@ -289,6 +292,47 @@ def rule_gfm_1_2_10(
...
+            languageFacts[fact.xmlLang][fact.qname].add(fact)
+    for language, qnames in languageFacts.items():
+        if language != defaultLang:
+            for qname, facts in qnames.items():
+
+                matchingQnames = languageFacts.get(cast(str, defaultLang), {}).get(qname, set())
```
aaroncameron-wk: With `languageFacts` being a `defaultdict`, I think `languageFacts[defaultLang][qname]` is sufficient here.

arelle/plugin/validate/EDINET/rules/gfm.py
```diff
@@ -289,6 +292,47 @@ def rule_gfm_1_2_10(
...
+                for fact in facts:
+                    if all(not fact.context.isEqual(mq.context) for mq in matchingQnames):
```
aaroncameron-wk: Is this supposed to be `isEqualTo`? Or maybe just compare context IDs. I noticed that in the [EDGAR implementation](https://github.com/Arelle/EDGAR/blob/20ae0017887aefeb34db6934be5dc3f894fb4406/validate/Filing.py#L946), they just use the context ID rather than checking value equality, so I expect that approach is fine here too.

aaroncameron-wk: Also, it would appear your test case doesn't ever find values for `matchingQnames`, which is why it doesn't fail on the `isEqual` call. Might be a good idea to introduce a non-Japanese language fact that _does_ have a Japanese counterpart so that this is tested.

austinmatherne-wk: IMO, this reads more cleanly and is easier to understand using `not any(...)`:
```suggestion
                    if not any(fact.context.isEqual(mq.context) for mq in matchingQnames):
```

arelle/plugin/validate/EDINET/rules/gfm.py
```diff
@@ -289,6 +292,47 @@ def rule_gfm_1_2_10(
...
+                        noMatchingFacts.add(fact)
+    if len(noMatchingFacts) > 0:
+        for fact in noMatchingFacts:
```
austinmatherne-wk: The if condition is unnecessary. If noMatchingFacts is empty the iteration won't produce a warning. In the spirit of generators, I'd drop the noMatchingFacts collection entirely and yield where you add to it instead.

arelle/plugin/validate/EDINET/rules/gfm.py
austinmatherne-wk: Nit: cast defaultLang to str where it's defined instead of every time it's used.

arelle/plugin/validate/EDINET/rules/gfm.py
allenbross-wf: I have updated the test to include "en" and "ja" pairs as well as an unpaired "en" fact.   

I do prefer to use the isEqualTo as it seems more explicit as to what it's comparing.

PR #1819, opened 2025-07-29
EDINET.EC8062W

aaroncameron-wk:
#### Reason for change
EDINET.EC8062W: The sum of all liabilities and equity must equal the sum of all assets.

#### Steps to Test
CI

**review**:
@Arelle/arelle


arelle/plugin/validate/EDINET/rules/edinet.py
```diff
@@ -105,3 +107,52 @@ def rule_EC8033W(
...
+        pluginData: PluginValidationDataExtension,
+        val: ValidateXbrl,
+        *args: Any,
+        **kwargs: Any,
+) -> Iterable[Validation]:
+    """
+    EDINET.EC8062W: The sum of all liabilities and equity must equal the sum of all assets.
+    """
+    deduplicatedFacts = pluginData.getDeduplicatedFacts(val.modelXbrl)
+    contextIdPattern = regex.compile(r'^(Prior[1-9]Year|CurrentYear|Prior[1-9]Interim|Interim)(Duration|Instant)$')
```
austinmatherne-wk, line 139: Are the context IDs required to use these particular patterns in EDINET reports?

arelle/plugin/validate/EDINET/rules/edinet.py
```diff
@@ -84,24 +86,86 @@ def rule_EC8033W(
...
+
+    factsByContextId = defaultdict(list)
+    for fact in deduplicatedFacts:
+        if fact.concept.qname not in (pluginData.assetsIfrsQn, pluginData.liabilitiesAndEquityIfrsQn):
```
austinmatherne-wk: You should guard against facts with None concepts (often occurs if the taxonomy can't be loaded/the user didn't provide a required taxonomy package). Alternatively, I believe `fact.qname` (like you use below) is safe because it doesn't lookup the concept from the taxonomy, it just uses the namespace declared in the instance.

arelle/plugin/validate/EDINET/rules/edinet.py
```diff
@@ -84,24 +86,86 @@ def rule_EC8033W(
...
+            continue
+        if not contextIdPattern.match(fact.contextID):
```
austinmatherne-wk: You probably also need to guard against None context IDs.

arelle/plugin/validate/EDINET/rules/edinet.py
```diff
@@ -84,24 +86,86 @@ def rule_EC8033W(
...
+            continue
+        factsByContextId[fact.contextID].append(fact)
+
+    for contextId, facts in factsByContextId.items():
```
austinmatherne-wk: nit: keys aren't used:
```suggestion
    for facts in factsByContextId.values():
```

arelle/plugin/validate/EDINET/rules/edinet.py
aaroncameron-wk: There are sort of some "standard" context IDs, and based on examples only those standard contexts should have this rule applied. Others have only an assets fact without a liabilities fact, for example, which would violate this rule. Hopefully we'll get some clarity around that as we learn more about context rules.

austinmatherne-wk (approved):

arelle/plugin/validate/EDINET/rules/edinet.py
austinmatherne-wk: Encoding meaning and requirements into IDs is both disappointing and not surprising.

allenbross-wf:
QA +10
All tests passed.

arelle/plugin/validate/EDINET/rules/edinet.py
```diff
@@ -84,24 +86,86 @@ def rule_EC8033W(
...
+    deduplicatedFacts = pluginData.getDeduplicatedFacts(val.modelXbrl)
+    contextIdPattern = regex.compile(r'^(Prior[1-9]Year|CurrentYear|Prior[1-9]Interim|Interim)(Duration|Instant)$')
+
+    factsByContextId = defaultdict(list)
+    for fact in deduplicatedFacts:
+        if fact.qname not in (pluginData.assetsIfrsQn, pluginData.liabilitiesAndEquityIfrsQn):
+            continue
+        if fact.contextID is None or not contextIdPattern.match(fact.contextID):
+            continue
+        factsByContextId[fact.contextID].append(fact)
```
brettkail-wk, line 147: It seems conceptually wrong to use `contextID` for a validation like this.  I guess non-duplicate contexts are required anyway(?), but it seems better to use `ContextHashKey` if only to avoid copy/paste mistakes.

arelle/plugin/validate/EDINET/rules/edinet.py
```diff
@@ -84,24 +86,86 @@ def rule_EC8033W(
...
+        for fact in facts:
+            if isinstance(fact.xValue, float):
+                value = Decimal(fact.xValue)
+            else:
+                value = cast(Decimal, fact.xValue)
+            if fact.qname == pluginData.assetsIfrsQn:
+                assetSum += value
+            elif fact.qname == pluginData.liabilitiesAndEquityIfrsQn:
+                liabilitiesAndEquitySum += value
+        if assetSum != liabilitiesAndEquitySum:
```
brettkail-wk, line 161: I wonder if this needs to consider rounding.

arelle/plugin/validate/EDINET/rules/edinet.py
```diff
@@ -84,24 +86,86 @@ def rule_EC8033W(
...
+        if fact.contextID is None or not contextIdPattern.match(fact.contextID):
+            continue
+        factsByContextId[fact.contextID].append(fact)
+
+    for facts in factsByContextId.values():
+        assetSum = Decimal(0)
+        liabilitiesAndEquitySum = Decimal(0)
+        for fact in facts:
+            if isinstance(fact.xValue, float):
+                value = Decimal(fact.xValue)
```
brettkail-wk, line 154: Out of curiosity, how could this ever happen?  It looks like all relevant facts are `monetaryItemType`.

PR #1820, opened 2025-07-29
Fix ROS.19 error around casting

derekgengenbacher-wf:
#### Reason for change
ROS.19 fails to cast to int if the fact value contains a decimal point.

#### Description of change
Cast to float prior to casting to int

#### Steps to Test
CI

**review**:
@Arelle/arelle

### DIFF
--- a/arelle/plugin/validate/ROS/rules/ros.py
+++ b/arelle/plugin/validate/ROS/rules/ros.py
@@ -4,10 +4,11 @@ See COPYRIGHT.md for copyright information.
 from __future__ import annotations
 
 import os
-from typing import Any
+from typing import Any, cast
 
 from collections import Counter
 from collections.abc import Iterable
+from decimal import Decimal
 
 from arelle.typing import TypeGetText
 from arelle.ValidateXbrl import ValidateXbrl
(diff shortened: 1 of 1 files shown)


allenbross-wf (approved):

arelle/plugin/validate/ROS/rules/ros.py
```diff
@@ -300,19 +300,19 @@ def rule_ros19(
     largest_equity_fact = None
     for e_fact in equity_facts:
         if e_fact.xValid >= VALID:
-            if largest_equity_fact is None or abs(int(e_fact.value)) > abs(int(largest_equity_fact.value)):
+            if largest_equity_fact is None or abs(int(float(e_fact.value))) > abs(int(float(largest_equity_fact.value))):
```
austinmatherne-wk: Presumably here and below you should use the parsed xValue or at least a decimal type.

austinmatherne-wk: Here's relevant code Aaron recently wrote: https://github.com/Arelle/Arelle/blob/ba94bcfe295ffb2c8c73d552050e1fcfba9b817b/arelle/plugin/validate/EDINET/rules/edinet.py#L153-L156

PR #1822, opened 2025-07-30
EDINET.EC8024E

aaroncameron-wk:
#### Reason for change
EDINET.EC8024E: Instance values of the same element, context, and unit may not have different values or different decimals attributes. Correct the element with the relevant context ID. For elements in an inline XBRL file that have duplicate elements, context IDs, and unit IDs, set the same values for the value and decimals attribute.

#### Steps to Test
CI

**review**:
@Arelle/arelle


allenbross-wf (approved):

derekgengenbacher-wf:

PR #1823, opened 2025-07-30
EDINET.EC8013W

aaroncameron-wk:
#### Reason for change
EDINET.EC8013W: The context ID of the element associated with the financial statement extended link role begins with one of the following strings:
- Prior?YearDuration
- CurrentYearDuration
- Prior?YearInstant
- CurrentYearInstant
- InterimDuration
- InterimInstant
- Prior?InterimDuration
- Prior?InterimInstant

Where ? is a digit from 1 to 9.

#### Steps to Test
CI

**review**:
@Arelle/arelle


derekgengenbacher-wf (dismissed):

arelle/plugin/validate/EDINET/PluginValidationDataExtension.py
```diff
@@ -45,6 +47,8 @@ def __init__(self, name: str):
         self.assetsIfrsQn = qname(jpigpNamespace, 'AssetsIFRS')
         self.liabilitiesAndEquityIfrsQn = qname(jpigpNamespace, "LiabilitiesAndEquityIFRS")
 
+        self.contextIdPattern = r'(Prior[1-9]Year|CurrentYear|Prior[1-9]Interim|Interim)(Duration|Instant)'
```
austinmatherne-wk: How about storing this as a reusable compiled regex pattern in PluginData, and then using match or fullmatch to control whether it matches just the beginning or the entire ID?”

arelle/plugin/validate/EDINET/rules/contexts.py
```diff
@@ -16,6 +21,77 @@
 _: TypeGetText
 
 
+# Per "Framework Design of EDINET  Taxonomy", ELR definitions contain a 6-digit
```
austinmatherne-wk: nit: extra whitespace

```suggestion
# Per "Framework Design of EDINET Taxonomy", ELR definitions contain a 6-digit
```

arelle/plugin/validate/EDINET/rules/contexts.py
```diff
@@ -16,6 +21,77 @@
...
+        if not pattern.match(fact.contextID):
+            invalidContextIdMap[fact.contextID].append(fact)
+    for contextId, facts in invalidContextIdMap.items():
+        yield Validation.warning(
+            codes='EDINET.EC8013W',
+            msg=_("The context ID (id=%(contextId)s) for the element related to the extended link role "
+                  "in Financial Statement is not per convention. Please set the context ID of the element "
+                  "related to the extended link role of financial statements according to the rules. For "
+                  "the financial statements, please refer to \"3-4-2 Setting the Context\" in the "
+                  "\"Validation Guidelines\"."),
```
austinmatherne-wk, line 88: This rule is unfortunate. From what I can tell, the intent behind the context ID naming convention is to indicate whether a fact appears in the primary financial statements. This distinction helps clarify when a concept is used both in the core financials and elsewhere in the notes.

However, this introduces a validation challenge. Because we can’t reliably infer the fact’s intended placement, we risk raising unnecessary warnings, flagging facts outside the financial statements simply because the same concept is also used within them.

One potential improvement would be to make the warning logic more dynamic. For example, if a concept is only ever used in the financial statements, then a strict context ID requirement makes sense. But if the concept also appears in other roles, the warning could be softened or annotated as a possible false positive.

If you wanted to take this further, you could conditionally suppress the warning by checking whether the fact is dimensionally valid in only either the financial statement roles or non-financial roles. That said, this added complexity and work may not be worth the effort.

aaroncameron-wk: Those do sound like improvements, but I'm wary of trying to improve on the rule without having a more comprehensive understanding of the ruleset. For example, I wouldn't be surprised if another rule warned against using these same concepts outside of financial statements. Since none of the sample filings violate this rule as-implemented, I'm inclined to keep it as-is.

austinmatherne-wk: That's reasonable.

arelle/plugin/validate/EDINET/PluginValidationDataExtension.py
```diff
@@ -37,6 +39,8 @@ class PluginValidationDataExtension(PluginData):
     assetsIfrsQn: QName
     liabilitiesAndEquityIfrsQn: QName
 
+    contextIdPattern: str
```
austinmatherne-wk: Needs to be updated to `regex.Pattern[str]`

PR #1825, opened 2025-07-31
EDINET.EC8054W

aaroncameron-wk:
#### Reason for change
EDINET.EC8054W: For any context with ID containing "NonConsolidatedMember", the scenario element within must be set to "NonConsolidatedMember".

#### Steps to Test
CI

**review**:
@Arelle/arelle


allenbross-wf (approved):

derekgengenbacher-wf (approved):

austinmatherne-wk (approved):

arelle/plugin/validate/EDINET/rules/contexts.py
```diff
@@ -147,3 +147,40 @@ def rule_EC8033W(
...
+) -> Iterable[Validation]:
+    """
+    EDINET.EC8054W: For any context with ID containing "NonConsolidatedMember",
+    the scenario element within must be set to "NonConsolidatedMember".
+    """
+    for context in val.modelXbrl.contexts.values():
+        if pluginData.nonConsolidatedMemberQn.localName not in context.id:
+            continue
+        memberQnames = set()
+        for scenarioElt in context.iterdescendants(XbrlConst.qnXbrlScenario.clarkNotation):
```
brettkail-wk, line 170: It seems like `qnameDims` would be better than `iterdescendants`.

`typedMember` is not necessarily a QName (and is likely not a QName).  I'm not sure this taxonomy uses typed dimensions, but the code seems wrong.

Should the code just look at the `ConsolidatedOrNonConsolidatedAxis` dimension member specifically?

PR #1830, opened 2025-08-01
EDINET.EC5800E (Manifest validations)

aaroncameron-wk:
#### Reason for change
EDINET.EC5800E.FATAL_ERROR_TOC_TREE_NOT_DEFINED:
A table of contents item specified an extended link role that does not exist in the presentation linkbase.
EDINET.EC5800E.ERROR_ELEMENT_NOT_DEFINED_IN_EXTENDED_LINK_ROLE:
A table of contents item specified an extended link role that does not contain the element specified by the parent attribute of the insert element or the start or end attribute of the item element. Check the extended link role and correct it.
EDINET.EC5800E.ERROR_ENDING_ELEMENT_NOT_DEFINED_UNDER_STARTING_ELEMENT:
A table of contents item specified an end element that is not a descendant of the start element within the specified extended link role. Please check and correct the table of contents items.

#### Steps to Test
CI

**review**:
@Arelle/arelle


derekgengenbacher-wf (approved):

arelle/plugin/validate/EDINET/__init__.py
```diff
@@ -63,6 +65,7 @@ def fileSourceEntrypointFiles(filesource: FileSource, inlineOnly: bool, *args: A
                 filesource.select(str(ixbrlFile))
                 entrypoints.append({"file": filesource.url})
             entrypointFiles.append({'ixds': entrypoints})
+    manifestsByFileSource[filesource] = manifests
```
austinmatherne-wk: Won't this cause a memory leak of filesource? Also, we have a lot of other global state in Arelle, but we should be trying to move away from it.

arelle/plugin/validate/EDINET/rules/manifests.py
```diff
@@ -0,0 +1,91 @@
...
+    of the start element within the specified extended link role. Please check
+    and correct the table of contents items.
+    """
+    manifest = pluginData.getManifest(val.modelXbrl)
+    if manifest is None:
+        return
+    errors: list[Validation] = []
+    for tocItem in manifest.tocItems:
+        _validateTocItem(val.modelXbrl, tocItem, errors)
+    yield from errors
```
austinmatherne-wk: Rather than appending to an errors collection, couldn’t _validateTocItem yield each error directly?”

arelle/plugin/validate/EDINET/__init__.py
aaroncameron-wk: Moved to using controller-level plugin data.

derekgengenbacher-wf (approved):

arelle/Validate.py
```diff
@@ -298,6 +298,7 @@ def _validateTestcaseVariation(self, testcase, modelTestcaseVariation):
             self._testcaseValidateInputDTS(testcase, modelTestcaseVariation, errorCaptureLevel, parameters, inputDTSes, baseForElement, resultIsXbrlInstance)
         # update ui thread via modelManager (running in background here)
         self.modelXbrl.modelManager.viewModelObject(self.modelXbrl, modelTestcaseVariation.objectId())
+        self.modelXbrl.modelManager.cntlr._clearPluginData()
```
austinmatherne-wk, line 301: Why does this need to be called here?

aaroncameron-wk: Each test case variation runs within the same controller, so to "simulate" an actual run we should clear the plugin data after each variation.

Side note: this is the cause of quite a bit of other pain with this conformance suite, so I have a plan for a new conformance suite runner based on the Session API that's more true to how Arelle runs.

austinmatherne-wk: That makes sense. It's unfortunate that the test case variation discovery is handled within the validation function.

austinmatherne-wk (approved):

PR #1844, opened 2025-08-11
Validate.FileSource Plugin Hook

aaroncameron-wk:
#### Reason for change
Some jurisdictions define rules that apply to the filing document(s) as a whole and not individual instances. Currently we implement messy workarounds to prevent multi-instance filings from firing these validations multiple times.

#### Description of change
Implement a new endpoint for validating the file source and entry points.

#### Steps to Test
CI

**review**:
@Arelle/arelle

### DIFF
--- a/arelle/utils/PluginHooks.py
+++ b/arelle/utils/PluginHooks.py
@@ -573,6 +573,7 @@ class PluginHooks(ABC):
     def validateFileSource(
             cntlr: Cntlr,
             fileSource: FileSource,
+            entrypoints: list[dict[str, Any]] | None = None,
             *args: Any,
             **kwargs: Any,
     ) -> None:
@@ -592,6 +593,7 @@ class PluginHooks(ABC):
 
         :param cntlr: The [Cntlr](#arelle.Cntlr.Cntlr) instance.
         :param fileSource: The [FileSource](#arelle.FileSource.FileSource) involved in loading the entrypoint files.
+        :param entrypoints: A list of entrypoint configurations.
         :param args: Argument capture to ensure new parameters don't break plugin hook.
         :param kwargs: Argument capture to ensure new named parameters don't break plugin hook.
         :return: None
--- a/arelle/utils/validate/ValidationPlugin.py
+++ b/arelle/utils/validate/ValidationPlugin.py
@@ -105,6 +105,7 @@ class ValidationPlugin:
         self,
         cntlr: Cntlr,
         fileSource: FileSource,
+        entrypoints: list[dict[str, Any]] | None = None,
         *args: Any,
         **kwargs: Any,
     ) -> None:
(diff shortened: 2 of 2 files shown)


arelle/utils/validate/ValidationPlugin.py
```diff
@@ -158,7 +180,7 @@ def validateXbrlDtsDocument(
         :param kwargs: Argument capture to ensure new named parameters don't break plugin hook.
         :return: None
         """
-        self._executeValidations(ValidationHook.XBRL_DTS_DOCUMENT, val, modelDocument, isFilingDocument, *args, **kwargs)
+        self._executeInstanceValidations(ValidationHook.XBRL_DTS_DOCUMENT, val, modelDocument, isFilingDocument, *args, **kwargs)
```
austinmatherne-wk: Nit: Since some of these hooks validate the taxonomy, a more neutral name like `_executeModelValidations` might be clearer.

arelle/utils/PluginHooks.py
```diff
@@ -568,6 +569,35 @@ def modelXbrlLoadComplete(
         """
         raise NotImplementedError
 
+    @staticmethod
+    def validateEntrypointFiles(
+            cntlr: Cntlr,
+            fileSource: FileSource,
+            *args: Any,
+            **kwargs: Any,
+    ) -> None:
```
austinmatherne-wk: The hook is called `Validate.EntrypointFiles`, and while entry points are passed by the controller and validate function, it's not in the signature here, and it isn't used by the plugin. Perhaps `Validate.FileSource` would be a more accurate name?

austinmatherne-wk (dismissed):

arelle/utils/validate/ValidationPlugin.py
```diff
@@ -99,6 +101,26 @@ def modelXbrlLoadComplete(
     ) -> ModelDocument | LoadingException | None:
         raise NotImplementedError
 
+    def validateFileSource(
+        self,
+        cntlr: Cntlr,
+        fileSource: FileSource,
+        *args: Any,
+        **kwargs: Any,
+    ) -> None:
```
austinmatherne-wk: This hook doesn't get called for filings opened from the GUI, does it?

Also, this signature is missing the entry points passed from the CLI (could just not pass them since they're unused).

aaroncameron-wk: Entrypoint detection doesn't run in the GUI either, so at this point EDINET can't really run through the GUI. 

I updated the signature.

austinmatherne-wk: Thanks for the explanation. I created #1848 as a followup.

austinmatherne-wk (approved):

PR #1847, opened 2025-08-12
Set base 64 encoded hrefs as custom log attributes

austinmatherne-wk:
#### Reason for change
Follow up to #1839. The hrefs for the base 64 encoded SVGs are still misleading, as they can't be used from the source document.

#### Description of change
Set the href attribute for embedded SVG to the containing document element that embeds the SVG and instead capture the relative fragment identifiers in a nestedHrefs argument.

#### Steps to Test
* See #1839

Log should be updated to:
```xml
<entry code="ESEF.2.5.1.incorrectMIMETypeSpecified" level="error">
 <message imgType="jpeg" headerType="png" cssSelectors="span.image-f8f2520a75680b9394df35eb3106d62dcbc9ea188879b5a60d87bd62d12a11e4">[ESEF.2.5.1.incorrectMIMETypeSpecified] File type png inferred from file signature does not match the data URL media subtype (MIME subtype) jpeg - p-2025-08-07-en.xhtml 1, 3, 14</message>
 <ref href="p-2025-08-07-en.xhtml#element(/1/1/3)" sourceLine="14">
  <property name="QName" value="style"/>
  <property name="type" value="text/css"/>
 </ref>
 
 <ref href="p-2025-08-07-en.xhtml#element(/1/1/3)" sourceLine="3" nestedHrefs="['#element(/1/1)']">
  <property name="QName" value="image"/>
  <property name="preserveAspectRatio" value="none"/>
  <property name="width" value="100%"/>
  <property name="height" value="100%"/>
  <property name="href" value="data:image/svg+xml;base64,..."/>
 </ref>
 
 <ref href="p-2025-08-07-en.xhtml#element(/1/1/3)" sourceLine="1" nestedHrefs="['#element(/1/1)', '#element(/1/1)']">
  <property name="QName" value="image"/>
  <property name="xlink:href" value="data:image/jpeg;base64,..."/>
 </ref></entry>
```

Instead of:
```xml
<entry code="ESEF.2.5.1.incorrectMIMETypeSpecified" level="error">
 <message imgType="jpeg" headerType="png" cssSelectors="span.image-f8f2520a75680b9394df35eb3106d62dcbc9ea188879b5a60d87bd62d12a11e4">[ESEF.2.5.1.incorrectMIMETypeSpecified] File type png inferred from file signature does not match the data URL media subtype (MIME subtype) jpeg -  1, 3, p-2025-08-07-en.xhtml 14</message>
 <ref href="p-2025-08-07-en.xhtml#element(/1/1/3)" sourceLine="14">
  <property name="QName" value="style"/>
  <property name="type" value="text/css"/>
 </ref>
 
 <ref href="#element(/1/1)" sourceLine="3">
  <property name="QName" value="image"/>
  <property name="preserveAspectRatio" value="none"/>
  <property name="width" value="100%"/>
  <property name="height" value="100%"/>
  <property name="href" value="data:image/svg+xml;base64,..."/>
 </ref>
 
 <ref href="#element(/1/1)" sourceLine="1">
  <property name="QName" value="image"/>
  <property name="xlink:href" value="data:image/jpeg;base64,..."/>
 </ref></entry>
```

**review**:
@Arelle/arelle


arelle/ErrorManager.py
```diff
@@ -204,13 +204,22 @@ def propValues(properties: Any) -> Any:
...
+                            if not hasattr(_arg, 'modelDocument') and _arg.namespaceURI == XbrlConst.svg and len(refs) > 0:
                                 # This is an embedded SVG document without its own file.
-                                ref["href"] = "#" + fragmentIdentifier
+                                # Set the href to the containing document element that defined the encoded SVG.
+                                # and define a nestedHrefs attribute with the fragment identifier.
+                                priorRef = refs[-1]
+                                ref["href"] = priorRef["href"]
+                                priorNestedHrefs = priorRef.get("customAttributes", {}).get("nestedHrefs", [])
+                                ref["customAttributes"] = {
+                                    "nestedHrefs": [*priorNestedHrefs, fragmentIdentifier]
```
brettkail-wk, line 216: Out of curiosity, where does the odd `['#element(/1/1)']` string format come from?  It seems like it would be nice if it were an `xs:list` `xs:anyURI`.

I don't completely follow why `XbrlConst.svg` is treated specially by the error handling code, but I guess it's the only case of an embedded doc with nested content that we have?

brettkail-wk (approved):

aaroncameron-wk (approved):

arelle/ErrorManager.py
austinmatherne-wk: > Out of curiosity, where does the odd ['#element(/1/1)'] string format come from? It seems like it would be nice if it were an xs:list xs:anyURI.

I think it’s based on the [XPointer element() scheme](https://www.w3.org/TR/xptr-element/), but Herm would know for sure. 

> I don't completely follow why XbrlConst.svg is treated specially by the error handling code, but I guess it's the only case of an embedded doc with nested content that we have?

That’s correct. I’m not aware of any other embedded object types. There’s already some handling elsewhere in the logger for objects with missing model documents, but it’s unclear in what all cases that code is triggered. I didn’t have time to dig into it, and I didn’t want to change behavior for those objects without a full understanding.

derekgengenbacher-wf:

PR #1852, opened 2025-08-13
Adding EDINET.EC5700W.GFM.1.2.14

allenbross-wf:
#### Description of change

Adding EDINET.EC5700W.GFM.1.2.14

#### Steps to Test

CI

**review**:
@Arelle/arelle


arelle/plugin/validate/EDINET/rules/gfm.py
```diff
@@ -3,11 +3,13 @@
...
 from __future__ import annotations
 
+import os
 from collections import defaultdict
 from datetime import timedelta
+from lxml.etree import XML, DTD
 from typing import Any, cast, Iterable
 
-import regex
+import regex as re
```
aaroncameron-wk: I could be wrong, but I think the only reason we have the `regex as re` alias elsewhere is because we wanted to swap out the module without having to update every individual reference. I don't know if it makes sense to reintroduce the alias here.

derekgengenbacher-wf (dismissed):

arelle/plugin/validate/EDINET/rules/gfm.py
```diff
@@ -349,6 +352,44 @@ def rule_gfm_1_2_13(
...
+    dtd = DTD(os.path.join(val.modelXbrl.modelManager.cntlr.configDir, xhtmlDtdExtension))
+    htmlBodyTemplate = "<body><div>\n{0}\n</div></body>\n"
+    namedEntityPattern = re.compile("&[_A-Za-z\xC0-\xD6\xD8-\xF6\xF8-\xFF\u0100-\u02FF\u0370-\u037D\u037F-\u1FFF\u200C-\u200D\u2070-\u218F\u2C00-\u2FEF\u3001-\uD7FF\uF900-\uFDCF\uFDF0-\uFFFD]"
+                                    r"[_\-\.:"
+                                    "\xB7A-Za-z0-9\xC0-\xD6\xD8-\xF6\xF8-\xFF\u0100-\u02FF\u0370-\u037D\u037F-\u1FFF\u200C-\u200D\u2070-\u218F\u2C00-\u2FEF\u3001-\uD7FF\uF900-\uFDCF\uFDF0-\uFFFD\u0300-\u036F\u203F-\u2040]*;")
+    XMLpattern = re.compile(r".*(<|&lt;|&#x3C;|&#60;)[A-Za-z_]+[A-Za-z0-9_:]*[^>]*(/>|>|&gt;|/&gt;).*", re.DOTALL)
+
+    for fact in val.modelXbrl.facts:
+        concept = fact.concept
+        if fact.xsiNil != "true" and concept is not None and concept.isTextBlock and XMLpattern.match(fact.value):
```
aaroncameron-wk: I'm guessing this is pre-existing from another implementation, but do you know if there's a reason why `fact.isNil` isn't suitable here? Seems the main practical difference is the value of `1`, which seems odd to make a special case for here.

aaroncameron-wk (approved):

arelle/plugin/validate/EDINET/rules/gfm.py
```diff
@@ -349,6 +352,44 @@ def rule_gfm_1_2_13(
...
+    namedEntityPattern = regex.compile("&[_A-Za-z\xC0-\xD6\xD8-\xF6\xF8-\xFF\u0100-\u02FF\u0370-\u037D\u037F-\u1FFF\u200C-\u200D\u2070-\u218F\u2C00-\u2FEF\u3001-\uD7FF\uF900-\uFDCF\uFDF0-\uFFFD]"
+                                    r"[_\-\.:"
+                                    "\xB7A-Za-z0-9\xC0-\xD6\xD8-\xF6\xF8-\xFF\u0100-\u02FF\u0370-\u037D\u037F-\u1FFF\u200C-\u200D\u2070-\u218F\u2C00-\u2FEF\u3001-\uD7FF\uF900-\uFDCF\uFDF0-\uFFFD\u0300-\u036F\u203F-\u2040]*;")
+    XMLpattern = regex.compile(r".*(<|&lt;|&#x3C;|&#60;)[A-Za-z_]+[A-Za-z0-9_:]*[^>]*(/>|>|&gt;|/&gt;).*", regex.DOTALL)
+
+    for fact in val.modelXbrl.facts:
+        concept = fact.concept
+        if not fact.isNil and concept is not None and concept.isTextBlock and XMLpattern.match(fact.value):
+            for xmlText in [fact.value] + CDATApattern.findall(fact.value):
+                xmlBodyWithoutEntities = htmlBodyTemplate.format(namedEntityPattern.sub("", xmlText).replace('&','&amp;'))
```
brettkail-wk, line 382: What is the purpose of the `namedEntityPattern.sub`?  It looks like this wrongly accepts non-escaped XML `<ix:nonNumeric escape="false">&amp;</ix:nonNumeric>`.

I guess this was possibly copied from `validateTextBlockFacts`?  That looks broken/wrong too since a fact value like `x & <p/>` is wrongly accepted.  More broadly, the `XMLpattern.match` looks wrong for DTR facts, which must always be valid XML (even if they don't match a regex); that regex is probably for loose US GAAP types that didn't require valid XML.

PR #1853, opened 2025-08-13
EDINET.EC5613W

aaroncameron-wk:
#### Reason for change
EDINET.EC5613W: Please set the DEI 'Accounting Standard' value to one of the following: 'Japan GAAP', 'US GAAP', 'IFRS'.

#### Steps to Test
CI

**review**:
@Arelle/arelle


arelle/plugin/validate/EDINET/PluginValidationDataExtension.py
```diff
@@ -233,8 +235,14 @@ def getManifestInstance(self, modelXbrl: ModelXbrl) -> ManifestInstance | None:
...
-        return any(fact.xValid >= VALID and not fact.isNil for fact in requiredFacts)
+        return any(fact is not None for fact in self.iterValidNonNilFacts(modelXbrl, qname))
 
     def isStandardTaxonomyUrl(self, uri: str, modelXbrl: ModelXbrl) -> bool:
         return modelXbrl.modelManager.disclosureSystem.hrefValidForDisclosureSystem(uri)
+
+    def iterValidNonNilFacts(self, modelXbrl: ModelXbrl, qname: QName) -> Iterable[ModelFact]:
+        facts = modelXbrl.factsByQname.get(qname, set())
+        for fact in facts:
+            if fact.xValid < VALID or fact.isNil:
```
allenbross-wf: Wouldn't it be more efficient to yield a fact if the fact.xValid >=VALID and not fact.isNil?

arelle/plugin/validate/EDINET/PluginValidationDataExtension.py
```diff
@@ -233,8 +235,14 @@ def getManifestInstance(self, modelXbrl: ModelXbrl) -> ManifestInstance | None:
         return controllerPluginData.matchManifestInstance(modelXbrl.ixdsDocUrls)
 
     def hasValidNonNilFact(self, modelXbrl: ModelXbrl, qname: QName) -> bool:
-        requiredFacts = modelXbrl.factsByQname.get(qname, set())
-        return any(fact.xValid >= VALID and not fact.isNil for fact in requiredFacts)
+        return any(fact is not None for fact in self.iterValidNonNilFacts(modelXbrl, qname))
```
austinmatherne-wk, line 238: nit: checking if valid facts are not none is unnecessary. 
```python
return any(self.iterValidNonNilFacts(modelXbrl, qname))
```

arelle/plugin/validate/EDINET/rules/edinet.py
```diff
@@ -146,6 +147,35 @@ def rule_EC5002E(
...
+) -> Iterable[Validation]:
+    """
+    EDINET.EC5613W: Please set the DEI "Accounting Standard" value to one
+    of the following: "Japan GAAP", "US GAAP", "IFRS".
+    """
+    errorFacts = []
+    for fact in pluginData.iterValidNonNilFacts(val.modelXbrl, pluginData.accountingStandardsDeiQn):
+        accountingStandard = next((s for s in AccountingStandard if s.value == fact.xValue), None)
+        if accountingStandard is None:
+            errorFacts.append(fact)
```
austinmatherne-wk: nit: O(1) vs O(n) for the generator search:
```python
validAccountingStandards = {s.value for s in AccountingStandard}
errorFacts = [
    fact for fact in pluginData.iterValidNonNilFacts(val.modelXbrl, pluginData.accountingStandardsDeiQn)
    if fact.xValue not in validAccountingStandards
]
```

austinmatherne-wk (approved):

derekgengenbacher-wf (approved):

arelle/plugin/validate/EDINET/PluginValidationDataExtension.py
aaroncameron-wk: I prefer that as well but it does trigger this `FutureWarning`, which we might not care about but I decided to avoid it in this case. I just saw that it's possible to silence/filter these warnings, so I suppose that is an option as well.
```
FutureWarning: Truth-testing of elements was a source of confusion and will always return True in future versions. Use specific 'len(elem)' or 'elem is not None' test instead.
```

austinmatherne-wk (dismissed):

arelle/plugin/validate/EDINET/PluginValidationDataExtension.py
austinmatherne-wk: > Truth-testing of elements was a source of confusion

Indeed. The [\_\_bool\_\_ impl](https://github.com/lxml/lxml/blob/56f0eef494e8a30721c3a7bb2766dee752de16ef/src/lxml/etree.pyx#L1322-L1333) is doing far more than I anticipated. Is not None is the way to go 👍

arelle/plugin/validate/EDINET/PluginValidationDataExtension.py
aaroncameron-wk: Line-count wise, yes. I do lean towards guard statements over nesting to optimize horizontal space but that's not really saving anything in this case.

arelle/plugin/validate/EDINET/PluginValidationDataExtension.py
aaroncameron-wk: I guess just `True for fact ...` would be even better here... I even saw that elsewhere recently but didn't think of it. I'll update on another PR.

PR #1854, opened 2025-08-13
GUI Entrypoint Detection

aaroncameron-wk:
#### Reason for change
The GUI pathway was not running the same entrypoint detection/parsing as the CLI.

#### Description of change
CLI and GUI use same code to parse entrypoints from filename/path/url input.
Implement new GUI-specific plugin hook analogous to `CntlrCmdLine.Filing.Start`.

#### Steps to Test
- [ ] Load non-IXDS filing
- [ ] Load local EDINET conformance suite filing
- [ ] Load EDINET filing from web

**review**:
@Arelle/arelle


derekgengenbacher-wf (approved):

arelle/utils/EntryPointDetection.py
```diff
@@ -1,12 +1,65 @@
...
+        try: # may be a json list
+            _entryPoints = json.loads(_f)
+            _checkIfXmlIsEis = False # json entry objects never specify an xml EIS archive
+        except ValueError as e:
+            # is it malformed json?
+            if _f.startswith("[{") or _f.endswith("]}") or '"file:"' in _f:
+                cntlr.addToLog(_("File name parameter appears to be malformed JSON: {}\n{}").format(e, _f),
+                              messageCode="FileNameFormatError",
+                              level=logging.ERROR)
+                success = False
```
austinmatherne-wk: Looks like this branch was missed in the refactor. Malformed json returns `None, [], True` below because this var isn't used.

arelle/CntlrCmdLine.py
```diff
@@ -979,40 +979,9 @@ def monitorParentProcess():
...
-        _entrypointFiles = _entryPoints
-        if filesource and not filesource.selection and not (sourceZipStream and len(_entrypointFiles) > 0):
-            try:
-                filesourceEntrypointFiles(filesource, _entrypointFiles)
-            except Exception as err:
-                self.addToLog(str(err), messageCode="error", level=logging.ERROR)
-                return False
+        filesource, _entrypointFiles, success = parseEntrypointFileInput(self, options.entrypointFile, sourceZipStream)
+        if not success:
+            return False
```
austinmatherne-wk: There's a distinction in the previous code between `success = False` and exiting early with `return False`. I don't think it matters, but worth considering.

arelle/utils/EntryPointDetection.py
```diff
@@ -34,7 +87,7 @@ def filesourceEntrypointFiles(filesource, entrypointFiles=None, inlineOnly=False
                     entrypointFiles.extend(reportEntries)
                 elif not inlineOnly:
                     entrypointFiles.append({"file": report.fullPathPrimary})
-        else:
+        elif autoSelect:
```
austinmatherne-wk: I would've expected report package entries not to be appended to entrypointFiles a few lines above if autoSelect is False

arelle/CntlrWinMain.py
```diff
@@ -816,15 +819,16 @@ def fileOpenFile(self, filename, importToDTS=False, selectTopView=False):
         if filename:
             for xbrlLoadedMethod in pluginClassMethods("CntlrWinMain.Xbrl.Open"):
                 filename = xbrlLoadedMethod(self, filename) # runs in GUI thread, allows mapping filename, mult return filename
-            filesource = None
+            filesource, entrypointFiles, success = parseEntrypointFileInput(self, filename, autoSelect=False)
```
austinmatherne-wk: Do [the inlineXbrlDocumentSet plugin file menu open dialogs](https://github.com/Arelle/Arelle/blob/b65047cabd0c79cfbd0920e021ec4db0998cebea/arelle/plugin/inlineXbrlDocumentSet.py#L521-L528) still serve a purpose after these changes?

Right now, the standard GUI open dialog only works for non-iXDS files (or EDINET manifests, since both are supported by the same plugin). To open iXDS reports, or any report that relies on a manifest, you have to use the separate file menu entries added by the inlineXbrlDocumentSet plugin (this is obviously undesirable). But that's no longer the case after these changes, you just have to have the inlineXbrlDocumentSet plugin enabled?

arelle/CntlrWinMain.py
```diff
@@ -816,15 +819,16 @@ def fileOpenFile(self, filename, importToDTS=False, selectTopView=False):
...
             # check for archive files
-            filesource = openFileSource(filename, self,
-                                        checkIfXmlIsEis=self.modelManager.disclosureSystem and
-                                        self.modelManager.disclosureSystem.validationType == "EFM")
             if filesource.isArchive:
                 if not filesource.selection and not filesource.isReportPackage: # or filesource.isRss:
                     from arelle import DialogOpenArchive
                     filename = DialogOpenArchive.askArchiveFile(self, filesource)
+                    if filename is not None:
+                        entrypointFiles.append({"file": filename})
```
austinmatherne-wk: Should the user be able to select additional files if parseEntrypointFileInput populated entrypointFiles? If yes, do we need to make sure they haven't selected duplicates from parseEntrypointFileInput?

arelle/utils/EntryPointDetection.py
```diff
@@ -1,12 +1,65 @@
...
+from arelle.UrlUtil import isHttpUrl
+from arelle.typing import TypeGetText
+
+_: TypeGetText
+
+if TYPE_CHECKING:
+    from arelle.Cntlr import Cntlr
+
+
+def parseEntrypointFileInput(cntlr: Cntlr, entrypointFile: str | None, sourceZipStream=None, autoSelect=True) -> tuple[FileSource, list[dict[str, Any]], bool]:
```
austinmatherne-wk: This return type would benefit from a dataclass or named tuple.

arelle/CntlrWinMain.py
aaroncameron-wk: > Should the user be able to select additional files if parseEntrypointFileInput populated entrypointFiles?

I don't think so. Especially for the EDINET use case. I'll add an additional condition above to prevent this dialog if an entrypoint has already been added.

arelle/CntlrWinMain.py
aaroncameron-wk: Those dialogs allow multi-selection of entrypoints, which the standard Open dialogs do not. I'll look into consolidating them for this PR, but may end up saving it for a future PR.

arelle/utils/EntryPointDetection.py
aaroncameron-wk: Fair, `autoSelect=False` is misleading because the logic still allows plugins (and report package detection) to automatically select entrypoints. I'll rename to `fallbackSelect` to better describe that the flag only affects the last-step fallback behavior. I'm also noticing how these changes affect loading of report packages, so will do some additional testing and investigation there.

PR #1858, opened 2025-08-15
Toolbar Icon Update

aaroncameron-wk:
#### Reason for change
Existing icons are a bit dated and don't scale appropriately to modern screen sizes.

#### Description of change
- Increase toolbar height to accommodate 24px toolbar icons.
- Add new set of toolbar icons for both dark and light modes ([Iconoir](https://iconoir.com/), MIT License)
- Delete unused icons
- Known issue: icons do not update based on theme at runtime (which is also the case for other elements of tkinter styling)

#### Steps to Test

Screenshot 2025-08-15 at 9 51 21 AM
Screenshot 2025-08-15 at 9 50 49 AM

**review**:
@Arelle/arelle


requirements.txt
```diff
@@ -1,5 +1,6 @@
 # core modules
 certifi==2025.8.3
+darkdetect==0.8.0
```
austinmatherne-wk, line 3: This needs to be included in the pyproject dependencies for library users.

austinmatherne-wk (commented):
QA will need to include testing frozen builds on all platforms and the GUI from the [python lib GUI script](https://github.com/Arelle/Arelle/blob/9f130513e23ff7db5c717942ecba145b21244d10/pyproject.toml#L78).

PR #1869, opened 2025-08-21
Dependency Cleanup

austinmatherne-wk:
#### Reason for change
Dependency definitions are duplicated across three places:
- pyproject.toml (for library users)
- requirements files (for dev, source, and builds)
- distro.py (for frozen build inclusions)

These had drifted out of sync, for example, some dependencies were marked as core in one place but EDGAR only in another, and vice versa.

#### Description of change
* Align dependency requirements
* Standardize dependency range formatting to match Dependabot’s update style
* Create arelle[XULE] target for library users who need XULE without EDGAR

#### Steps to Test
* CI

**review**:
@Arelle/arelle

### DIFF
--- a/.github/actions/test_conformance_suite/action.yml
+++ b/.github/actions/test_conformance_suite/action.yml
@@ -48,7 +48,7 @@ runs:
       shell: bash
       run: |
         python -m pip install --upgrade pip setuptools wheel
-        pip install -r requirements-test.txt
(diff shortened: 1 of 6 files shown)


derekgengenbacher-wf (approved):

allenbross-wf:
QA +10 
All tests passed.

PR #1871, opened 2025-08-21
Add UKSEF Conformance Suite

austinmatherne-wk:
#### Reason for change
It was reported that valid test cases in the UKSEF conformance suite were triggering validation errors in Arelle. The issue stemmed from Arelle incorrectly mapping unused contexts (with FRC taxonomy concepts) and units to the default (ESEF) target. The iXBRL specification only requires mapping used references to target instances, allowing unused ones to be ignored. While unused contexts and units should not be added to the default target, they still need to be recorded so that validation rules applying to the entire iXBRL document can function properly.

The FRC permits both the final and candidate recommendation (CR) document URIs for report packages, but Arelle currently only recognizes the final version. While this PR does not address that limitation, we will need to decide whether to extend support for CR URIs across all report packages or configure it through a plugin so that only FRC reports are allowed to use them, ensuring other users are warned against using the old URIs.

#### Description of change
- Store unmapped contexts and units in separate collections rather than assigning them to the default target.
- Update EDINET context and unit validations to check unmapped contexts and units (these rules appear intended to capture them, otherwise they would never fire).
- Add the UKSEF conformance suite to CI with expected failures until we implement the FRC XBRL Tagging Guide.

#### Steps to Test
* CI

**review**:
@Arelle/arelle


allenbross-wf (approved):

allenbross-wf:
QA +10
All tests passed.

PR #1874, opened 2025-08-22
Updating logic for edinet-gfm-1.2.14

allenbross-wf:
#### Description of change

Updated logic for edinet-grm-1.2.14

#### Steps to Test

CI

**review**:
@Arelle/arelle

### DIFF
--- a/arelle/plugin/validate/EDINET/PluginValidationDataExtension.py
+++ b/arelle/plugin/validate/EDINET/PluginValidationDataExtension.py
@@ -7,15 +7,17 @@ from collections import defaultdict
 from dataclasses import dataclass
 from decimal import Decimal
 from functools import lru_cache
+from lxml.etree import DTD, XML
 from operator import attrgetter
 from typing import Callable, Hashable, Iterable, cast
 
+import os
 import regex
 
 from arelle.LinkbaseType import LinkbaseType
 from arelle.ModelDocument import Type as ModelDocumentType
 from arelle.ModelDtsObject import ModelConcept
-from arelle.ModelInstanceObject import ModelFact, ModelUnit, ModelContext
+from arelle.ModelInstanceObject import ModelFact, ModelUnit, ModelContext, ModelInlineFact
 from arelle.ModelObject import ModelObject
 from arelle.ModelValue import QName, qname
 from arelle.ModelXbrl import ModelXbrl
@@ -24,7 +26,7 @@ from arelle.ValidateDuplicateFacts import getDeduplicatedFacts, DeduplicationTyp
 from arelle.XmlValidate import VALID
 from arelle.typing import TypeGetText
 from arelle.utils.PluginData import PluginData
-from .Constants import CORPORATE_FORMS, FormType
+from .Constants import CORPORATE_FORMS, FormType, xhtmlDtdExtension
 from .ControllerPluginData import ControllerPluginData
 from .ManifestInstance import ManifestInstance
 from .Statement import Statement, STATEMENTS, BalanceSheet, StatementInstance, StatementType
@@ -165,6 +167,22 @@ class PluginValidationDataExtension(PluginData):
                 )
             )
         return balanceSheets
+
+    def getProblematicTextBlocks(self, modelXbrl: ModelXbrl) -> list[ModelInlineFact]:
+        problematicTextBlocks: list[ModelInlineFact] = []
+        dtd = DTD(os.path.join(modelXbrl.modelManager.cntlr.configDir, xhtmlDtdExtension))
+        htmlBodyTemplate = "<body><div>\n{0}\n</div></body>\n"
+        for fact in modelXbrl.facts:
+            concept = fact.concept
+            if isinstance(fact, ModelInlineFact) and not fact.isNil and concept is not None and concept.isTextBlock and not fact.isEscaped:
+                xmlBody = htmlBodyTemplate.format(fact.value)
+                try:
+                    textblockXml = XML(xmlBody)
+                    if not dtd.validate(textblockXml):
+                        problematicTextBlocks.append(fact)
+                except Exception:
+                    problematicTextBlocks.append(fact)
+        return problematicTextBlocks
 
     @lru_cache(1)
     def getStatementInstance(self, modelXbrl: ModelXbrl, statement: Statement) -> StatementInstance | None:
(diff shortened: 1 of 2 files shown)


arelle/plugin/validate/EDINET/PluginValidationDataExtension.py
```diff
@@ -166,6 +168,22 @@ def getBalanceSheets(self, modelXbrl: ModelXbrl, statement: Statement) -> list[B
             )
         return balanceSheets
 
+    def getProblematicTexBlocks(self, modelXbrl: ModelXbrl) -> list[ModelFact]:
```
aaroncameron-wk: nit: Typo

arelle/plugin/validate/EDINET/rules/gfm.py
```diff
@@ -368,28 +368,16 @@ def rule_gfm_1_2_14(
...
-                    yield Validation.warning(
-                        codes='EDINET.EC5700W.GFM.1.2.14',
-                        msg=_('The content of an element with a data type of nonnum:textBlockItemType is not well-formed XML (a format that conforms to XML grammar, '
-                              'such as all start and end tags being in pairs, and the end tag of a nested tag not coming after the end tag of its parent tag). '
-                              'Correct the content so that it is well-formed.'),
-                        modelObject = fact
-                    )
+    problematicFacts = pluginData.getProblematicTexBlocks(val.modelXbrl)
+    if len(problematicFacts) > 0:
+        for fact in problematicFacts:
```
aaroncameron-wk: Maybe just one warning with `modelObject=problematicFacts`?

arelle/plugin/validate/EDINET/PluginValidationDataExtension.py
```diff
@@ -166,6 +168,22 @@ def getBalanceSheets(self, modelXbrl: ModelXbrl, statement: Statement) -> list[B
...
 
+    def getProblematicTextBlocks(self, modelXbrl: ModelXbrl) -> list[ModelFact]:
+        problematicTexBlocks: list[ModelFact] = []
+        CDATApattern = regex.compile(r"<!\[CDATA\[(.+)\]\]")
+        dtd = DTD(os.path.join(modelXbrl.modelManager.cntlr.configDir, xhtmlDtdExtension))
+        htmlBodyTemplate = "<body><div>\n{0}\n</div></body>\n"
+        XMLpattern = regex.compile(r".*(<|&lt;|&#x3C;|&#60;)[A-Za-z_]+[A-Za-z0-9_:]*[^>]*(/>|>|&gt;|/&gt;).*", regex.DOTALL)
+        for fact in modelXbrl.facts:
+            concept = fact.concept
+            if not fact.isNil and concept is not None and concept.isTextBlock and XMLpattern.match(fact.value):
```
brettkail-wk: This `XMLpattern` check looks wrong.  I believe `<ix:nonNumeric>&lt;</ix:nonNumeric>` should cause a validation, but I believe this code is wrongly accepting it.  It should fail with an error because the fact value (`<`) is not valid XML, but the regex is wrongly skipping validation because it doesn't look like XML.

Note this relies on the `escape="false"` default, which is unusual/unexpected but should otherwise be valid iXBRL.  If there's some other EDINET requirement for `escape="true"` (as added to ESEF in the last couple years), then that validation would effectively supersede this one (with `escape="true"`, it's impossible to get malformed XML), but in general, I think this `XMLpattern` check should be removed.

arelle/plugin/validate/EDINET/PluginValidationDataExtension.py
```diff
@@ -166,6 +168,22 @@ def getBalanceSheets(self, modelXbrl: ModelXbrl, statement: Statement) -> list[B
...
+                for xmlText in [fact.value] + CDATApattern.findall(fact.value):
+                    xmlBody = htmlBodyTemplate.format(xmlText.replace('&','&amp;'))
```
brettkail-wk: This `replace` looks wrong.  I believe `<ix:nonNumeric escape="false">&amp;</ix:nonNumeric>` should cause a validation, but I believe this code is wrongly accepting it.

Note this simple test string won't work until the `XMLpattern` issue is fixed.

arelle/plugin/validate/EDINET/PluginValidationDataExtension.py
```diff
@@ -166,6 +168,25 @@ def getBalanceSheets(self, modelXbrl: ModelXbrl, statement: Statement) -> list[B
...
 
+    def getProblematicTextBlocks(self, modelXbrl: ModelXbrl) -> list[ModelInlineFact]:
+        problematicTextBlocks: list[ModelInlineFact] = []
+        CDATApattern = regex.compile(r"<!\[CDATA\[(.+)\]\]")
+        dtd = DTD(os.path.join(modelXbrl.modelManager.cntlr.configDir, xhtmlDtdExtension))
+        htmlBodyTemplate = "<body><div>\n{0}\n</div></body>\n"
+        for fact in modelXbrl.facts:
+            concept = fact.concept
+            if isinstance(fact, ModelInlineFact) and not fact.isNil and concept is not None and concept.isTextBlock and not fact.isEscaped:
+                for xmlText in [fact.value] + CDATApattern.findall(fact.value):
```
brettkail-wk: I don't understand this CDATApattern logic (pre-existing from [this code](https://github.com/Arelle/Arelle/blob/34b058b6edd316874ebc2ede5a69a73793959006/arelle/ValidateFilingText.py#L19)).  It appears to be extracting embedded CDATA and then trying to parse escaped text as if it's XML when it should be treated as literal text.  I think `<ix:nonNumeric>&lt;![CDATA[&lt;]]&gt;</ix:nonNumeric>` is wrongly rejected by this code when it should be accepted.

(This loop is wrongly reassigning `xmlBody`, so I think `<ix:nonNumeric>&amp; &lt;[CDATA[]]&gt;</ix:nonNumeric>` is wrongly accepted by this code when it should rejected.)

I think `CDATApattern` and the loop should be removed altogether, and the fact value should just be processed directly:
```
if isinstance(fact, ModelInlineFact) and not fact.isNil and concept is not None and concept.isTextBlock and not fact.isEscaped:
  xmlBody = htmlBodyTemplate.format(fact.value)
  ...
```

arelle/plugin/validate/EDINET/PluginValidationDataExtension.py
```diff
@@ -166,6 +168,25 @@ def getBalanceSheets(self, modelXbrl: ModelXbrl, statement: Statement) -> list[B
...
+                    xmlBody = htmlBodyTemplate.format(xmlText)
+                try:
+                    textblockXml = XML(xmlBody)
+                    if not dtd.validate(textblockXml):
+                        problematicTextBlocks.append(fact)
+                except Exception as e:
+                    if isinstance(e, XMLSyntaxError):
```
brettkail-wk: Why this `if isinstance`?  Somehow, we need to avoid ignoring other exceptions, perhaps just removing this `if isinstance`.

brettkail-wk (approved):

austinmatherne-wk (approved):

PR #1881, opened 2025-08-26
EDINET.EC0192E

aaroncameron-wk:
#### Reason for change
EDINET.EC0192E: The cover file for PrivateDoc cannot be set because it uses a PublicDoc cover file.

#### Description of change
Validate that the root PrivateDoc directory does not contain a cover page document.

#### Steps to Test
CI

**review**:
@Arelle/arelle


arelle/plugin/validate/EDINET/rules/upload.py
```diff
@@ -341,6 +341,34 @@ def rule_EC0188E(
...
+    """
+    EDINET.EC0192E: The cover file for PrivateDoc cannot be set because it uses a
+    PublicDoc cover file. Please delete the cover file from PrivateDoc and upload
+    it again.
+    """
+    uploadContents = pluginData.getUploadContents(fileSource)
+    for coverPagePath in uploadContents.coverPagePaths:
+        if len(coverPagePath.parts) != 2 or coverPagePath.parts[0] != InstanceType.PRIVATE_DOC.value:
+            continue
+        yield Validation.error(
```
austinmatherne-wk, line 364: This is purely stylistic, so feel free to disagree, there’s no right answer. Guard clauses are great for multiple checks, distinct failure paths, or to avoid deep nesting, but for a simple case like this I find the positive conditional reads a bit more cleanly.

austinmatherne-wk (approved):

arelle/plugin/validate/EDINET/rules/upload.py
aaroncameron-wk: Fair, I'll clean up in my next PR which will have a merge conflict here anyway.

PR #1882, opened 2025-08-26
EDINET.EC0200E

aaroncameron-wk:
#### Reason for change
EDINET.EC0200E: Report document filenames should not use characters other than those allowed (alphanumeric characters, '-' and '_').

#### Description of change
This is very similar to EC0121E, so they've been combined into one validation function.
Also, reworked how we categorize upload filepaths.

#### Steps to Test
CI

**review**:
@Arelle/arelle


tests/resources/conformance_suites/edinet/EC0237E/index.xml
```diff
@@ -17,6 +17,7 @@
             <instance readMeFirst="true">invalid01.zip</instance>
         </data>
         <result>
+            <error>EDINET.EC0121E</error> <!-- Also exceeds 31 character filename limit. -->
```
aaroncameron-wk, line 20: EC0121E was previously missing files within subdirectories within report folders.

austinmatherne-wk (approved):

PR #1892, opened 2025-09-04
EDINET: EC1013E, EC1014E

aaroncameron-wk:
#### Reason for change
EDINET.EC1013E: The URI in the HTML specifies a path not under a subdirectory.
EDINET.EC1014E: The URI in the HTML specifies a path to a file that doesn't exist.

#### Description of change
Consolidate validations involving HTML URIs into single function.

#### Steps to Test
CI

**review**:
@Arelle/arelle

### DIFF
--- a/arelle/plugin/validate/EDINET/rules/upload.py
+++ b/arelle/plugin/validate/EDINET/rules/upload.py
@@ -635,7 +635,7 @@ def rule_EC1006E(
     hook=ValidationHook.XBRL_FINALLY,
     disclosureSystems=[DISCLOSURE_SYSTEM_EDINET],
 )
-def rule_EC1007E(
+def rule_html_uris(
         pluginData: PluginValidationDataExtension,
         val: ValidateXbrl,
         *args: Any,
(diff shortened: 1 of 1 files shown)


arelle/plugin/validate/EDINET/rules/upload.py
```diff
@@ -656,6 +658,31 @@ def rule_EC1007E(
                     line=elt.sourceline,
                     modelObject=elt,
                 )
+                continue
+            path = Path(value)
+            if len(path.parts) < 2:
```
brettkail-wk, line 663: Is this the actual intent?  I would have guessed it  "don't link to thinks outside the filing".  For example, given `filingDir/subdir/file.htm`, I would have thought `other.htm` is ok (reference file in same directory), `../top.htm` is ok (referencing a file in a parent dir but still within the file), `../subdir2/other2.htm` is ok, but `../../outside.htm` and `/outside.htm` are not ok.

(If it is actually the intent, I guess `./other.htm` is accepted but shouldn't be?)

aaroncameron-wk: Yes, they want any files referenced by the HTML to be in a subfolder of the same directory. Other rules, somewhat redundantly, prohibit referencing files in top-level/parent directories (EC1015E, EC1035E). In hindsight I could have included those implementations in this PR, but they will be included in this same function at some point.

`image.jpg` and `./image.jpg` have the same `paths` value of `('image.jpg',)`. Both those cases are included in the testcase filing.

derekgengenbacher-wf (approved):

PR #1893, opened 2025-09-05
Validation.Complete plugin hook

aaroncameron-wk:
#### Reason for change
EDINET.EC1017E: There is an unused file.

This rule highlighted the need for a final controller-level validation plugin hook.
This is valuable for any validation that isn't instance-specific, but requires information that is only available after models have been parsed and/or validated.

#### Steps to Test
CI

**review**:
@Arelle/arelle


arelle/plugin/validate/EDINET/rules/upload.py
```diff
@@ -646,13 +646,21 @@ def rule_html_uris(
...
     EDINET.EC1014E: The URI in the HTML specifies a path to a file that doesn't exist.
     """
-    for doc in val.modelXbrl.urlDocs.values():
+    if not isinstance(val.modelXbrl.fileSource.basefile, str):
+        return
+    basePath = Path(val.modelXbrl.fileSource.basefile)
+    for uri, doc in val.modelXbrl.urlDocs.items():
+        docPath = Path(uri)
+        if docPath.is_relative_to(basePath):
+            pluginData.addUsedFilepath(val.modelXbrl, docPath.relative_to(basePath))
```
brettkail-wk: Out of curiosity, why this `addUsedFilepath`?  I would have expected the `addUsedFilepath` in the loop below to be the critical one for considering "used".  It seems like this logic is making ~all files be considered "used", or am I misreading?

arelle/plugin/validate/EDINET/rules/upload.py
```diff
@@ -676,13 +684,15 @@ def rule_html_uris(
...
-                          "File name: %(file)s (line %(line)s). "
+                          "File name: '%(file)s' (line %(line)s). "
                           "Please update the URI to reference a file."),
                     file=doc.basename,
                     line=elt.sourceline,
                     modelObject=elt,
                 )
                 continue
+            fileSourcePath = fullPath.relative_to(basePath)
+            pluginData.addUsedFilepath(val.modelXbrl, fileSourcePath)
```
brettkail-wk: It seems somewhat fragile for the other issue's logic to depend on this rule running.  I wonder if the processing of the HTML URIs should be stored in plugin data so this issue and the other can work off the same?

arelle/plugin/validate/EDINET/rules/upload.py
aaroncameron-wk: This first call is only adding files that are a part of the XBRL instance and taxonomy (or at least thats my understanding of what `urlDocs` returns), the second call adds files that are referenced by HTML URIs. I suspect the first call may not be necessary for EC1017E since that rule only concerns files in subdirectories, and I'm pretty sure XBRL content is not allowed in subdirectories (although I'm not sure that's explicitly stated).

Either way, the thought was to have `getUsedFilepaths` be more general use.

arelle/plugin/validate/EDINET/rules/upload.py
aaroncameron-wk: Agreed on it not being ideal for one rule to rely on another. In this case, the other rule is running in a different, controller-level context, which makes it different from common situations where a shared/cached `PluginData` function can serve both validations. I've added `PluginData._initialize` so we can ensure that the needed data is generated for both the instance- and controller-level validations regardless of which validation rules run.

arelle/plugin/validate/EDINET/PluginValidationDataExtension.py
```diff
@@ -94,6 +109,29 @@ def _contextMatchesStatement(self, modelXbrl: ModelXbrl, contextId: str, stateme
...
+        if not isinstance(modelXbrl.fileSource.basefile, str):
+            return
+        controllerPluginData = ControllerPluginData.get(modelXbrl.modelManager.cntlr, self.name)
+        basePath = Path(modelXbrl.fileSource.basefile)
+        for uri, doc in modelXbrl.urlDocs.items():
+            docPath = Path(uri)
+            if docPath.is_relative_to(basePath):
+                controllerPluginData.addUsedFilepath(docPath.relative_to(basePath))
+            else:
+                continue
```
austinmatherne-wk: nit: else condition reads a bit strange.
```suggestion
            if not docPath.is_relative_to(basePath):
                continue
            controllerPluginData.addUsedFilepath(docPath.relative_to(basePath))
```

austinmatherne-wk (approved):

PR #1900, opened 2025-09-09
EDINET: EC1021E, EC1023E

aaroncameron-wk:
#### Description of change
EDINET.EC1021E: The URI in the HTML specifies a path to a file that doesn't exist.
EDINET.EC1023E: The URI in the HTML specifies a path to a PDF file.

Also, correct EDINET.EC1014E as it seems the original interpretation was actually that of EC1021E and EC1014E was supposed to check if the URI referenced a _directory_ rather than a file.

#### Steps to Test
CI

**review**:
@Arelle/arelle

### DIFF
--- a/arelle/plugin/validate/EDINET/rules/upload.py
+++ b/arelle/plugin/validate/EDINET/rules/upload.py
@@ -715,7 +715,6 @@ def rule_uri_references(
                 modelObject=uriReference.element,
             )
             continue
-        fullPath = Path(uriReference.document.uri).parent / path
         if not val.modelXbrl.fileSource.exists(str(fullPath)):
             yield Validation.error(
                 codes='EDINET.EC1021E',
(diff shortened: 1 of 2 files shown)


arelle/plugin/validate/EDINET/rules/upload.py
```diff
@@ -670,11 +670,10 @@ def rule_uri_references(
                 modelObject=uriReference.element,
             )
             continue
-        fullPath = Path(uriReference.document.uri).parent / path
-        if not val.modelXbrl.fileSource.exists(str(fullPath)):
+        if not path.suffix:
```
austinmatherne-wk: I'm not sure this is quite right. A file doesn't have to include a suffix. Does is_dir() with a zipfile.Path work here?

aaroncameron-wk: In the context of an EDINET filing, a file without a suffix would fire suffix validation errors, but I'll check out your suggestion.

PR #1904, opened 2025-09-11
Add GFM.1.5.7 and GFM.1.5.8

derekgengenbacher-wf:
#### Reason for change
- Added GFM.1.5.7
- Added GFM.1.5.8

#### Steps to Test
- CI

**review**:
@Arelle/arelle


tests/resources/conformance_suites/edinet/EC5700W.GFM.1.5.7/index.xml
```diff
@@ -0,0 +1,35 @@
...
+<?xml-stylesheet type="text/xsl" href="../testcase.xsl"?>
+<testcase
+        xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+        xmlns="http://xbrl.org/2005/conformance"
+        name="EDINET.EC5700W.GFM.1.5.7"
+        description="EDINET.EC5700W.GFM.1.5.7"
+        outpath=''
+        owner="support@arelle.org"
+        xsi:schemaLocation="http://xbrl.org/2005/conformance https://www.xbrl.org/2005/conformance.xsd">
+    <variation id="invalid01" name="invalid01">
```
aaroncameron-wk, line 11: To save on runtime, both cases could be introduced into a single variation with `<warning num="2">EDINET.EC5700W.GFM.1.5.7</warning>`, though I'm fine either way.

arelle/plugin/validate/EDINET/rules/gfm.py
```diff
@@ -701,3 +704,75 @@ def rule_gfm_1_5_6(
...
+                if DISALLOWED_LABEL_CHARACTERS.findall(label.viewText()):
+                    yield Validation.warning(
+                        codes='EDINET.EC5700W.GFM.1.5.7',
+                        msg=_("The concept of '%(concept)s' has a label classified as '%(role)s that contains the '<' character: %(label)s"),
+                        concept=concept.qname,
+                        role=label.role,
+                        label=label.viewText(),
+                        modelObject=label
+                    )
+                elif DISALLOWED_LABEL_WHITE_SPACE_CHARACTERS.findall(label.viewText()):
```
austinmatherne-wk: nit: Here (and in the other patterns), `pattern.search(...)` is a better fit than `pattern.findall(...)`. It’s slightly faster and more clearly communicates the intent, since you’re only checking for the existence of a match rather than using the results.

arelle/plugin/validate/EDINET/rules/gfm.py
```diff
@@ -701,3 +704,75 @@ def rule_gfm_1_5_6(
...
+    EDINET.EC5700W: [GFM 1.5.8] A label should not begin or end with a white space character
+    """
+    labelRelationshipSet = val.modelXbrl.relationshipSet(XbrlConst.conceptLabel)
+    if labelRelationshipSet is None:
+        return
+    for concept in val.modelXbrl.qnameConcepts.values():
+        labelRels = labelRelationshipSet.fromModelObject(concept)
+        for rel in labelRels:
+            label = rel.toModelObject
+            if label.textValue is not None and BEGINNING_AND_ENDING_WHITE_SPACE.findall(label.textValue):
```
austinmatherne-wk: nit: comparing to strip is faster and more pythonic IMO:
```suggestion
            if label.textValue is not None and label.textValue != label.textValue.strip():
```

arelle/plugin/validate/EDINET/rules/gfm.py
```diff
@@ -701,3 +703,75 @@ def rule_gfm_1_5_6(
...
+    """
+    labelRelationshipSet = val.modelXbrl.relationshipSet(XbrlConst.conceptLabel)
+    if labelRelationshipSet is None:
+        return
+    for concept in val.modelXbrl.qnameConcepts.values():
+        labelRels = labelRelationshipSet.fromModelObject(concept)
+        for rel in labelRels:
+            label = rel.toModelObject
+            if label.role != XbrlConst.documentationLabel and label.viewText() is not None:
+                if DISALLOWED_LABEL_CHARACTERS.search(label.viewText()):
```
brettkail-wk: `viewText()` appears to return the extracted text, not the raw XML, so I don't understand this regex.  If the rule means "the label should not contain XML elements", then this needs to somehow look at the XML childrne, not just the extracted text.  If the rule really means "the label should not contain `<`" (seems like a weird rule but maybe?), then `if "<" in label.viewText():` should be sufficient.

derekgengenbacher-wf: The rule states: The text of `link:label` must not have any consecutive whitespace characters and no
occurrences of `<` unless its `xlink:role` attribute is
`http://www.xbrl.org/2003/label/documentation`. This rule includes the prohibition of alternative serializations of the `<` character such as `&lt;`, `&#x3C;`, `&#60;`.

brettkail-wk: > The rule states: The text of `link:label` must not have any consecutive whitespace characters and no
occurrences of `<` unless its `xlink:role` attribute is
`http://www.xbrl.org/2003/label/documentation`

It's still unclear to me whether the intent is "the _text nodes_ cannot `<`" or "the _XML_ cannot contain `<` (non-text element nodes)"...

> This rule includes the prohibition of alternative serializations of the `<` character such as `&lt;`, `&#x3C;`, `&#60;`.

I believe `textValue` is checking the extracted text the same as `viewText`, not the "raw"/encoded XML, so it is already covering `&lt;`, `&#x3C;`.  Do you have an example where `if "<" in label.textValue():` is insufficient?

derekgengenbacher-wf: I dont have any examples of that. So Ill update to `if "<" in label.textValue():`.

arelle/plugin/validate/EDINET/rules/gfm.py
```diff
@@ -701,3 +702,75 @@ def rule_gfm_1_5_6(
...
+    if labelRelationshipSet is None:
+        return
+    for concept in val.modelXbrl.qnameConcepts.values():
+        labelRels = labelRelationshipSet.fromModelObject(concept)
+        for rel in labelRels:
+            label = rel.toModelObject
+            if label.textValue is not None and label.textValue != label.textValue.strip():
+                yield Validation.warning(
+                    codes='EDINET.EC5700W.GFM.1.5.8',
+                    msg=_("The concept of '%(concept)s' has a label that contains disallowed white space either at the begining or the end: '%(label)s'"),
```
allenbross-wf, line 884: I don't know if it's significant, but these validation messages are far different from the one in the validation spreadsheet.

aaroncameron-wk (dismissed):

austinmatherne-wk (dismissed):

arelle/plugin/validate/EDINET/rules/gfm.py
```diff
@@ -813,3 +783,106 @@ def rule_gfm_1_3_31(
...
+    disclosureSystems=[DISCLOSURE_SYSTEM_EDINET],
+)
+def rule_gfm_1_5_6(
+        pluginData: PluginValidationDataExtension,
+        val: ValidateXbrl,
+        *args: Any,
+        **kwargs: Any,
+) -> Iterable[Validation]:
+    """
+    EDINET.EC5700W: [GFM 1.5.6] The length of a label must be less than 511 characters unless it its role is documentation.
```
brettkail-wk: This description suggests the code should say `>= 511`, not `> 511`.

nit: "it its role".

derekgengenbacher-wf: Reread and yes we need to catch things that are 511 characters or longer.

PR #1905, opened 2025-09-12
EDINET: Filing Formats, Cover Page Rules

aaroncameron-wk:
#### Reason for change
Establish data structures that help in identifying EDINET filing formats, which is required for implementing rules such as these cover page rules.

EDINET.EC1001E: A required item is missing from the cover page.
EDINET.EC1002E: A duplicate item is included on the cover page.
EDINET.EC1003E: An unnecessary item is included on the cover page.

#### Steps to Test
CI

**review**:
@Arelle/arelle


arelle/plugin/validate/EDINET/CoverPageRequirements.py
```diff
@@ -0,0 +1,97 @@
...
+    def _load(self) -> list[list[CoverPageItemStatus | None]]:
+        if self._data is None:
+            with open(self._csvPath, encoding='utf-8') as f:
+                self._data = [
+                    [
+                        CoverPageItemStatus.parse(cell) for cell in line.strip().split(',')
+                    ]
+                    for line in f.readlines()
+                ]
+        return self._data
```
brettkail-wk, line 49: Is it worth asserting the number of lines matches `COVER_PAGE_ITEM_LOCAL_NAMES` and the number of columns of every row matches `FILING_FORMATS` (if I got that right)?  If it's possible to make the method signature be `get(self, item: QName, format: FilingFormat)`, that might be a little clearer.  Post-processing to use `dict[QName,dict[FilingFormat,CoverPageItemStatus]]` might also be nice vs needing to indirect through indexes.

derekgengenbacher-wf (approved):

derekgengenbacher-wf:

PR #1909, opened 2025-09-15
EDINET GFM Element rules

allenbross-wf:
#### Description of change

EDINET. GFM.1.3.20: The nillable attribute value of an xsd:element must equal "true".

EDINET.GFM.1.3.23: If the abstract attribute of xsd:element is "true", then the xbrli:periodType attribute must be "duration".

EDINET.GFM.1.3.30: If xsd:element type attribute equals "nonnum:domainItemType" then the abstract attribute must equal to "true".

EDINET.GFM.1.3.31: If xsd:element type attribute equals "nonnum:domainItemType" then the abstract attribute must equal to "true".

#### Steps to Test

CI

**review**:
@Arelle/arelle


arelle/plugin/validate/EDINET/PluginValidationDataExtension.py
```diff
@@ -138,6 +145,33 @@ def _isDebitConcept(self, concept: ModelConcept) -> bool:
...
+        instantDomainElements: set[ModelConcept] = set()
+        nonAbstractDomainElements: set[ModelConcept] = set()
+        nonNillableElements: set[ModelConcept] = set()
+        xbrlPeriodType = "{http://www.xbrl.org/2003/instance}periodType"
+        for modelDocument in modelXbrl.urlDocs.values():
+            for elt in modelDocument.xmlRootElement.iterdescendants(tag="{http://www.w3.org/2001/XMLSchema}element"):
+                if elt.namespaceURI == XbrlConst.xsd:
+                    if elt.get("nillable") == "false":
+                        nonNillableElements.add(elt)
+                    if elt.get("abstract") == "true" and elt.get(xbrlPeriodType) == "instant":
```
aaroncameron-wk: I think this is unnecessarily re-parsing these values when it's already done as a part of concept parsing. For example, this EDGAR [implementation](https://github.com/Arelle/EDGAR/blob/0d42a79acebae3cc125b55d5762e790261d69138/validate/DTS.py#L233-L235) just references ModelConcepts. 

`ElementData` might not be necessary at that point if each of these rules was just a `for concept in modelXbrl.qnameConcepts.values()` instead of needing to walk each document's element tree, but I don't feel strongly about that part.

tests/resources/conformance_suites/edinet/EC5700W.GFM.1.3.20/index.xml
```diff
@@ -0,0 +1,22 @@
...
+        xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+        xmlns="http://xbrl.org/2005/conformance"
+        name="EDINET.EC5700W.GFM.1.3.20"
+        description="EDINET.EC5700W.GFM.1.3.20"
+        outpath=''
+        owner="support@arelle.org"
+        xsi:schemaLocation="http://xbrl.org/2005/conformance https://www.xbrl.org/2005/conformance.xsd">
+    <variation id="invalid01" name="invalid01">
+        <description>
+            The XSD contains a non-nillable element.
```
aaroncameron-wk: It's helpful to leave a comment explaining what original valid filing was used and what was changed. That way, if we need to recreate it at some point, it's a bit easier.

arelle/plugin/validate/EDINET/PluginValidationDataExtension.py
```diff
@@ -138,6 +145,33 @@ def _isDebitConcept(self, concept: ModelConcept) -> bool:
...
+    @lru_cache(1)
+    def checkElements(self, modelXbrl: ModelXbrl) -> ElementData:
+        instantAbstractElements: set[ModelConcept] = set()
+        instantDomainElements: set[ModelConcept] = set()
+        nonAbstractDomainElements: set[ModelConcept] = set()
+        nonNillableElements: set[ModelConcept] = set()
+        for modelDocument in modelXbrl.urlDocs.values():
+            for concept in modelDocument.xmlRootElement.iterdescendants(tag="{http://www.w3.org/2001/XMLSchema}element"):
+                if isinstance(concept,ModelConcept):
+                    if concept.namespaceURI == XbrlConst.xsd:
```
aaroncameron-wk: Is there a reason not to do it this way? 
```suggestion
        for concept in modelXbrl.qnameConcepts.values():
```
As I alluded to in my other comment, if we've avoided the need to re-walk the XML tree for these rules, caching the results in `ElementData` doesn't seem to have a ton of benefit over just iterating over concepts separately one time for each rule (which would decouple these rules from each other).

PR #1912, opened 2025-09-17
ROS:20 Check currencies dimension

austinmatherne-wk:
#### Reason for change
PrincipalCurrencyUsedInBusinessReport is a fixedItemType concept which requires a zero length value. The currencies dimension should be checked instead.

#### Description of change
Use the ISO code defined in the taxonomy for the currency dimension member and check it against the unit IDs.

#### Steps to Test
* CI

**review**:
@Arelle/arelle

### DIFF
--- a/arelle/plugin/validate/ROS/rules/ros.py
+++ b/arelle/plugin/validate/ROS/rules/ros.py
@@ -30,7 +30,7 @@ from arelle.ValidateXbrlCalcs import inferredDecimals, rangeValue
 from arelle.XbrlConst import qnXbrliMonetaryItemType, qnXbrliXbrl, xhtml
 from arelle.XmlValidateConst import VALID
 from ..ValidationPluginExtension import CURRENCIES_DIMENSION, EQUITY, PRINCIPAL_CURRENCY, TURNOVER_REVENUE
-from ..PluginValidationDataExtension import MANDATORY_ELEMENTS,  SCHEMA_PATTERNS, TR_NAMESPACES, PluginValidationDataExtension
+from ..PluginValidationDataExtension import MANDATORY_ELEMENTS, SCHEMA_PATTERNS, TR_NAMESPACES, UK_REF_NS_PATTERN, PluginValidationDataExtension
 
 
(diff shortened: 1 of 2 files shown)


arelle/plugin/validate/ROS/rules/ros.py
```diff
@@ -337,27 +338,48 @@ def rule_ros20(
...
-        for unit, count in unit_counts.items():
+        for count in unit_counts.values():
             if count > principal_currency_value_count:
                 yield Validation.warning(
                     "ROS.20",
-                    _("'PrincipalCurrencyUsedInBusinessReport' has a value of %(principal_currency_value)s, "
+                    _("'PrincipalCurrencyUsedInBusinessReport' has a %(currencies_dimension)s value of %(principal_currency_value)s, "
                       "which must match the functional(majority) unit of the financial statement."),
                     modelObject=principal_currency_facts,
+                    CurrenciesDimension=CURRENCIES_DIMENSION,
```
brettkail-wk:
```suggestion
                    currencies_dimension=CURRENCIES_DIMENSION,
```

arelle/plugin/validate/ROS/rules/ros.py
```diff
@@ -337,27 +338,48 @@ def rule_ros20(
...
+    for dim, mem in fact.context.qnameDims.items():
+        if dim.localName != CURRENCIES_DIMENSION:
+            continue
+        if mem.xValid < VALID:
+            return None
+        mem_concept = modelXbrl.qnameConcepts.get(mem.xValue)
+        if mem_concept is None:
+            return None
+        for ref_rel in modelXbrl.relationshipSet(XbrlConst.conceptReference).fromModelObject(mem_concept):
+            if code := ref_rel.toModelObject.findtext("{*}Code"):
```
brettkail-wk: It's somewhat loose to match without a namespace.  Perhaps a small comment saying that this is intended to match `uk-ref:Code`?

aaroncameron-wk (approved):

PR #1914, opened 2025-09-17
EDINET: EC1015E , EC1035E

aaroncameron-wk:
#### Description of change
EDINET.EC1015E: A URI in an HTML file within a subdirectory must not specify a path directly beneath the report folder.
EDINET.EC1035E: A URI in an HTML file must not specify a path to a location higher than the report path.

#### Steps to Test
CI

**review**:
@Arelle/arelle


arelle/plugin/validate/EDINET/rules/upload.py
```diff
@@ -814,7 +816,33 @@ def rule_uri_references(
                 modelObject=uriReference.element,
             )
             continue
-        fullPath = Path(uriReference.document.uri).parent / path
+        documentFullPath = Path(uriReference.document.uri)
+        documentPathInfo = uploadContents.uploadPathsByFullPath.get(documentFullPath)
+        if documentPathInfo is not None and any(part in documentPathInfo.path.parts for part in path.parts):
```
austinmatherne-wk: Does this match if a sub directory has the same name as a parent directory? For example, if `XBRL/001.jpg` was used in the test doc instead of `images/001.jpg`?

aaroncameron-wk: After looking into this a bit more, I'm actually thinking this could be the intent. This set of rules is not very well described and, depending on interpretation, could have rules that are very redundant. This interpretation is the only one I could come up with that wasn't redundant with the others in place. It's not unlikely that we'll end up changing this in the future, but I think that applies to a lot of these rules.

austinmatherne-wk: Which rule is it redundant with if it excludes subdirectories?

derekgengenbacher-wf (dismissed):

derekgengenbacher-wf:

PR #1918, opened 2025-09-18
Plugin Loading Improvements 

austinmatherne-wk:
#### Reason for change
Developers running Arelle from source and working with EDGAR plugins have faced issues handling dependencies on plugins defined in other repositories (EDGAR and XULE).

These changes target two common scenarios:
* Loading EDGAR plugins outside `arelle/plugin`: Relative imports (e.g., `EDGAR/render` → `EDGAR/validate`) currently fail, forcing developers to manually provide paths for all plugins including their plugin dependencies.
* Using the full XULE repo under arelle/plugin: Since the plugin isn’t located at `arelle/plugin/xule/__init__.py` but instead at `arelle/plugin/<repo-name>/plugin/xule/__init__.py`, it fails to load.

#### Description of change
* Support relative plugin dependency imports from plugins loaded outside of `arelle/plugin`.
* Fallback to a recursive search within the plugin directory to locate matching plugins.
* Update GUI plugin display priority:
  * Prefer the new EDGAR plugin over Edgar Renderer.
  * Move XULE and Inline XBRL Document Set plugins higher in the list.

#### Steps to Test
* Clone the [EDGAR plugin repo](https://github.com/Arelle/Edgar) into a directory outside of `arelle/plugin`.
* Modify the [EDGAR/validate plugin info import](https://github.com/Arelle/EDGAR/blob/0d42a79acebae3cc125b55d5762e790261d69138/validate/__init__.py#L950) to include `xule`.
* Clone the [XULE plugin repo](https://github.com/xbrlus/xule).
* Symlink or move the entire XULE repo under `arelle/plugin/<repo-name>`.
* Load the EDGAR plugin from the external directory and confirm validation runs with XULE rules:
  * `python arelleCmdLine.py --file <report-file> --plugins /path/to/repo/EDGAR --validate --disclosureSystem efm`

**review**:
@Arelle/arelle

### DIFF
--- a/arelle/Cntlr.py
+++ b/arelle/Cntlr.py
@@ -161,6 +161,7 @@ class Cntlr:
         disable_persistent_config: bool = False,
         betaFeatures: dict[str, bool] | None = None
     ) -> None:
+        self.logger = None
         if betaFeatures is None:
             betaFeatures = {}
         self.betaFeatures = {
--- a/arelle/DialogPluginManager.py
+++ b/arelle/DialogPluginManager.py
@@ -386,15 +386,17 @@ class DialogPluginManager(Toplevel):
         group = {
             "ixbrl-viewer": "1",  # pip installed Arelle viewer
             "iXBRLViewerPlugin": "2",  # git clone installed Arelle viewer
-            "Edgar Renderer": "3",
+            "EDGAR": "3",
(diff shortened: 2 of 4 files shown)


derekgengenbacher-wf:

PR #1919, opened 2025-09-18
HMRC: Add validations for audited group companies and audited group LLPs

derekgengenbacher-wf:
#### Reason for change
- Added validation for audited group companies
- Added validation for audited group LLPs

#### Steps to Test
CI

**review**:
@Arelle/arelle


arelle/plugin/validate/UK/ValidateUK.py
```diff
@@ -646,6 +651,31 @@ def _evaluateProfLossOrCharityFundsFact(self, code: str) -> CodeResult:
...
+        """
+        Logs an error if BalanceSheetDate with GroupCompanyDataDimension and Consolidated must exist as non-nil and
+        equal BalanceSheetDate with GroupCompanyDataDimension and Consolidated
+        """
+        consolidatedFact = None
+        defaultFact = None
+        for balanceSheetDateFact in self._getFacts(CONCEPT_BALANCE_SHEET_DATE):
+            if self._checkValidFact(balanceSheetDateFact):
+                if not balanceSheetDateFact.context.qnameDims:
+                    defaultFact = balanceSheetDateFact
```
austinmatherne-wk: Can there be more than one non-dimensional fact? Such as tagged for 2024 and 2024, or tagged with USD or EUR.

arelle/plugin/validate/UK/ValidateUK.py
```diff
@@ -646,6 +651,31 @@ def _evaluateProfLossOrCharityFundsFact(self, code: str) -> CodeResult:
...
+                for qname, value in balanceSheetDateFact.context.qnameDims.items():
+                    if value.xValid < VALID:
+                        continue
+                    if qname.localName == CONCEPT_GROUP_COMPANY_DATA_DIMENSION and cast(str, value.xValue.localName) == CONCEPT_CONSOLIDATED:
+                        consolidatedFact = balanceSheetDateFact
```
austinmatherne-wk, line 677: nit: you can `break` out of the qname loop once `consolidatedFact` is found.

arelle/plugin/validate/UK/ValidateUK.py
```diff
@@ -646,6 +651,31 @@ def _evaluateProfLossOrCharityFundsFact(self, code: str) -> CodeResult:
             )
         return CodeResult()
 
+    def _evaluateGroupFacts(self) -> CodeResult:
+        """
+        Logs an error if BalanceSheetDate with GroupCompanyDataDimension and Consolidated must exist as non-nil and
+        equal BalanceSheetDate with GroupCompanyDataDimension and Consolidated
+        """
```
austinmatherne-wk: I'm not sure what this comment is trying to say, but it seems inconsistent with the code below (which doesn't check if values are nil).

derekgengenbacher-wf: Reworded to make more sense in english, but _checkValidFact is called which verifies that the fact is not nil

arelle/plugin/validate/UK/ValidateUK.py
derekgengenbacher-wf: No this should only be tagged once.

austinmatherne-wk: Should be, but what happens if the do? Is a different error raised?

derekgengenbacher-wf: If there are multiple, then the duplicate fact check should kick in. Because this is a date fact and with default dimensions any duplicates of this fact that are not consistent would be flagged.

austinmatherne-wk: To be clear, I'm not talking about duplicate facts, but distinct facts tagged without dimensions and different values and contexts. A similar issue exists for multiple facts of different dates tagged with the dimension and member:

For example the error will be raised only depending on the order in which these facts are processed as either of the first two are eligible to be the `defaultFact`:
```xml
<bus:BalanceSheetDate contextRef="instant-2022">2022-01-01</bus:BalanceSheetDate>
<bus:BalanceSheetDate contextRef="instant-2021">2021-01-01</bus:BalanceSheetDate>
<bus:BalanceSheetDate contextRef="instant-2021-with-member">2021-01-01</bus:BalanceSheetDate>
```

derekgengenbacher-wf: Resolved by checking context instant date against the end date fact which is mandatory.

allenbross-wf (approved):

aaroncameron-wk (approved):

PR #1923, opened 2025-09-19
Code tidy and improvement in LeiUtil.py

strr:
#### Reason for change
The code looked like it would benefit from a make-over.

#### Description of change

- Make the code clearer
- Make the code faster (seems to be about 10x faster)
- Move the tests to a unit-test and out of `__main__`
- No functional change expected

I hope to do a follow-on pull request: wiring-up LeiUtil as a native implementation of the lei taxonomy's `validate_checksum` formula function.

#### Steps to Test

**review**:
@Arelle/arelle


austinmatherne-wk:
Thank you for the contribution! 🙌

For context: the EFM conformance suite CI check failed here only because the branch is behind master. It tried to pull an older version of the suite from AWS, which we no longer host there.

PR #1924, opened 2025-09-19
Add GFM.1.6.1, GFM.1.7.1, GFM.1.8.1

derekgengenbacher-wf:
#### Reason for change
- Added GFM.1.6.1
- Added GFM.1.7.1
- Added GFM.1.8.1

#### Steps to Test
- CI

**review**:
@Arelle/arelle


arelle/plugin/validate/EDINET/rules/gfm.py
```diff
@@ -970,3 +970,80 @@ def rule_gfm_1_5_10(
...
+    presentationRelationshipSet = val.modelXbrl.relationshipSet(XbrlConst.parentChild)
+    if presentationRelationshipSet is None:
+        return
+    for rel in presentationRelationshipSet.modelRelationships:
+        if not rel.arcElement.get("order"):
+            yield Validation.warning(
+                codes='EDINET.EC5700W.GFM.1.6.1',
+                msg=_("The presentation relationship is missing the order attribute"),
+                modelObject=rel
+            )
```
aaroncameron-wk, line 998: Should we fire one warning with multiple refs, rather than one for each? Same for other two.

arelle/plugin/validate/EDINET/rules/gfm.py
```diff
@@ -970,3 +970,80 @@ def rule_gfm_1_5_10(
...
+        val: ValidateXbrl,
+        *args: Any,
+        **kwargs: Any,
+) -> Iterable[Validation]:
+    """
+    EDINET.EC5700W: [GFM 1.8.1] All definition relationships must have an order attribute
+    """
+    definitionRelationshipSet = val.modelXbrl.relationshipSet(
+        (XbrlConst.all, XbrlConst.dimensionDefault, XbrlConst.dimensionDomain, XbrlConst.domainMember, XbrlConst.hypercubeDimension)
+    )
```
aaroncameron-wk:
```suggestion
    definitionRelationshipSet = val.modelXbrl.relationshipSet(
        tuple(LinkbaseType.DEFINITION.getArcroles())
    )
```
Likewise with LinkbaseType.CALCULATION, LinkbaseType.PRESENTATION above

aaroncameron-wk (dismissed):

arelle/plugin/validate/EDINET/rules/gfm.py
derekgengenbacher-wf: I thought about that, but that could get kind of messy real quick if they dont have order on a lot of things. I think I would prefer one per error.

aaroncameron-wk: Ah, I equate less individual errors with being less messy, but I guess it depends on how those errors/refs are displayed.

aaroncameron-wk (approved):

PR #1926, opened 2025-09-19
Add GFM.1.7.2, GFM.1.7.3 and GFM.1.7.6

derekgengenbacher-wf:
#### Reason for change
- Added GFM.1.7.2
- Added GFM.1.7.3
- Added GFM.1.7.6

#### Steps to Test
CI

**review**:
@Arelle/arelle


arelle/plugin/validate/EDINET/rules/gfm.py
```diff
@@ -1023,6 +1023,97 @@ def rule_gfm_1_7_1(
...
+    """
+    calculationRelationshipSet = val.modelXbrl.relationshipSet(tuple(LinkbaseType.CALCULATION.getArcroles()))
+    if calculationRelationshipSet is None:
+        return
+    for modelObject, rels in calculationRelationshipSet.loadModelRelationshipsFrom().items():
+        if len(rels) <= 1:
+            continue
+        relsByOrder = defaultdict(list)
+        for rel in rels:
+            relsByOrder[(rel.arcElement.get("order"), rel.linkrole)].append(rel)
```
austinmatherne-wk: Should this rule take into account default order values (`1.0` if not set)? `rel.order` handles this automatically.

arelle/plugin/validate/EDINET/rules/gfm.py
```diff
@@ -1023,6 +1023,97 @@ def rule_gfm_1_7_1(
...
+        *args: Any,
+        **kwargs: Any,
+) -> Iterable[Validation]:
+    """
+    EDINET.EC5700W: [GFM 1.7.6] Calculation relationships must have unique order attributes
+    """
+    calculationRelationshipSet = val.modelXbrl.relationshipSet(tuple(LinkbaseType.CALCULATION.getArcroles()))
+    if calculationRelationshipSet is None:
+        return
+    for modelObject, rels in calculationRelationshipSet.loadModelRelationshipsFrom().items():
```
austinmatherne-wk: nit: We typically call `relationshipSet.fromModelObjects()` instead of `relationshipSet.loadModelRelationshipsFrom()` (they do the same thing).

arelle/plugin/validate/EDINET/rules/gfm.py
derekgengenbacher-wf: The rule specifically calls out that Presentation, Calculation and Definition relationships have to have an `order` attribute. If Arelle defaults to `1.0` that hides the fact that the attribute is not present in the XML.

aaroncameron-wk (approved):

arelle/plugin/validate/EDINET/rules/gfm.py
aaroncameron-wk: For this rule, it makes sense not to use the `1.0` default, but maybe we should not add elements without an `order` attribute here since it will be flagged by the other rule. As-is, two arcs without an `order` attribute would be flagged as duplicates here, which seems redundant with the other rule.

derekgengenbacher-wf: Yes I agree partially. If we use the 1.0 here, then that will flag other `1.0` orders which seems like a false positive.

arelle/plugin/validate/EDINET/rules/gfm.py
```diff
@@ -1089,6 +1091,99 @@ def rule_gfm_1_7_1(
...
+    """
+    EDINET.EC5700W: [GFM 1.7.3] The concepts participating in a calculation relationship must have the same period type
+    """
+    calculationRelationshipSet = val.modelXbrl.relationshipSet(tuple(LinkbaseType.CALCULATION.getArcroles()))
+    if calculationRelationshipSet is None:
+        return
+    for rel in calculationRelationshipSet.modelRelationships:
+        fromConcept = rel.fromModelObject
+        toConcept = rel.toModelObject
+        if fromConcept.periodType != toConcept.periodType:
```
austinmatherne-wk: We should guard against undefined attribute access errors here. both `rel.fromModelObject` and `rel.toModelObject` can be `None`.

PR #1930, opened 2025-09-22
EDINET: EC5700W.GFM.1.3.25

allenbross-wf:
#### Description of change

EDINET: EC5700W.GFM.1.3.25
> Correct the element name so that it does not end with "Axis", or correct the
    substitutionGroup to "xbrldt:dimensionItem".

#### Steps to Test

CI

**review**:
@Arelle/arelle


arelle/plugin/validate/EDINET/rules/gfm.py
```diff
@@ -783,6 +783,32 @@ def rule_gfm_1_3_23(
...
+    substitutionGroup to "xbrldt:dimensionItem".
+    """
+    improperlyNamedAxes= set()
+    for concept in pluginData.getExtensionConcepts(val.modelXbrl):
+        if concept.qname.localName.endswith("Axis") and not concept.substitutionGroupQname == XbrlConst.qnXbrldtDimensionItem:
+            improperlyNamedAxes.add(concept)
+    if len(improperlyNamedAxes) > 0:
+        yield Validation.warning(
+            codes='EDINET.EC5700W.GFM.1.3.25',
+            msg=_("Modify the element name so that it does not end with 'Axis', or modify the substitutionGroup to 'xbrldt:dimensionItem'."),
```
austinmatherne-wk: When fired in the GUI or CLI without ref logging this message won't inform the user which concepts need to be fixed.

arelle/plugin/validate/EDINET/rules/gfm.py
```diff
@@ -783,6 +783,30 @@ def rule_gfm_1_3_23(
...
+        val: ValidateXbrl,
+        *args: Any,
+        **kwargs: Any,
+) -> Iterable[Validation]:
+    """
+    EDINET.EC5700W: [GFM 1.3.25] Correct the element name so that it does not end with "Axis", or correct the
+    substitutionGroup to "xbrldt:dimensionItem".
+    """
+    for concept in pluginData.getExtensionConcepts(val.modelXbrl):
+        if concept.qname.localName.endswith("Axis") and not concept.substitutionGroupQname == XbrlConst.qnXbrldtDimensionItem:
```
aaroncameron-wk: The original GFM documentation:
> The xsd:element substitutionGroup attribute must equal "xbrldt:dimensionItem" **if and only if** the name attribute ends with "Axis".

Based on that, I think the error should fire for an "Axis" concept that is not a dimension item OR a non-"Axis" concept that is a dimension item. ([EDGAR implementation](https://github.com/Arelle/EDGAR/blob/0d42a79acebae3cc125b55d5762e790261d69138/validate/DTS.py#L250-L251))

arelle/plugin/validate/EDINET/rules/gfm.py
```diff
@@ -812,6 +812,30 @@ def rule_gfm_1_3_23(
...
+        val: ValidateXbrl,
+        *args: Any,
+        **kwargs: Any,
+) -> Iterable[Validation]:
+    """
+    EDINET.EC5700W: [GFM 1.3.25] Correct the element name so that it does not end with "Axis", or correct the
+    substitutionGroup to "xbrldt:dimensionItem".
+    """
+    for concept in pluginData.getExtensionConcepts(val.modelXbrl):
+        if concept.qname.localName.endswith("Axis") ^ (concept.substitutionGroupQname == XbrlConst.qnXbrldtDimensionItem):
```
austinmatherne-wk: nit: `!=` is more pythonic IMO:
```suggestion
        if concept.qname.localName.endswith("Axis") 
    != (concept.substitutionGroupQname == XbrlConst.qnXbrldtDimensionItem):
```

arelle/plugin/validate/EDINET/rules/gfm.py
```diff
@@ -812,6 +812,30 @@ def rule_gfm_1_3_23(
...
+        val: ValidateXbrl,
+        *args: Any,
+        **kwargs: Any,
+) -> Iterable[Validation]:
+    """
+    EDINET.EC5700W: [GFM 1.3.25] Correct the element name so that it does not end with "Axis", or correct the
+    substitutionGroup to "xbrldt:dimensionItem".
+    """
+    for concept in pluginData.getExtensionConcepts(val.modelXbrl):
+        if concept.qname.localName.endswith("Axis") != (concept.substitutionGroupQname == XbrlConst.qnXbrldtDimensionItem):
```
brettkail-wk, line 832: A similar question as [here](https://github.com/Arelle/Arelle/pull/1932#pullrequestreview-3259124512): it's unclear whether to me whether this code should check `and concept.substitutionQname != XbrlConst.qnXbrldtDimensionItem`

allenbross-wf: I have updated the comment on the rule.  The GFM rule uses and if and only if relationship that is unclear in the EDINET translation.

PR #1932, opened 2025-09-23
EDINET: GFM.1.3.26, GFM.1.3.28, GFM.1.3.29

allenbross-wf:
#### Description of change

GFM.1.3.26: The xsd:element name attribute must ends with "Table" if and only ifsubstitutionGroup attribute equals "xbrldt:hypercubeItem".

GFM.1.3.28: If xsd:element name attribute ends with "LineItems" then the abstract attribute must equal "true".

GFM.1.3.29: The xsd:element name attribute must end with "Domain" or "Member" if and only if the type attribute equals "nonnum:domainItemType".

#### Steps to Test

CI

**review**:
@Arelle/arelle


arelle/plugin/validate/EDINET/rules/gfm.py
```diff
@@ -812,6 +811,82 @@ def rule_gfm_1_3_23(
...
+        val: ValidateXbrl,
+        *args: Any,
+        **kwargs: Any,
+) -> Iterable[Validation]:
+    """
+    EDINET.EC5700W: [GFM 1.3.29] If the element name of an element extended by the submitter-specific taxonomy ends with
+    "Domain" or "Member", please set the type attribute to "nonnum:domainItemType".
+    """
+    for concept in pluginData.getExtensionConcepts(val.modelXbrl):
+        isConceptDomain = bool(concept.type and getattr(concept.type, "isDomainItemType", False))
```
austinmatherne-wk: `concept.type` can be `None` or `ModelType`, so this should be written as:
```suggestion
        isConceptDomain = concept.type.isDomainItemType if concept.type is not None else False
```

arelle/plugin/validate/EDINET/rules/gfm.py
```diff
@@ -812,6 +811,82 @@ def rule_gfm_1_3_23(
...
+        if ((concept.qname.localName.endswith("Domain")  or concept.qname.localName.endswith("Member")) != isConceptDomain):
```
austinmatherne-wk: nit: whitespace
```suggestion
        if ((concept.qname.localName.endswith("Domain") or concept.qname.localName.endswith("Member")) != isConceptDomain):
```

arelle/plugin/validate/EDINET/rules/gfm.py
```diff
@@ -812,6 +811,82 @@ def rule_gfm_1_3_23(
...
+        val: ValidateXbrl,
+        *args: Any,
+        **kwargs: Any,
+) -> Iterable[Validation]:
+    """
+    EDINET.EC5700W: [GFM 1.3.26] Correct the element name so that it does not end with "Table", or correct the
+    substitutionGroup to "xbrldt:hypercubeItem".
+    """
+    for concept in pluginData.getExtensionConcepts(val.modelXbrl):
+        if concept.qname.localName.endswith("Table") != (concept.substitutionGroupQname == XbrlConst.qnXbrldtHypercubeItem):
```
brettkail-wk, line 832: The docstring implies to me only that concepts ending with "Table" must be hypercubes, not that all hypercubes must end with "Table".  i.e., the docstring implies to me:
```
if concept.qname.localNAme.endsWith("Table") and concept.substitutionGroupQname != XbrlConst.qnXbrldtHypercubeItem:
```
Is the docstring or code wrong?

Similar question for the domain rule in this PR.

allenbross-wf: I have updated the docstrings for the with the acutal GFM wording.  In both cases they specify  "if and only if" relationships.

brettkail-wk: Sounds good, thanks.

PR #1934, opened 2025-09-23
Add GFM.1.7.4 and GFM.1.8.4

derekgengenbacher-wf:
#### Reason for change
- Added GFM.1.7.4
- Added GFM.1.8.4

#### Steps to Test
- CI

**review**:
@Arelle/arelle


arelle/plugin/validate/EDINET/rules/gfm.py
```diff
@@ -1184,6 +1184,35 @@ def rule_gfm_1_7_3(
...
+        val: ValidateXbrl,
+        *args: Any,
+        **kwargs: Any,
+) -> Iterable[Validation]:
+    """
+    EDINET.EC5700W: [GFM 1.7.4] A calculation link can not contain directed cycles.
+    """
+    cycleObjects = set()
+    fromRelationships = val.modelXbrl.relationshipSet(tuple(LinkbaseType.CALCULATION.getArcroles())).fromModelObjects()
+    for fromObject, rels in fromRelationships.items():
```
austinmatherne-wk: nit: only using keys:
```suggestion
    for fromObject in fromRelationships:
```

arelle/plugin/validate/EDINET/rules/gfm.py
```diff
@@ -1184,6 +1184,35 @@ def rule_gfm_1_7_3(
...
+        if fromObject in cycleObjects:
+            continue
```
austinmatherne-wk: I think there's a potential gap in the cycle detection logic here.

Consider this scenario:
```
- A → B, B → A (forms cycle 1)
- B → C, C → B (forms cycle 2)
```

If A is processed first, it will:
1. Detect the A → B → A cycle
2. Add B to `cycleObjects`
3. When B is later processed, it gets skipped due to being included in `cycleObjects`
4. The B → C → B cycle is never detected

This means overlapping cycles could be missed depending on the iteration order. We should either remove the `cycleObjects` optimization or modify it to properly handle cases where a node participates in multiple cycles.

arelle/plugin/validate/EDINET/PluginValidationDataExtension.py
```diff
@@ -500,3 +500,18 @@ def iterValidNonNilFacts(self, modelXbrl: ModelXbrl, qname: QName) -> Iterable[M
     def addUsedFilepath(self, modelXbrl: ModelXbrl, path: Path) -> None:
         controllerPluginData = ControllerPluginData.get(modelXbrl.modelManager.cntlr, self.name)
         controllerPluginData.addUsedFilepath(path)
+
+    def directedCycle(self, fromObject: ModelObject, origin: ModelObject, fromRelationships: dict[Any, list[ModelRelationship]], path: set[ModelObject]) -> None | list[ModelRelationship]:
+        if fromObject in fromRelationships:
+            for rel in fromRelationships[fromObject]:
+                toObject = rel.toModelObject
```
austinmatherne-wk: It'll fall through and end up returning None later, so it's okay, but it would be better to explicitly handle the None toModelObject case to make this more resilient to future refactors.

arelle/plugin/validate/EDINET/rules/gfm.py
```diff
@@ -1184,6 +1184,35 @@ def rule_gfm_1_7_3(
...
+        pluginData: PluginValidationDataExtension,
+        val: ValidateXbrl,
+        *args: Any,
+        **kwargs: Any,
+) -> Iterable[Validation]:
+    """
+    EDINET.EC5700W: [GFM 1.7.4] A calculation link can not contain directed cycles.
+    """
+    cycleObjects = set()
+    fromRelationships = val.modelXbrl.relationshipSet(tuple(LinkbaseType.CALCULATION.getArcroles())).fromModelObjects()
```
brettkail-wk: Cycles are usually defined per arcrole (`cyclesAllowed`), but this seems to detect cycles across arcroles.  Is that deliberate?

Or is the rule just intending to specify `cyclesAllowed` for the XBRL 2.1 calc arcrole, which otherwise doesn't specify a restriction (but was considered an oversight and fixed in calc 1.1)?  If so, I wonder if it would be simpler to adjust `XbrlValidate` to allow plugin-specific `standardArcroleCyclesAllowed`.

arelle/plugin/validate/EDINET/rules/gfm.py
derekgengenbacher-wf: In this scenario:
```
A → B, B → A (forms cycle 1)
B → C, C → B (forms cycle 2)
```
We loop through all of the from concepts, in this case it would look something like this:

1) Loop through all descendants of A
2) Detect A -> B -> A cycle
3) Add B to `cycleObjects`
4) Skip looping through all descendants of B thereby removing a duplicate error for the yet to be found cycle 2. Cycle 1 wont be found again because we will consistently find cycle 2 because of descendant ordering.
5) Loop through all descendants of C
6) Detect B -> C -> B cycle

The key here is that the descendants are looped through by order and we are only firing an error when the initial concept is found in the path for a 2nd time.

If we remove the `cycleObjects` parameter, then we get a duplicate error for cycle 2 because it is found twice. Once when searching B and again when searching C.

arelle/plugin/validate/EDINET/rules/gfm.py
derekgengenbacher-wf: This was an oversight on my part. Ive updated the code accordingly

arelle/plugin/validate/EDINET/rules/gfm.py
derekgengenbacher-wf: I have updated the test to take into account this scenario.

austinmatherne-wk: You're right, but then:

```
A → B → A (cycle 1)
C → D → C (cycle 2)
B → D → B (cycle 3)
```

Only reports 2 of 3 cycles, depending on what order they're processed?

1. Loop through all descendants of `A`
2. Detect cycle 1 `A → B → A`
3. Add `B` to `cycleObjects`
4. Loop through all descendants of `C`
5. Detect cycle 2 `C → D → C`
6. Add `D` to `cycleObjects`
7. Skip checking descendants of `B`
8. Skip checking descendants of `D`
9. Cycle `B → D → B` is never detected

I'm okay if we decide this isn't worth the effort, but a code comment would be helpful in case a future dev has to debug this 🙂

tests/resources/conformance_suites/edinet/EC5700W.GFM.1.8.4/index.xml
```diff
@@ -3,21 +3,22 @@
...
+            Two undirected cycles
+            <!--valid03.zip with 2 undirected cycles-->
         </description>
         <data>
             <instance readMeFirst="true">invalid01.zip</instance>
         </data>
         <result>
-            <warning>EDINET.EC5700W.GFM.1.10.14</warning>
+            <error>EDINET.EC5700W.GFM.1.8.4</error>
+            <error>EDINET.EC5700W.GFM.1.8.4</error>
```
aaroncameron-wk:
```suggestion
            <error num="2">EDINET.EC5700W.GFM.1.8.4</error>
```

tests/integration_tests/validation/conformance_suite_configurations/edinet.py
```diff
@@ -24,15 +24,6 @@
...
         },
-        "EC8024E/index.xml:invalid01": {
-            "EDINET.EC8027W": 1,
-        },
-        "EC8058W/index.xml:invalid01": {
-            "EDINET.EC8027W": 1,
-        },
-        "EC8062W/index.xml:invalid01": {
-            "EDINET.EC8027W": 1,
-        },
```
aaroncameron-wk, line 35: Why did these code changes affect EC8027W?

derekgengenbacher-wf: I fixed the filings to only throw the error that it was meant to throw.

aaroncameron-wk (approved):

PR #1939, opened 2025-09-26
EDINET: Table of Contents

aaroncameron-wk:
#### Reason for change
EDINET.EC2001E: There must be at least one table of contents entry in each file.
EDINET.EC2002W: The table of contents number must be present.
EDINET.EC2003E: The table of contents must be no longer than 384 bytes.
EDINET.EC2004E: Corner brackets (【) must not overlap.
EDINET.EC2005E: Table of contents entries must not be duplicated.
EDINET.EC2007E: The table of contents entries must be enclosed in square brackets (】).
EDINET.EC2008E: Table of contents entries must not contain HTML tags.
EDINET.EC3002E: Table of contents numbers for table of contents entries must not be repeated within the same hierarchy.
EDINET.EC3003E: There must be no gaps in the table of contents numbers within the same hierarchy.
EDINET.EC3005E: The table of contents numbers for the table of contents entries must be as specified in the format.

#### Steps to Test
CI

**review**:
@Arelle/arelle


arelle/plugin/validate/EDINET/TableOfContentsBuilder.py
```diff
@@ -1,9 +1,12 @@
+"""
+See COPYRIGHT.md for copyright information.
+"""
+from __future__ import annotations
```
brettkail-wk, line 4: Out of curiosity, what does this `from`/`import` do?  A quick search suggests it's a no-op on Python 3.9+.

aaroncameron-wk: It's still necessary in 3.9 to resolve forward references in type hints. In this case, I suppose this module is imported before `ModelDocument` and/or `ModelObject` (in type-hints below). Not exactly sure, but we have a test that verifies imports and it failed in 3.9 without this statement. I expected we're dropping 3.9 soon, so hopefully an issue we'll be able to forget before long.

allenbross-wf (approved):

allenbross-wf:
QA +10
All tests passed

PR #1948, opened 2025-09-30
Add GFM.1.7.5

derekgengenbacher-wf:
#### Reason for change
- Added GFM.1.7.5

#### Steps to Test
- CI

**review**:
@Arelle/arelle


arelle/plugin/validate/EDINET/rules/gfm.py
```diff
@@ -1392,6 +1344,50 @@ def rule_gfm_1_7_3(
...
+        conceptsMissingRels = []
+        concepts = [rel.fromModelObject, rel.toModelObject]
+        presentationRelationshipSet = val.modelXbrl.relationshipSet(tuple(LinkbaseType.PRESENTATION.getArcroles()), rel.linkrole)
+        if presentationRelationshipSet is None:
+            conceptsMissingRels.extend([concept for concept in concepts if concept is not None])
+        else:
+            for concept in concepts:
+                if concept is None:
+                    continue
+                conceptPresentationRels = presentationRelationshipSet.fromModelObject(concept) + presentationRelationshipSet.toModelObject(concept)
```
brettkail-wk: This appears to only check that the concepts appear in the role, not that they're in the same arc?  For example, if the role has calc A->B and pres X->[A,B], it seems like that should be an error.

It also appears to be quadratic.  It seems like for each role with calc arcs, we should collect set[tuple[from,to]], and then for each calc arc, verify that the role contains the tuple.

derekgengenbacher-wf: The rule states that the source and target of a calculation relationship must appear in either:
1) a presentation relationship with each other
2) two presentation relationships with any other elements that also share a single ELR.

So your example above is valid because A and B would share the same ELR. As far as I can tell we just need to check to make sure that the source and target appear in a presentation relationship in the same ELR. So checking exact tuples would eliminate option 2 of the rule.

brettkail-wk: > This appears to only check that the concepts appear in the role, not that they're in the same arc?

Derek pointed me to the GFM rule, which says more completely (emphasis mine).

> > If an instance contains _non-empty facts for the source and target of an effective calculation relationship_, then at least one effective presentation relationship that the source and target appear in must be either (a) a relationship with each other or **(b) two relationships with any other elements that share a single extended link role.**

Does the rule need to check the _non-empty facts_ part?

> It also appears to be quadratic.

I misread the code, it's just looking for relationships for the specific concepts (which will be indexed).  It might be marginally more efficient to check `len(from) == 0 and len(to) == 0` rather than concatenating, but whatever.

PR #1951, opened 2025-10-01
Add GFM.1.8.5

derekgengenbacher-wf:
#### Reason for change
- Added GFM.1.8.5

#### Steps to Test
- CI

**review**:
@Arelle/arelle


arelle/plugin/validate/EDINET/__init__.py
```diff
@@ -24,6 +24,8 @@
         "xbrl.4.8.2:sharesFactUnit-notSingleMeasure": ("ERROR", None),
         # Silence, duplicated by EDINET.EC5700W.GFM.1.7.2
         "xbrl.5.2.5.2.1:zeroWeight": ("ERROR", None),
+        # Silence, duplicated by EDINET.EC5700W.GFM.1.8.5
+        "xbrl.5.1.4.3:cycles": ("ERROR", None),
```
aaroncameron-wk: Are there situations in which `xbrl.5.1.4.3:cycles` would normally fire that `EDINET.EC5700W.GFM.1.8.5` would not? If so, adding `xbrl.5.1.4.3:cycles` to the testcase results for `EDINET.EC5700W.GFM.1.8.5` would be preferred.

PR #1953, opened 2025-10-02
EDINET: EC5613W -> EC5613E

aaroncameron-wk:
#### Reason for change
EDINET.EC5623W: Instances using IFRS taxonomies must set the DEI 'Accounting Standard'' value to 'IFRS'.

Modified EC5613W (warning) -> EC5613E (error). Documentation specifies that the error level code should fire for "next-gen" taxonomies.

#### Steps to Test
CI

**review**:
@Arelle/arelle


derekgengenbacher-wf (approved):

arelle/plugin/validate/EDINET/rules/edinet.py
```diff
@@ -325,6 +325,33 @@ def rule_EC5614E(
...
+    disclosureSystems=[DISCLOSURE_SYSTEM_EDINET],
+)
+def rule_EC5623W(
+        pluginData: PluginValidationDataExtension,
+        val: ValidateXbrl,
+        *args: Any,
+        **kwargs: Any,
+) -> Iterable[Validation]:
+    """
+    EDINET.EC5623W: Instances using IFRS taxonomies must set the DEI "Accounting Standard" value to "IFRS".
```
brettkail-wk, line 339: Do we need to verify that a fact exists at all?  Is that covered elsewhere?

aaroncameron-wk: EC5614E validates the presence of DEI facts (including AccountingStandardsDEI) based on the form.

brettkail-wk: Ah, got it, thanks.  I had grep'ed by `accountingStandardsDeiQn` and didn't find it.  It might be nice if `DEI_LOCAL_NAMES` was replaced with a list of QNames populated from the same local vars rather than having the bare string in two places.  (Or remove the `accountingStandardsDeiQn` and just inline `qname(jpdeiNamespace, 'AccountingStandardsDEI')`, or define separate constants for the local name.)

PR #1957, opened 2025-10-03
WIP EDINET.EC5700W.GFM.1.3.18

allenbross-wf:
#### Description of change

EDINET.EC5700W.GFM.1.3.18

#### Steps to Test

CI

**review**:
@Arelle/arelle


derekgengenbacher-wf (dismissed):

arelle/plugin/validate/EDINET/rules/gfm.py
```diff
@@ -647,6 +647,43 @@ def rule_gfm_1_3_1(
...
+            for refDoc in document.referencesDocument.values():
+                if 'import' in refDoc.referenceTypes:
+                    namespace = refDoc.referringModelObject.attrib.get('namespace')
+                    if namespace:
+                        if namespace.endswith('iod'):
+                            namespace = namespace[:-3]
+                        if namespace.startswith(EDINET_NAMESPACE_PREFIX):
+                            schemaLocation = refDoc.referringModelObject.attrib.get('schemaLocation')
+                            if schemaLocation and not schemaLocation.startswith(namespace):
+                                     yield Validation.warning(
```
aaroncameron-wk: It's generally preferred to not nest a bunch of `if` statements that lead to excessive indentation.
```suggestion
        if not pluginData.isExtensionUri(document.uri, val.modelXbrl) or not document.type == ModelDocument.Type.SCHEMA:
            continue
        for refDoc in document.referencesDocument.values():
            if 'import' not in refDoc.referenceTypes:
                continue
            namespace = refDoc.referringModelObject.attrib.get('namespace')
            if not namespace:
                continue
            if namespace.endswith('iod'):
               namespace = namespace[:-3]
            if namespace.startswith(EDINET_NAMESPACE_PREFIX):
                schemaLocation = refDoc.referringModelObject.attrib.get('schemaLocation')
                if schemaLocation and not schemaLocation.startswith(namespace):
                    yield Validation.warning(
```

arelle/plugin/validate/EDINET/rules/gfm.py
```diff
@@ -647,6 +647,43 @@ def rule_gfm_1_3_1(
...
+    schema, then its schemaLocation attribute must be the standard taxonomy assigned
+    to that namespace.
+    """
+    for document in val.modelXbrl.urlDocs.values():
+        if pluginData.isExtensionUri(document.uri, val.modelXbrl) and document.type == ModelDocument.Type.SCHEMA:
+            for refDoc in document.referencesDocument.values():
+                if 'import' in refDoc.referenceTypes:
+                    namespace = refDoc.referringModelObject.attrib.get('namespace')
+                    if namespace:
+                        if namespace.endswith('iod'):
```
aaroncameron-wk: It's not clear to me why this is necessary. Is there something EDINET-specific going on here?

aaroncameron-wk: Could we use `Namespace` values in `edinet-taxonomies.xml` to determine when "an xsd:import element has a namespace attribute equal to a standard taxonomy schema", and then compare the `schemaLocation` to `Href`?

allenbross-wf: It is specific to EDINET and the Identification and Ordering namespace.  The namespace as imported is: `http://disclosure.edinet-fsa.go.jp/taxonomy/common/2013-08-31/iod`. The schema location for the XSD is  `http://disclosure.edinet-fsa.go.jp/taxonomy/common/2013-08-31/identificationAndOrdering_2013-08-31.xsd`

aaroncameron-wk: I don't love relying on the pattern that the schema location URL _starts with_ the corresponding namespace URI, and then needing to hardcode in the exceptions to that rule. Ideally these implementations would be portable to other jurisdictions. We have this existing design around "standard taxonomies" defined in the XML that I think we can make use of.

arelle/plugin/validate/EDINET/rules/gfm.py
```diff
@@ -696,6 +696,46 @@ def rule_gfm_1_3_1(
...
+) -> Iterable[Validation]:
+    """
+    EDINET.EC5700W: [GFM 1.3.2] The schemaLocation attribute of the xsd:import element specifying the EDINET taxonomy does not point to a valid location.
+
+    GFM-1.3.2:  If an xsd:import element has a namespace attribute equal to a standard taxonomy
+    schema, then its schemaLocation attribute must be the standard taxonomy assigned
+    to that namespace.
+    """
+    for document in val.modelXbrl.urlDocs.values():
+        if not pluginData.isExtensionUri(document.uri, val.modelXbrl) or not document.type == ModelDocument.Type.SCHEMA:
```
brettkail-wk, line 749: Why the first `not`?  Doesn't that mean this code is only checking extension taxonomies?

arelle/plugin/validate/EDINET/rules/gfm.py
```diff
@@ -696,6 +696,46 @@ def rule_gfm_1_3_1(
...
+            continue
+        for refDoc in document.referencesDocument.values():
+            if 'import' not in refDoc.referenceTypes:
+                continue
+            namespace = refDoc.referringModelObject.attrib.get('namespace')
+            if not namespace:
+                continue
+            if namespace.endswith('iod'):
+                namespace = namespace[:-3]
```
brettkail-wk: Out of curiosity, what is this matching?

arelle/plugin/validate/EDINET/rules/gfm.py
```diff
@@ -696,6 +696,46 @@ def rule_gfm_1_3_1(
...
+            if namespace.startswith(EDINET_NAMESPACE_PREFIX):
+                schemaLocation = refDoc.referringModelObject.attrib.get('schemaLocation')
+                if schemaLocation and not schemaLocation.startswith(namespace):
```
brettkail-wk: It seems like we should be checking `schemaLocation == namespace + ".xsd"` to reject cases where the `schemaLocation` is valid but contains unwanted trialing characters (for example, a query parameter).

That only works if EDINET namespace URIs and XSD URLs are consistent enough.  Otherwise, I guess we need a full table of namespace/XSD pairs like EDGAR?  I wonder whether `http` vs `https` should be tolerated now or will need to be in the future.

arelle/plugin/validate/EDINET/rules/gfm.py
brettkail-wk: I missed Aaron's comment [here](https://github.com/Arelle/Arelle/pull/1957#discussion_r2403397520), but my two comments are largely the same concerns.

(If this particular logic is correct, it seems worth adding a comment explaining what's going on.)

arelle/plugin/validate/EDINET/rules/gfm.py
aaroncameron-wk: It doesn't seem like it adds value for us to repeatedly validate the EDINET/XBRL schemas at a significant processing cost, so I think only checking extension taxonomies for this rule is appropriate.

arelle/plugin/validate/EDINET/rules/gfm.py
```diff
@@ -728,6 +728,44 @@ def rule_gfm_1_3_1(
...
+        if not pluginData.isExtensionUri(document.uri, val.modelXbrl) or not document.type == ModelDocument.Type.SCHEMA:
+            continue
+        for refDoc in document.referencesDocument.values():
+            if 'import' not in refDoc.referenceTypes:
+                continue
+            namespace = refDoc.referringModelObject.attrib.get('namespace')
+            schemaLocation = refDoc.referringModelObject.attrib.get('schemaLocation')
+            if not namespace or not schemaLocation:
+                continue
+            expectedXSDLocation = val.disclosureSystem.standardTaxonomiesDict.get(namespace)
```
aaroncameron-wk: Nit: variable name could be clearer to indicate that this is multiple URIs.

arelle/plugin/validate/EDINET/rules/gfm.py
```diff
@@ -728,6 +728,44 @@ def rule_gfm_1_3_1(
...
+            if expectedXSDLocation is not None and schemaLocation not in expectedXSDLocation:
+                yield Validation.warning(
+                    codes='EDINET.EC5700W.GFM.1.3.2',
+                    msg=_("The schemaLocation attribute of the xsd:import element specifying the EDINET taxonomy does not point to a valid location. "
+                          "The schemaLocation attribute value '%(schemaLocation)s'."),
+                    schemaLocation = schemaLocation,
+                    modelObject = refDoc
```
aaroncameron-wk: Should probably use `referringModelObject` as the ref. I'm not sure a `ModelDocumentReference` can be used as a ref.

Issue #1788, opened 2025-07-16
ESEF Conformance Suite 2024 results


TomaszBukowski123:
### What happened?

Hi,

Hope you doing well. While reviewing ESEF Conformance Suite 2024, I noticed a case where only one of the expected errors is returned during validation. It is about **G3.1.1** in which it includes two expected errors for two situations : 
<img width="688" height="495" alt="Image" src="https://github.com/user-attachments/assets/9fc0b492-2598-4487-9c68-b63951a015a4" />

and test cases `TC2_invalid` and `TC3_invalid` from Conformance Suite **G3-1-1_2** expect these two errors:
<img width="1132" height="834" alt="Image" src="https://github.com/user-attachments/assets/f8c55f85-ea03-4cb0-8f25-c48158d6d50a" />

 but unfortunately only `TC3_invalid` reports both while `TC2_invalid` reports only `linkbasesNotSeparateFiles` without reporting `extensionTaxonomyWrongFilesStructure`:
<img width="1201" height="82" alt="Image" src="https://github.com/user-attachments/assets/7154c357-df5c-49d2-b388-b74591cc3885" />

I've also noticed similar situation in:
- **G3-1-2** `TC7_invalid`
- **G3-4-2_4** `TC3_invalid`
- **RTS_Annex_III_Par_1** `TC3_invalid`
- **RTS_Annex_IV_Par_14_G2-5-1** `TC2_invalid`

Is this validation behavior correct and expected? Should we look only for any expected error not all of them? Neither the manual nor the suite nor the documentation for it explicitly state whether we should require all or any of the expected errors. 

Thanks in advance for your support.

Best regards,
Tomasz

### Documents

[ConformanceExpectedErrors.zip](https://github.com/user-attachments/files/21252810/ConformanceExpectedErrors.zip)

### Interface

GUI (arelleGUI)

### Version

Arelle(r) 2.37.35 (64bit)

### Download

GitHub release (bundled Python)

### Operating System

Windows

TomaszBukowski123:
Hi again, I thought that maybe something changed in a meanwhile, but on newest release of Arelle (2.37.53 (64bit)) issue still persists. What's more i noticed similar cases even in older ESEF Conformance Suite 2023 (haven't checked all of them yet, but in **G3-1-1_2** `TC2_invalid` case situation is exactly the same as described above). 

Best regards,
Tomasz

PR #1961, opened 2025-10-06
Improve ESEF conformance tests as detected in issue #1788

gmongelli:
#### Reason for change
This change aims at making the current ESEF plug-in compliant with the  ESEF conformance tests 2023 and 2024.

For a test case in G3-1-1_2 (`TC2_invalid`), the error 'extensionTaxonomyWrongFilesStructure' is missing  (several linkbases are included in a single file).

For a test case in G3-1-2 (`TC7_invalid`), the error 'incorrectEsefTaxonomyVersionUsed' is missing (case where the taxonomy entry point of the case is used in an outdated taxonomy).

For a test case in RTS_Annex_IV_Par_14_G2-5-1 (`TC2_invalid`), the error 'inlineXbrlDocumentContainsExternalReferences' is  missing (the HTML instance tries to download some executable code from a network source).

#### Description of change

For the above mentioned test cases, the errors are now issued.

The improvement for G3-1-2 is not very useful, but it makes the specific test case pass.

#### Steps to Test

Run the ESEF conformance tests G3-1-1_2, G3-1-2 and RTS_Annex_IV_Par_14_G2-5-1 and make sure that the above mentioned errors are generated in the corresponding test cases.

**review**:
@Arelle/arelle

### DIFF
--- a/arelle/plugin/validate/ESEF/ESEF_Current/ValidateXbrlFinally.py
+++ b/arelle/plugin/validate/ESEF/ESEF_Current/ValidateXbrlFinally.py
@@ -33,7 +33,7 @@ from arelle.PythonUtil import strTruncate
 from arelle.utils.Contexts import partitionModelXbrlContexts
 from arelle.utils.Units import partitionModelXbrlUnits
 from arelle.utils.validate.DetectScriptsInXhtml import containsScriptMarkers
-from arelle.UrlUtil import isHttpUrl, scheme
+from arelle.UrlUtil import isHttpUrl, isExternalUrl
 from arelle.ValidateUtr import ValidateUtr
 from arelle.ValidateXbrl import ValidateXbrl
 from arelle.ValidateXbrlCalcs import inferredDecimals, rangeValue
@@ -360,13 +360,13 @@ def validateXbrlFinally(val: ValidateXbrl, *args: Any, **kwargs: Any) -> None:
                                 # Check if the attribute 'archive' contains an external reference!
                                 archiveAttr = elt.get("archive", "").strip()
                                 for possibleURI in archiveAttr.split(" "):
-                                    if scheme(possibleURI.strip()) in ("http", "https", "ftp"):
+                                    if isExternalUrl(possibleURI.strip()):
                                         externalReferenceFound = True
                                         break
                             elif eltTag == "script":
                                 # Check if the attribute 'src' contains an external reference!
                                 srcAttr = elt.get("src", "").strip()
-                                if scheme(srcAttr) in ("http", "https", "ftp"):
+                                if isExternalUrl(srcAttr):
                                     externalReferenceFound = True
                             if externalReferenceFound:
                                 modelXbrl.error(
@@ -392,7 +392,7 @@ def validateXbrlFinally(val: ValidateXbrl, *args: Any, **kwargs: Any) -> None:
                         #    or to other sections of the annual financial report.'''
                         #elif eltTag == "a":
                         #    href = elt.get("href","").strip()
-                        #    if scheme(href) in ("http", "https", "ftp"):
+                        #    if isExternalUrl(href):
                         #        modelXbrl.error("ESEF.4.1.6.xHTMLDocumentContainsExternalReferences" if val.unconsolidated
                         #                        else "ESEF.3.5.1.inlineXbrlDocumentContainsExternalReferences",
                         #            _("Inline XBRL instance documents MUST NOT contain any reference pointing to resources outside the reporting package: %(element)s"),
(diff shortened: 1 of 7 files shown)


arelle/plugin/validate/ESEF/ESEF_Current/DTS.py
```diff
@@ -386,6 +386,16 @@ def lc3wordAdjust(word: str) -> str:
             val.modelXbrl.error("ESEF.3.1.1.linkbasesNotSeparateFiles",
                 _("Each linkbase type MUST be provided in a separate linkbase file, found: %(linkbasesFound)s."),
                 modelObject=modelDocument.xmlRootElement, linkbasesFound=", ".join(sorted(linkbasesFound)))
+            # found linkbases are not valid, so enable top level warning message for 3.1.1
+            for foundLinkBase in linkbasesFound:
+                if foundLinkBase == "calculationLink":
+                    val.hasExtensionCal = False
```
brettkail-wk: If the package contains (for example) a valid calc file and also an extraneous invalid linkbase containing multiple links including calc, it seems wrong to report `extensionTaxonomyWrongFilesStructure`.

It seems like the setting of `hasExtensionCal` (etc.) should only be set when we're certain it's correct rather than trying to "undo" that setting here.

arelle/plugin/validate/ESEF/ESEF_Current/ValidateXbrlFinally.py
```diff
@@ -355,6 +355,27 @@ def checkFootnote(elt: ModelInlineFootnote | ModelResource, text: str) -> None:
                             modelXbrl.error(f"{contentOtherThanXHTMLGuidance}.executableCodePresent",
                                 _("Inline XBRL documents MUST NOT contain executable code: %(element)s"),
                                 modelObject=elt, element=eltTag)
+                            externalReferenceFound = False
+                            if eltTag == "object":
+                                # Check if the attribute 'archive' contains an external reference!
+                                archiveAttr = elt.get("archive", "").strip()
+                                for possibleURI in archiveAttr.split(" "):
+                                    if scheme(possibleURI.strip()) in ("http", "https", "ftp"):
```
brettkail-wk: This is fine, but it might be nice to add an `isExternalUrl` helper to consolidate the logic with the image check.

Well, it's odd that stylesheets are checked using `isInArchive` instead?  It seems like that check might be preferable.  Do you know why the different code paths are using different mechanisms for validating "external"?

arelle/plugin/validate/ESEF/ESEF_Current/DTS.py
gmongelli: If there is a valid calc file and an extraneous linkbase file that is not part of the DTS, that file will not be scanned. If the extraneous linkbase is part of the DTS, the message `extensionTaxonomyWrongFilesStructure` is appropriate.

The problem with setting `hasExtensionCal` immediately with the right value is that you have to scan the complete DTS and to remember all found linkbases. With the implemented algorithm, there is no need to save the encountered linkbase files. You just keep the current state. However, I'll check what happens if you encounter first an invalid file and then the valid one. I think in this case the error message willl no longer be issued. I'll check that!

arelle/plugin/validate/ESEF/ESEF_Current/ValidateXbrlFinally.py
gmongelli: I'll try to add a uniform isExternalUrl check! Thank you for the suggestion!

I do not know why there are different mechanisms for validating "external". For me it seems natural to first check the scheme.  Having checked how `isInArchive` works, it does not check by default if the file is actually included in the archive file. You have to specify explicitly `checkExistence=True` for that purpose. I'll check if the current code needs to be fine-tuned. In any case, it is not safe to accept absolute URLs even if they get mapped through an XML catalog file. Thus, I would not use `isInArchive` with a normalised URL: imagine you unzip a report package and you open a report in your browser. Your browser will not read any XML catalog. It will interpret the "http..." address as an internet URL and try to download it as usual. For that reason, I prefer the test for the scheme!

arelle/plugin/validate/ESEF/ESEF_Current/DTS.py
gmongelli: As a matter of fact, the algorithm was failing  if you encounter first an invalid file and then the valid one(s).

I simply changed the code to output explicitly both required messages...

arelle/plugin/validate/ESEF/ESEF_Current/ValidateXbrlFinally.py
gmongelli: `isExternalUrl` has been added and is now used all over the source code

PR #1963, opened 2025-10-06
EDINET: Various Fixes and Improvements

aaroncameron-wk:
#### Reason for change
I've implemented a process that scrapes public filings from the EDINET index and runs our validations against them. The following fixes (review by commit) have been implemented as a result of the findings from this process (or were otherwise identified as improvements to be made).

#### Steps to Test
CI

**review**:
@Arelle/arelle


requirements.txt
```diff
@@ -4,6 +4,7 @@ certifi==2025.10.5
 cheroot==11.0.0
 filelock==3.19.1
 isodate==0.7.2
+jaconv==0.4.0
```
austinmatherne-wk, line 7: This also needs to be included in pyproject.toml dependencies for library users.

arelle/ValidateDuplicateFacts.py
```diff
@@ -210,6 +210,8 @@ def deduplicateConsistentPairs(self) -> list[ModelFact]:
         if not self.areNumeric:
             # Consistency is equivalent to completeness for non-numeric facts
             return facts
+        if len(facts) < 2:
+            return facts
```
brettkail-wk: Out of curiosity, what issue is this fixing?  Is it just skipping a bit of working building the map?

arelle/plugin/validate/EDINET/PluginValidationDataExtension.py
```diff
@@ -502,7 +502,7 @@ def getFilingFormat(self, modelXbrl: ModelXbrl) ->  FilingFormat | None:
                 continue
             prefixes = {taxonomy.value for taxonomy in filingFormat.taxonomies}
             if not any(
-                str(fact.xValue).startswith(filingFormat.documentType.value) and
+                str(fact.xValue).strip().startswith(filingFormat.documentType.value) and
```
brettkail-wk, line 502: Pre-existing, what `fact` variable is this using?  Is this using the last iteration of the `documentTitleFacts` loop above?  Is there a chance that `fact` can be `None` here?  Regardless, is there some way to make the logic clearer?

I was looking at the code because I was trying to understand what concept this code is verifying and trying to understand why `strip()` is needed rather than being able to rely on `collapse`.

arelle/ValidateDuplicateFacts.py
aaroncameron-wk: This came up because I encountered a scenario where a complete duplicate subset consisted of all nil facts. This results in `deduplicateCompleteSubsets` returning a single fact. The `decimals is not None` assertion then fails, but we really don't even need to be iterating over the facts at all when there's a single fact.

arelle/plugin/validate/EDINET/PluginValidationDataExtension.py
aaroncameron-wk: This statement iterates over `documentTitleFacts` which can not have `None` elements (our type-checking would have flagged that here).

I can simplify the `documentTitleFacts` population above to eliminate the conflicting local variable names.

I'm not sure what you mean re: relying on `collapse`.

brettkail-wk: > This statement iterates over `documentTitleFacts`

🤦 I missed that it's a generator expression since I'm just not used to reading Python.  I don't think there's any change you need to make.  Sorry for the distraction.

> I'm not sure what you mean re: relying on collapse.

I meant the [`whiteSpace` facet](https://www.w3.org/TR/2012/REC-xmlschema11-2-20120405/datatypes.html#rf-whiteSpace).  Depending on the datatype of the concept, the `strip` shouldn't be needed, but I assume that the relevant concepts are using types that aren't that.  (This was all just for my own understanding/investigation.)

PR #1965, opened 2025-10-07
Add FRTA.4.2.2, FRTA.4.2.4, FRTA.4.2.7 and FRTA.4.2.11

derekgengenbacher-wf:
#### Reason for change
- Added FRTA.4.2.2
- Added FRTA.4.2.4
- Added FRTA.4.2.7
- Added FRTA.4.2.11

#### Steps to Test
- CI

**review**:
@Arelle/arelle

### DIFF
--- a/arelle/plugin/validate/EDINET/__init__.py
+++ b/arelle/plugin/validate/EDINET/__init__.py
@@ -24,8 +24,6 @@ RELEVELER_MAP: dict[str, dict[str, tuple[str, str | None]]] = {
         "xbrl.4.8.2:sharesFactUnit-notSingleMeasure": ("ERROR", None),
         # Silence, duplicated by EDINET.EC5700W.GFM.1.7.2
         "xbrl.5.2.5.2.1:zeroWeight": ("ERROR", None),
-        # Silence, Duplicated by EDINET.EC5710W.FRTA.4.2.2
-        "xbrl.4.2.2:schemaRefHref": ("ERROR", None),
         # Silence, Duplicated by EDINET.EC5710W.FRTA.4.2.11
         "xbrl.5.1:emptyTargetNamespace": ("ERROR", None),
     },
--- a/tests/integration_tests/validation/conformance_suite_configurations/edinet.py
+++ b/tests/integration_tests/validation/conformance_suite_configurations/edinet.py
@@ -169,19 +169,6 @@ ADDITIONAL_INVALID_ERRORS = {
         "EDINET.EC5700W.GFM.1.8.4": 2,
         "xbrl.5.1.4.3:cycles": 1,
     },
-    "EC5710W.FRTA.4.2.2/index.xml:invalid01": {
-        # Modified version of valid09
-        "EDINET.EC5800E.FATAL_ERROR_TOC_TREE_NOT_DEFINED": 1,
-        "EDINET.EC5700W.GFM.1.2.8": 1,
-        "EDINET.EC5700W.GFM.1.2.27": 1,
-        "arelle:NoMatchingEdinetFormat": 1,
-        "EDINET.EC1057E": 1,
-        "ix10.11.1.2:missingReferences": 43,
-    },
-    "EC5710W.FRTA.4.2.4/index.xml:invalid01": {
-        # Modified version of valid09
-        "EDINET.FRTA.2.1.10": 1,
-    },
     "EC5806E/index.xml:invalid01": {
         # Instance duplicated means table of contents are included twice.
         "EDINET.EC2005E": 2,
--- a/tests/resources/conformance_suites/edinet/EC5710W.FRTA.4.2.2/index.xml
+++ b/tests/resources/conformance_suites/edinet/EC5710W.FRTA.4.2.2/index.xml
@@ -17,7 +17,14 @@
             <instance readMeFirst="true">invalid01.zip</instance>
         </data>
         <result>
+            <warning>arelle:NoMatchingEdinetFormat</warning>
+            <warning>EDINET.EC1057E</warning>
+            <warning>EDINET.EC5700W.GFM.1.2.8</warning>
+            <warning>EDINET.EC5700W.GFM.1.2.27</warning>
             <warning>EDINET.EC5710W.FRTA.4.2.2</warning>
+            <warning>EDINET.EC5800E.FATAL_ERROR_TOC_TREE_NOT_DEFINED</warning>
+            <warning num="43">ix10.11.1.2:missingReferences</warning>
+            <warning>xbrl.4.2.2:schemaRefHref</warning>
         </result>
     </variation>
 </testcase>
--- a/tests/resources/conformance_suites/edinet/EC5710W.FRTA.4.2.4/index.xml
+++ b/tests/resources/conformance_suites/edinet/EC5710W.FRTA.4.2.4/index.xml
@@ -18,6 +18,7 @@
             <instance readMeFirst="true">invalid01.zip</instance>
         </data>
         <result>
+            <warning>EDINET.FRTA.2.1.10</warning>
             <warning num="2">EDINET.EC5710W.FRTA.4.2.4</warning>
         </result>
     </variation>


arelle/plugin/validate/EDINET/rules/frta.py
```diff
@@ -180,3 +180,71 @@ def rule_frta_3_1_10(
...
+) -> Iterable[Validation]:
+    """
+    EDINET.EC5710W: [FRTA.4.2.2] Taxonomy schemas must be defined in XML documents in which the XML Schema 'schema'
+                                 element appears once only as the root element.
+    """
+    for modelDocument in val.modelXbrl.urlDocs.values():
+        if pluginData.isStandardTaxonomyUrl(modelDocument.uri, val.modelXbrl):
+            continue
+        for hrefElt, hrefedDoc, hrefId in modelDocument.hrefObjects:
+            if hrefElt.localName != "schemaRef":
```
brettkail-wk: What if the document is discovered via [any other DTS mechanism](https://www.xbrl.org/Specification/XBRL-2.1/REC-2003-12-31/XBRL-2.1-REC-2003-12-31+corrected-errata-2013-02-20.html#_3.2) than `schemaRef`?  I don't know whether Arelle actually supports non-root schemas elements, so I don't know if there's some more straightforward way to detect this.  If it does support those, maybe it's easiest to find the code that discovers those, and adjust it to maintain `ModelXbrl.schemas` as a list of the schema elements so this code can iterate that list and check whether the elements have a parent.

arelle/plugin/validate/EDINET/rules/frta.py
```diff
@@ -180,3 +180,71 @@ def rule_frta_3_1_10(
...
+)
+def rule_frta_4_2_4(
+        pluginData: PluginValidationDataExtension,
+        val: ValidateXbrl,
+        *args: Any,
+        **kwargs: Any,
+) -> Iterable[Validation]:
+    """
+    EDINET.EC5710W: [FRTA.4.2.4] Taxonomy schemas must be defined in XML documents in which the XML Schema 'schema'
+                                 element appears once only as the root element.
```
brettkail-wk: This docstring looks copy/paste from `rule_frta_4_2_2`.

arelle/plugin/validate/EDINET/rules/frta.py
```diff
@@ -180,3 +180,71 @@ def rule_frta_3_1_10(
...
+        if rootElt.get('elementFormDefault') != 'qualified' or rootElt.get('attributeFormDefault') != 'unqualified':
+            yield Validation.warning(
+                codes='EDINET.EC5710W.FRTA.4.2.4',
+                msg=_("The XMLSchema root in taxonomy schema files must have the 'elementFormDefault' atribute set as "
+                      "'qulaified' and the 'attributeFormDefault' attribute set as 'unqualified'"),
+                modelObject=modelDocument,
+            )
+        formUsages = []
+        for elt in rootElt.iterdescendants():
+            if elt.prefixedName not in ['xsd:element', 'xsd:attribute']:
```
brettkail-wk: It seems conceptually wrong to use `prefixedName` rather than comparing a QName, even if other rules in the taxonomy require using a canonical namespace prefix.

arelle/plugin/validate/EDINET/rules/frta.py
```diff
@@ -197,18 +197,17 @@ def rule_frta_4_2_2(
                                  element appears once only as the root element.
     """
     for modelDocument in val.modelXbrl.urlDocs.values():
-        if pluginData.isStandardTaxonomyUrl(modelDocument.uri, val.modelXbrl):
+        if pluginData.isStandardTaxonomyUrl(modelDocument.uri, val.modelXbrl) or modelDocument.type == ModelDocument.Type.SCHEMA:
```
brettkail-wk: Out of curiosity, why add the SCHEMA check here?  What if a schema file contains a nested schema in an `<xs:appinfo>`?

arelle/plugin/validate/EDINET/rules/frta.py
```diff
@@ -197,18 +197,17 @@ def rule_frta_4_2_2(
...
-            if hrefElt.localName != "schemaRef":
-                continue
-            hrefedElt = hrefedDoc.xmlRootElement
-            if hrefedElt.namespaceURI != XbrlConst.xsd:
-                yield Validation.warning(
-                    codes='EDINET.EC5710W.FRTA.4.2.2',
-                    msg=_("The root of a taxonomy schema file MUST be the XMLSchema element."),
-                    modelObject=modelDocument,
-                )
+        qnamesIncluded = {elt.namespaceURI for elt in modelDocument.xmlRootElement.iterdescendants()}
```
brettkail-wk: This doesn't seem like an accurate check, but it's probably good enough.  Perhaps just add a comment saying that that ideally this would visit the discovered schema elements and check the parent, but that's not easy, so we approximate by ensuring any file with containing a schema element is a schema.

It seems like we should check that the root element is `xs:schema`.  Otherwise, it looks like this code wrongly accepts a file like `<xs:appinfo><xs:schema>...</xs:schema></xs:appinfo>`.

arelle/plugin/validate/EDINET/rules/frta.py
```diff
@@ -180,3 +180,127 @@ def rule_frta_3_1_10(
...
+                                 element appears once only as the root element.
+    """
+    for modelDocument in val.modelXbrl.urlDocs.values():
+        if pluginData.isStandardTaxonomyUrl(modelDocument.uri, val.modelXbrl):
+            continue
+        # check for nested Schema declarations which are not allowed.
+        schemaElts = {
+            elt for elt in modelDocument.xmlRootElement.iterdescendants()
+            if hasattr(elt, 'elementQname') and elt.elementQname == XbrlConst.qnXsdSchema
+        }
```
aaroncameron-wk: Would `modelDocument.xmlRootElement.iterdescendants(XbrlConst.qnXsdSchema.clarkNotation)` work?
Similar for `FRTA.4.2.4` below. Unqualified `iterdescendants` calls should be avoided whenever possible.

arelle/plugin/validate/EDINET/rules/frta.py
```diff
@@ -180,3 +180,127 @@ def rule_frta_3_1_10(
...
+        if len(schemaElts) > 0:
+            yield Validation.warning(
+                codes='EDINET.EC5710W.FRTA.4.2.2',
+                msg=_("The root of a taxonomy schema file MUST be the XMLSchema element."),
+                modelObject=modelDocument,
+            )
```
aaroncameron-wk, line 209: Wouldn't this not fire if a schema file did not contain a `schema` element at all? Or is that not a schema file by definition?

`xbrl.4.2.2:schemaRefHref` seems to be checking any file _referenced_ as a schema.

arelle/plugin/validate/EDINET/rules/frta.py
```diff
@@ -180,3 +180,127 @@ def rule_frta_3_1_10(
...
+        *args: Any,
+        **kwargs: Any,
+) -> Iterable[Validation]:
+    """
+    EDINET.EC5710W: [FRTA.4.2.2] Taxonomy schemas must be defined in XML documents in which the XML Schema 'schema'
+                                 element appears once only as the root element.
+    """
+    for modelDocument in val.modelXbrl.urlDocs.values():
+        if pluginData.isStandardTaxonomyUrl(modelDocument.uri, val.modelXbrl):
+            continue
```
aaroncameron-wk, line 201: Can we also exit if the document is not of the `SCHEMA` type?

arelle/plugin/validate/EDINET/rules/frta.py
```diff
@@ -180,3 +180,127 @@ def rule_frta_3_1_10(
...
+)
+def rule_frta_4_2_4(
+        pluginData: PluginValidationDataExtension,
+        val: ValidateXbrl,
+        *args: Any,
+        **kwargs: Any,
+) -> Iterable[Validation]:
+    """
+    EDINET.EC5710W: [FRTA.4.2.4] Taxonomy schemas must declare elementFormDefault to be 'qualified',
+                                 attributeFormDefault must have the value 'unqualified', and the 'form attribute
```
aaroncameron-wk:
```suggestion
                                 attributeFormDefault must have the value 'unqualified', and the 'form' attribute
```

arelle/plugin/validate/EDINET/__init__.py
```diff
@@ -24,6 +24,10 @@
         "xbrl.4.8.2:sharesFactUnit-notSingleMeasure": ("ERROR", None),
         # Silence, duplicated by EDINET.EC5700W.GFM.1.7.2
         "xbrl.5.2.5.2.1:zeroWeight": ("ERROR", None),
+        # Silence, Duplicated by EDINET.EC5710W.FRTA.4.2.2
+        "xbrl.4.2.2:schemaRefHref": ("Error", None),
+        # Silence, Duplicated by EDINET.EC5710W.FRTA.4.2.11
+        "xbrl.5.1:emptyTargetNamespace": ("Error", None),
```
aaroncameron-wk: Nit:
```suggestion
        # Silence, Duplicated by EDINET.EC5710W.FRTA.4.2.2
        "xbrl.4.2.2:schemaRefHref": ("ERROR", None),
        # Silence, Duplicated by EDINET.EC5710W.FRTA.4.2.11
        "xbrl.5.1:emptyTargetNamespace": ("ERROR", None),
```

arelle/XbrlConst.py
```diff
@@ -27,6 +27,8 @@
 qnXsdSchema = qname("{http://www.w3.org/2001/XMLSchema}xsd:schema")
 qnXsdAppinfo = qname("{http://www.w3.org/2001/XMLSchema}xsd:appinfo")
 qnXsdDefaultType = qname("{http://www.w3.org/2001/XMLSchema}xsd:anyType")
+qnXsdElement = qname("{http://www.w3.org/2001/XMLSchema}element")
+qnXsdAttribute = qname("{http://www.w3.org/2001/XMLSchema}attribute")
```
aaroncameron-wk:
```suggestion
qnXsdElement = qname("{http://www.w3.org/2001/XMLSchema}xsd:element")
qnXsdAttribute = qname("{http://www.w3.org/2001/XMLSchema}xsd:attribute")
```

arelle/plugin/validate/EDINET/__init__.py
```diff
@@ -24,6 +24,10 @@
         "xbrl.4.8.2:sharesFactUnit-notSingleMeasure": ("ERROR", None),
         # Silence, duplicated by EDINET.EC5700W.GFM.1.7.2
         "xbrl.5.2.5.2.1:zeroWeight": ("ERROR", None),
+        # Silence, Duplicated by EDINET.EC5710W.FRTA.4.2.2
+        "xbrl.4.2.2:schemaRefHref": ("Error", None),
```
aaroncameron-wk: Worth considering if we can get away with just releveling this to the EDINET code, and skipping a new implementation, i.e.:

```suggestion
        "xbrl.4.2.2:schemaRefHref": ("ERROR", "EDINET.EC5710W.FRTA.4.2.2"),
```

arelle/plugin/validate/EDINET/rules/frta.py
derekgengenbacher-wf: This should fire only if there is a schema declaration below the root.

`xbrl.4.2.2:schemaRefHref` is checking the root only of any file referenced as a schema. That check does not validate if there are imbedded schema declarations within a root schema.

arelle/plugin/validate/EDINET/rules/frta.py
derekgengenbacher-wf: No, we need to make sure that there are not imbedded schema within a schema.

arelle/plugin/validate/EDINET/__init__.py
derekgengenbacher-wf: We cant do this because that only covers one of the many avenues that an imbedded schema can be found.

arelle/plugin/validate/EDINET/rules/frta.py
aaroncameron-wk: Right, but let's say an IXBRL file references an .xsd file via `schemaRef`, and that .xsd file doesn't contain a `schema` element at all. With `xbrl.4.2.2:schemaRefHref` silenced, that won't fire an error, and it won't be picked up by this implementation either.

arelle/plugin/validate/EDINET/__init__.py
```diff
@@ -24,6 +24,10 @@
         "xbrl.4.8.2:sharesFactUnit-notSingleMeasure": ("ERROR", None),
         # Silence, duplicated by EDINET.EC5700W.GFM.1.7.2
         "xbrl.5.2.5.2.1:zeroWeight": ("ERROR", None),
+        # Silence, Duplicated by EDINET.EC5710W.FRTA.4.2.2
+        "xbrl.4.2.2:schemaRefHref": ("ERROR", None),
```
aaroncameron-wk: I don't think this is actually fully replaced by the current `FRTA.4.2.2`. `FRTA.4.2.2` seems fine as-is, so maybe we just don't silence this error and have it be an expected result in the testcase variation?

derekgengenbacher-wf: `xbrl.4.2.2:schemaRefHref` checks all files that are linked by a schemaref. That file is checked to make sure that its root is the schema element. But it doesnt check anything other than that.

`FRTA.4.2.2` checks all files regardless of how they are linked in the DTS to make sure that an imbedded schema element does not exist. If a files root is not a schema element but the file contains a schema element(imbedded) and its linked via a schemaref then we cover the above validation. 

In other words, `xbrl.4.2.2:schemaRefHref` will never fire without `FRTA.4.2.2` also firing. But `FRTA.4.2.2` can fire without `xbrl.4.2.2:schemaRefHref`.

aaroncameron-wk: [invalid01.zip](https://github.com/user-attachments/files/22853790/invalid01.zip)
This is valid09.zip with this line added:
```
<link:schemaRef xlink:type="simple" xlink:href="jpcrp050300-esr-001_X99007-000_2025-04-10_01_2025-04-10_pre.xml" />
```
So a `schemaRef` import that references a document with _no_ `schema` element. Not captured by `FRTA.4.2.2` but _is_ captured by `xbrl.4.2.2:schemaRefHref`.

derekgengenbacher-wf: Ah good point. Removing.

PR #1967, opened 2025-10-09
Add GFM.1.5.1, GFM.1.5.2, GFM.1.5.3 and GFM.1.5.5

derekgengenbacher-wf:
#### Reason for change
- Added GFM.1.5.1
- Added GFM.1.5.2
- Added GFM.1.5.3
- Added GFM.1.5.5

#### Steps to Test
- CI

**review**:
@Arelle/arelle

### DIFF
--- a/tests/integration_tests/validation/conformance_suite_configurations/edinet.py
+++ b/tests/integration_tests/validation/conformance_suite_configurations/edinet.py
@@ -158,16 +158,6 @@ ADDITIONAL_INVALID_ERRORS = {
     "EC5700W.GFM.1.2.13/index.xml:invalid01": {
         "EDINET.EC5700W.GFM.1.1.3": 1,
     },
-    "EC5700W.GFM.1.5.1/index.xml:invalid01": {
-        "EDINET.EC5700W.GFM.1.5.3": 1,
-    },
-    "EC5700W.GFM.1.5.6/index.xml:invalid01": {
-        "EDINET.EC5700W.GFM.1.5.3": 1,
-        "EDINET.EC5700W.GFM.1.5.5": 1,
-    },
-    "EC5700W.GFM.1.5.10/index.xml:invalid01": {
-        "EDINET.EC5700W.GFM.1.5.3": 1,
-    },
     "EC5700W.GFM.1.7.3/index.xml:invalid01": {
         "EDINET.EC5700W.GFM.1.7.5": 1,
     },
@@ -180,9 +170,6 @@ ADDITIONAL_INVALID_ERRORS = {
         "EDINET.EC2002W": 1,
         "EDINET.EC5700W.GFM.1.8.4": 2,
         "xbrl.5.1.4.3:cycles": 1,
-    },
-    "EC5710W.FRTA.2.1.10/index.xml:invalid01": {
-        "EDINET.EC5700W.GFM.1.5.1": 1,
     },
     "EC5806E/index.xml:invalid01": {
         # Instance duplicated means table of contents are included twice.
--- a/tests/resources/conformance_suites/edinet/EC5700W.GFM.1.5.1/index.xml
+++ b/tests/resources/conformance_suites/edinet/EC5700W.GFM.1.5.1/index.xml
@@ -19,6 +19,7 @@
         </data>
         <result>
             <warning>EDINET.EC5700W.GFM.1.5.1</warning>
+            <warning>EDINET.EC5700W.GFM.1.5.3</warning>
         </result>
     </variation>
 </testcase>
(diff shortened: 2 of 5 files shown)


arelle/plugin/validate/EDINET/rules/frta.py
```diff
@@ -70,7 +70,7 @@ def rule_frta_2_1_10(
             errors.append(concept)
     if len(errors) > 0:
         yield Validation.warning(
-            codes='EDINET.FRTA.2.1.10',  # Not associated with EC5710W code.
+            codes='EDINET.EC5710W.FRTA.2.1.10',  # Not associated with EC5710W code.
```
aaroncameron-wk: For whatever reason, [documentation](https://disclosure2dl.edinet-fsa.go.jp/guide/static/disclosure/download/ESE140307.pdf) does not link some of the FRTA codes with `EC5710W`, which is why it was left out here. I don't mind including it for consistency across our FRTA validations, but you can remove the comment here in that case.

arelle/plugin/validate/EDINET/PluginValidationDataExtension.py
```diff
@@ -60,6 +60,8 @@
     'https://www.w3.org/1999/xlink'
 ))
 
+LANG_ATTRIBUTE_VALUES = frozenset({'ja', 'jp', 'ja-jp', 'JA', 'JP', 'JA-JP'})
```
aaroncameron-wk: Nit: `Constants.py`

arelle/plugin/validate/EDINET/Constants.py
```diff
@@ -171,3 +171,6 @@ class AccountingStandard(Enum):
 # Context IDs for facts associated with Financial Statements
 # Example: Prior2YearDuration
 FINANCIAL_STATEMENT_CONTEXT_ID_PATTERN = regex.compile(rf'^{PATTERN_CONTEXT_RELATIVE_PERIOD}{PATTERN_CONTEXT_DURATION}.*')
+
+# Accepted language codes for Japan
+JAPAN_LANGUAGE_CODES = frozenset({'ja', 'jp', 'ja-jp', 'JA', 'JP', 'JA-JP'})
```
brettkail-wk, line 176: Pre-existing, but is this list defined somewhere?  `ja` and `ja-JP` seem like the only correct language tags in this list.

derekgengenbacher-wf: These are defined as the acceptable language codes in EDINET.EC1011E

tests/integration_tests/validation/conformance_suite_configurations/edinet.py
```diff
@@ -154,6 +158,16 @@
...
+    "EC5700W.GFM.1.5.1/index.xml:invalid01": {
+        "EDINET.EC5700W.GFM.1.5.3": 1,
+    },
+    "EC5700W.GFM.1.5.6/index.xml:invalid01": {
+        "EDINET.EC5700W.GFM.1.5.3": 1,
+        "EDINET.EC5700W.GFM.1.5.5": 1,
+    },
+    "EC5700W.GFM.1.5.10/index.xml:invalid01": {
+        "EDINET.EC5700W.GFM.1.5.3": 1,
+    },
```
aaroncameron-wk: If these are expected errors based on the contents of the filing, they should be in the conformance suite. These "expected additional errors" are more for errors that we suspect should not be firing for the testcase. 

The TOC validation codes like EC2002W, for example, are something I don't have a high level of confidence in yet, and since they fire on the "valid" documents, it'll be easier to just update this file rather than a hundred testcase files if a fix is made. I/we should probably leave comments here as well to explain why they're here rather than the suite.

derekgengenbacher-wf: Those are not expected errors based on the valid document. All of those rules deal specifically with various permutations of label existence and label languages. The documents of GFM.1.5.1, GFM.1.5.6 and GFM.1.5.10 have label issues in the English label linkbase which triggers GFM.1.5.3 to fire because the added bad labels dont also exist in Japanese. It was an oversight on my part that I didnt create the label issues in the default language linkbase instead of the English.

aaroncameron-wk: These additional errors are for when we suspect that a problem with _our implementation_, not the testcase, is causing them to fire.

If someone was to run their own software against this conformance suite, would we expect these errors to fire? If so, then they belong in the conformance suite. If the invalid GFM.1.5.1, GFM.1.5.6 and GFM.1.5.10 testcases clearly also violate GFM.1.5.3, then GFM.1.5.3 should be in the expected results in the conformance suite.

arelle/plugin/validate/EDINET/Constants.py
brettkail-wk: It seems useful to add that to a comment now that this is no longer defined in `rule_html_elements`.

PR #1975, opened 2025-10-13
Support setting sourceZipStreamFileName from fairly standard Python 'name' attribute

strr:
#### Reason for change
At the moment, arelle/api/Session.py users struggle to pass report packages ending `.xbri` as Arelle converts their filename to `/POSTupload.zip`.

When you discover this, you read quite a bit of Arelle's source code and realise you have to coerce your zip stream to `FileNamedBytesIO`, which takes a fileName argument. Nothing else will do.

FileSource.py / openFileSource() will set `sourceZipStreamFileName` to what you have specified as the fileName attribute (rather than the default of `/POSTupload.zip`) and you can avoid rpe:documentTypeFileExtensionMismatch errors when validating an `.xbri` stream.

I'd like to make this dance a bit easier for API users.

#### Description of change

In `openFileSource()`, check to see if there's a `name` attribute on the sourceZipStream as `name` is what's used by convention to store the filename in Python IO streams. If there is, set `sourceZipStreamFileName` to its value, falling back to checking for FileNamedBytesIO, then falling back to `/POSTupload.zip`. 

This means API users setting name on their stream (for example because they have open()ed a zip from disk and passed it to Session.run()) will have things just work. However, also document this is what will happen on Session.run and why.

#### Steps to Test

Submit a known valid `.xbri` file directly to the Python API (just open() it) and it will get a rpe: error

After this, it should not as open sets .name

**review**:
@Arelle/arelle

### DIFF
--- a/arelle/api/Session.py
+++ b/arelle/api/Session.py
@@ -113,7 +113,6 @@ class Session:
     ) -> bool:
         """
         Perform a run using the given options.
-        
         When using a sourceZipStream, the name attribute on it will be checked
         and, if present, assumed to be the effective name of the zip file the
         stream represents. When validating '.xbri' files, via a sourceZipStream,


arelle/api/Session.py
```diff
@@ -113,6 +113,13 @@ def run(
     ) -> bool:
         """
         Perform a run using the given options.
+
```
austinmatherne-wk: Linter doesn't like the whitespace:

```suggestion

```

PR #1982, opened 2025-10-16
EDINET.EC8023W

aaroncameron-wk:
#### Reason for change
EDINET.EC8023W: In IXBRL files, 'nonFraction' elements should be immediately preceded by '△' if and only if the sign attribute is '-'.

#### Steps to Test
CI

**review**:
@Arelle/arelle

### DIFF
--- a/arelle/plugin/validate/EDINET/rules/upload.py
+++ b/arelle/plugin/validate/EDINET/rules/upload.py
@@ -1139,9 +1139,7 @@ def rule_EC8023W(
         if not pluginData.isExtensionUri(doc.uri, val.modelXbrl):
             continue
         nsmap = {
-            key: val
-            for key, val in doc.xmlRootElement.nsmap.items()
-            if key
+            'ix': XbrlConst.ixbrl
         }
         for elt in doc.xmlRootElement.xpath(negativeFactXpath, namespaces=nsmap):
             if elt.qname.namespaceURI == pluginData.jpigpNamespace:


arelle/plugin/validate/EDINET/rules/upload.py
```diff
@@ -1098,6 +1099,73 @@ def rule_EC1031E(
...
+    """
+    for doc in val.modelXbrl.urlDocs.values():
+        if doc.type != ModelDocumentType.INLINEXBRL:
+            continue
+        if not pluginData.isExtensionUri(doc.uri, val.modelXbrl):
+            continue
+        nsmap = {
+            key: val
+            for key, val in doc.xmlRootElement.nsmap.items()
+            if key
```
austinmatherne-wk: Out of curiosity, why do default namespaces need to be filtered out here?

arelle/plugin/validate/EDINET/rules/upload.py
```diff
@@ -1098,6 +1099,73 @@ def rule_EC1031E(
...
+    """
+    EDINET.EC8023W: In IXBRL files, 'nonFraction' elements should be immediately preceded by
+    '△' if and only if the sign attribute is '-'.
+
+    * Tagging using International Financial Reporting Standards taxonomy elements is not checked.
+    * Tagging using Japanese GAAP notes or IFRS financial statement filer-specific additional elements
+        may be identified as an exception and a warning displayed, even if the data content is correct.
+    """
+    negativeFactXpath = """
+    //ix:nonFraction[@sign="-"][
```
austinmatherne-wk, line 1121: Does this still work if the document uses non-standard namespace prefixes (something other than ix)? Or would we be better off defining our own namespace map for our usage of the ix prefix instead of using what's in the document?

arelle/plugin/validate/EDINET/rules/upload.py
```diff
@@ -1098,6 +1099,71 @@ def rule_EC1031E(
...
+        (preceding-sibling::text()[1] and normalize-space(preceding-sibling::text()[1]) != '△')
```
brettkail-wk, line 1122: Does this reject `<div>text △<ix:nonNumeric>123</ix:nonNumeric></div>`?  It seems like that should be allowed.

Does this reject `<div><span>△</span><ix:nonNumeric>123</ix:nonNumeric></div>`?  This is less clear, but it seems like it should be allowed too.

aaroncameron-wk: 1. `<div>text △<ix:nonFraction>5,461</ix:nonFraction></div>`
    1. Invalid because `△` precedes a positive value fact
1. `<div>text △<ix:nonFraction sign="-">5,462</ix:nonFraction></div>`
    1. Valid because `△` precedes a negative value fact
1. `<div><span>△</span><ix:nonFraction>5,463</ix:nonFraction></div>`
    1. Invalid because `△` precedes a positive value fact
1. `<div><span>△</span><ix:nonFraction sign="-">5,464</ix:nonFraction></div>`
    1. Valid because `△` precedes a negative value fact

In both of your examples, there's no `sign="-"` attribute, so `△` (used as a negative sign in Japanese financials) preceding it seems incorrect. If you left `sign="-"` out for simplification, then yes those are valid examples and this implementation does allow those.

brettkail-wk: > If you left `sign="-"` out for simplification

Sorry, yes I accidentally omitted it (I must have lost it while editing the comment).

> those are valid examples and this implementation does allow those.

Out of curiosity, how?
- I would expect `preceding-sibling::text()` to only match preceding text nodes, not preceding element nodes that contain text subelements.
- I would expect `text()` to return `"text △"`, which would mismatch.  It seems like we need "ends with"?

aaroncameron-wk: I've realized I had a misunderstanding of the xpath, masked by a logical misunderstanding when I was testing it. All passed over because the testcase I built was the only scenario where this worked. I'll revisit.

aaroncameron-wk: #2004

PR #1985, opened 2025-10-16
EDINET.EC5700W.GFM.1.4.8

allenbross-wf:
#### Description of change

GFM.1.4.8: Correct the priority attribute value so that it is less than 10.

#### Steps to Test

CI

**review**:
@Arelle/arelle


arelle/plugin/validate/EDINET/rules/gfm.py
```diff
@@ -1289,6 +1288,38 @@ def rule_gfm_1_4_4(
...
+
+    for modelDocument in val.modelXbrl.urlDocs.values():
+        if pluginData.isStandardTaxonomyUrl(modelDocument.uri, val.modelXbrl):
+            continue
+        rootElt = modelDocument.xmlRootElement
+        ns = {'xlink': 'http://www.w3.org/1999/xlink'}
+        for elt in rootElt.xpath('//*[@xlink:type="arc"]', namespaces=ns):
+            if elt.get("priority") is not None:
+                priority = elt.get("priority")
+                if int(priority) >= 10:
```
austinmatherne-wk: We should guard against type/value errors on the int conversion.

nit: it's also slightly more efficient to filter in xpath/c code for elements with a priority attribute.


```suggestion
        for elt in rootElt.xpath('//*[@xlink:type="arc"][@priority]', namespaces=ns):
            priority = elt.get("priority")
            with contextlib.suppress(ValueError, TypeError):
                if int(priority) >= 10:
```

austinmatherne-wk (approved):

PR #1991, opened 2025-10-21
EDINET.EC8030W

aaroncameron-wk:
#### Reason for change
EDINET.EC8030W: Any concept (other than DEI concepts) used in an instance must be in the presentation linkbase.

#### Steps to Test
CI

**review**:
@Arelle/arelle

### DIFF
--- a/arelle/plugin/validate/EDINET/rules/edinet.py
+++ b/arelle/plugin/validate/EDINET/rules/edinet.py
@@ -665,8 +665,6 @@ def rule_EC8030W(
     """
     usedConcepts = pluginData.getUsedConcepts(val.modelXbrl)
     relSet = val.modelXbrl.relationshipSet(tuple(LinkbaseType.PRESENTATION.getArcroles()))
-    if relSet is None:
-        return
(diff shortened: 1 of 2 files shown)


arelle/plugin/validate/EDINET/rules/edinet.py
```diff
@@ -649,6 +649,42 @@ def rule_EC8029W(
...
+        **kwargs: Any,
+) -> Iterable[Validation]:
+    """
+    EDINET.EC8030W: Any concept (other than DEI concepts) used in an instance must
+    be in the presentation linkbase.
+    """
+    usedConcepts = pluginData.getUsedConcepts(val.modelXbrl)
+    relSet = val.modelXbrl.relationshipSet(tuple(LinkbaseType.PRESENTATION.getArcroles()))
+    if relSet is None:
+        return
```
austinmatherne-wk: I don't think `relSet` can actually ever be `None` (it would return an empty relationship set), but if it could, wouldn't we want to raise an error on all tagged concepts?

PR #1994, opened 2025-10-22
Update generic dimension validations for HMRC

derekgengenbacher-wf:
#### Reason for change
- Updated the generic dimension validations based on the following [document](https://assets.publishing.service.gov.uk/media/681387ca7e56aaab3b525443/Generic-Dimension-Validations-4.9a.odt) from FRC.

#### Steps to Test
CI

**review**:
@Arelle/arelle


arelle/plugin/validate/UK/__init__.py
```diff
@@ -115,33 +121,6 @@
     }
 }
 
-GENERIC_DIMENSION_VALIDATION = {
-    # "taxonomyType": { "LocalName": (range of numbers if any, first item name, 2nd choice item name if any)
```
brettkail-wk, line 119: Is this format documented somewhere else, or should we move the comment?

PR #1997, opened 2025-10-24
EDINET.EC8060E

aaroncameron-wk:
#### Reason for change
EDINET.EC8060E: A context's scenario element must not have an explicit default member.

Note: In most cases, this is downgraded to a warning (EC8060W), and it was not trivial to generate a clean testcase for the error scenario. I left a note where we could configure a second testcase variation, but for now I am leaving it out.

#### Steps to Test
CI

**review**:
@Arelle/arelle

### DIFF
--- a/arelle/plugin/validate/EDINET/__init__.py
+++ b/arelle/plugin/validate/EDINET/__init__.py
@@ -21,7 +21,7 @@ DISCLOSURE_SYSTEM_VALIDATION_TYPE = "EDINET"
 RELEVELER_MAP: dict[str, dict[str, tuple[str, str | None]]] = {
     "ERROR": {
         # Re-code to EDINET version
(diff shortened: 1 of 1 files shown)


derekgengenbacher-wf (approved):

arelle/plugin/validate/EDINET/__init__.py
```diff
@@ -30,6 +30,8 @@
         "xbrl.5.2.5.2.1:zeroWeight": ("ERROR", None),
         # Silence, Duplicated by EDINET.EC5710W.FRTA.4.2.11
         "xbrl.5.1:emptyTargetNamespace": ("ERROR", None),
+        # Silence, Duplicated by EDINET.EC8060E
+        "xbrldie:DefaultValueUsedInInstanceError": ("ERROR", None),
```
austinmatherne-wk: I didn't see the others when they went in. But we shouldn't be silencing core specification errors. Certified XBRL validators are required to raise these.

PR #2003, opened 2025-10-28
Remove report package loading filters

austinmatherne-wk:
#### Reason for change
This code was part of an early draft that handled both report package loading and validation in a single class. It’s no longer needed and prevents expected errors such as `rpe:multipleReports` from being raised for .xbri reports when `--reportPackage` isn’t specified. Because those reports were loaded as plain zips, not report packages.

#### Description of change
Remove report package loading filters.

#### Steps to Test
* CI
* Validate an .xbri report with multiple reports raises a `rpe:multipleReports` error.

**review**:
@Arelle/arelle

### DIFF
--- a/arelle/CntlrWinMain.py
+++ b/arelle/CntlrWinMain.py
@@ -1209,7 +1209,7 @@ class CntlrWinMain (Cntlr.Cntlr):
                         pluginXbrlMethod(self, None, loadedModelXbrl)
                 if loadedModelXbrl.fileSource not in validatedFileSources:
                     validatedFileSources.add(loadedModelXbrl.fileSource)
-                    ValidateFileSource(self, loadedModelXbrl.fileSource).validate()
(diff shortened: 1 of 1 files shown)


arelle/packages/report/ReportPackage.py
```diff
@@ -217,13 +217,8 @@ def fromFileSource(filesource: FileSource) -> ReportPackage | None:
         reports = getAllReportEntries(filesource, stld)
         if reportPackageJsonFile is None and reports is None:
             return None
```
brettkail-wk, line 220: This fixes the `rpe:multipleReports` case I originally tried, but it seems to miss other cases where the suffix is `.xbr`/`.xbri` but you don't specify `--reportPackage`.  If you rename an SEC filing doc zip to `.xbri`, it seems you should get `rpe:invalidDirectoryStructure`, but instead it's silently accepted.  I also expected an `rpe:` error for a `.xbr`/`.xbri` with no zip entries, just a root directory, etc.

PR #2009, opened 2025-10-31
Add EC8073W

derekgengenbacher-wf:
#### Reason for change
- Added EC8073W

#### Steps to Test
- CI

**review**:
@Arelle/arelle

### DIFF
--- a/arelle/plugin/validate/EDINET/rules/edinet.py
+++ b/arelle/plugin/validate/EDINET/rules/edinet.py
@@ -926,7 +926,7 @@ def rule_EC8073W(
                 continue
             illegalChars = HALF_KANA.intersection(set(labelRel.toModelObject.textValue))
             if any(illegalChars):
-                yield Validation.error(
+                yield Validation.warning(
                     codes='EDINET.EC8073W',
                     msg=_("The concept: %(concept)s has a %(role)s label which contains characters that are not "
                           "allowed. Label: %(label)s, disallowed characters: %(characters)s"),


arelle/plugin/validate/EDINET/rules/edinet.py
```diff
@@ -888,12 +888,52 @@ def rule_EC8073E(
...
+    )
+    for concept in conceptsLabelsToCheck:
+        for labelRel in labelRelSet.fromModelObject(concept):
+            if (labelRel.toModelObject is None or
+                    labelRel.toModelObject.xmlLang not in JAPAN_LANGUAGE_CODES or
+                    labelRel.toModelObject.textValue is None):
+                continue
+            illegalChars = HALF_KANA.intersection(set(labelRel.toModelObject.textValue))
+            if any(illegalChars):
+                yield Validation.error(
```
aaroncameron-wk:
```suggestion
                yield Validation.warning(
```

arelle/plugin/validate/EDINET/rules/edinet.py
```diff
@@ -888,12 +888,52 @@ def rule_EC8073E(
...
-                    characters=illegalChars
+                    characters=list(illegalChars)
+                )
+
+
+@validation(
+    hook=ValidationHook.XBRL_FINALLY,
+    disclosureSystems=[DISCLOSURE_SYSTEM_EDINET],
+)
+def rule_EC8073W(
```
aaroncameron-wk, line 904: Could this just be rolled into EC8073E? It seems like it's a warning if it's related to `executiveOfficersMemberQn` but an error otherwise. Check out `EC8060E` for reference.

arelle/plugin/validate/EDINET/rules/edinet.py
derekgengenbacher-wf: You are joking me right?

arelle/plugin/validate/EDINET/rules/edinet.py
derekgengenbacher-wf: Not without making EC8073E more complicated. There is some overlap, but this one looks at `ExecutiveOfficersMember` and its descendants. So there is extra checking here.

PR #2012, opened 2025-11-03
EDINET.EC5000E

aaroncameron-wk:
#### Reason for change
EDINET.EC5000E: The encoding of the file must be UTF-8.

#### Description of change
Needed to make a change to core Arelle since `stripDeclarationBytes` was only working for UTF-8 encoded documents.

#### Steps to Test
CI

**review**:
@Arelle/arelle


derekgengenbacher-wf (approved):

arelle/plugin/validate/EDINET/rules/gfm.py
```diff
@@ -2145,6 +2144,20 @@ def rule_charsets(
         except Exception:
             pass
 
+        if modelDocument.type != ModelDocument.Type.INLINEXBRLDOCUMENTSET:
+            if xmlDeclaredEncoding is None or xmlDeclaredEncoding.lower() != 'utf-8':
```
brettkail-wk: Out of curiosity, why this `xmlDeclaredEncoding is None` check?  An XML file can be UTF-8 encoded even if it doesn't contain an explicit `encoding` ([XML 1.0 reference F.1](https://www.w3.org/TR/xml/#sec-guessing-no-ext-info)).  Does EDINET somewhere require explicit `encoding`?

arelle/FileSource.py
```diff
@@ -917,7 +917,11 @@ def openXmlFileStream(
             text = stripDeclarationText(text)
         return (FileNamedStringIO(filepath, initial_value=text), encoding)
 
-def stripDeclarationBytes(xml: bytes) -> bytes:
+def stripDeclarationBytes(xml: bytes, encoding: str | None) -> bytes:
+    if encoding is not None and not encoding.lower().startswith('utf-8'):
```
brettkail-wk, line 921: This change seems plausible, though I assume the extra encode/decode will slow down loading of non-UTF-8 files.

Out of curiosity, in what way does this change help?  I pre-existing don't understand the rationale for `stripDeclaration`, so I'm probably overlooking some subtlety.

arelle/plugin/validate/EDINET/rules/gfm.py
aaroncameron-wk: GFM.1.10.4 does require it for IXBRL files, but EC5000E does apply to XML/XSD files as well. I've updated this to check the actual document encoding and modified the testcase to include a UTF-8 file with no declaration to verify that it does not fire this error.

arelle/FileSource.py
aaroncameron-wk: It boils down to `lxml` quirks. The way Arelle parses XML through this pathway, it deals with UTF-8 internally, so it gets confused if it encounters a declaration indicating any other encoding. Removing the declaration solves this. `lxml` is capable of parsing UTF-16 files without stripping the declaration, but I suspect something about loading the XML file from a zip archive limits our options in that regard. 

This code was causing Arelle to fail to parse a non-UTF-8 doc at all. I suspect it hasn't been encountered because (a) UTF-8 is generally the default and (b) many use cases are rerouted to `ValidateFilingText` [here](https://github.com/Arelle/Arelle/blob/2186c0ae734c1f71d168665522a772077cf49e83/arelle/ModelDocument.py#L171-L177), which ultimately [strips the declaration as text](https://github.com/Arelle/Arelle/blob/2186c0ae734c1f71d168665522a772077cf49e83/arelle/ValidateFilingText.py#L480-L486).

Another solution I explored was to search for the correct encoding of `b"<?xml"` within the bytes to retain the byte-based search, but it proved non-trivial since Arelle does not distinguish between BE and LE, and ultimately the complexity didn't seem worth the performance gain for a relatively rare use case.

austinmatherne-wk (approved):

PR #2017, opened 2025-11-06
Add EDINET Required Role Rules

austinmatherne-wk:
#### Description of change
Implements EDINET rules:
- EDINET.EC8038W
- EDINET.EC8039W
- EDINET.EC8040W
- EDINET.EC8041W
- EDINET.EC8042W
- EDINET.EC8043W
- EDINET.EC8044W
- EDINET.EC8045W
- EDINET.EC8046W
- EDINET.EC8047W
- EDINET.EC8048W
- EDINET.EC8049W
- EDINET.EC8050W
- EDINET.EC8061W
- EDINET.EC8063W
- EDINET.EC8065W
- EDINET.EC8066W
- EDINET.EC8067W
- EDINET.EC8068W

Initially I was working on dimensionally validating facts against cubes within the roles, but that appears to not be needed for EDINET and not worth the processing overhead.

#### Steps to Test
* CI

**review**:
@Arelle/arelle

### DIFF
--- a/tests/integration_tests/validation/conformance_suite_configurations/edinet.py
+++ b/tests/integration_tests/validation/conformance_suite_configurations/edinet.py
@@ -19,18 +19,13 @@ VALID_EXPECTED_ERRORS = {
         "EDINET.EC5700W.GFM.1.8.4": 2,
         "EDINET.EC8023W": 2,
         "EDINET.EC8027W": 1,
-        "EDINET.EC8065W": 1,
-        "EDINET.EC8067W": 1,
     },
     "valid/index.xml:valid04": {
         "EDINET.EC2002W": 2,
         "EDINET.EC2005E": 1,
-        "EDINET.EC8050W": 1,
     },
     "valid/index.xml:valid05": {
         "EDINET.EC2002W": 1,
-        "EDINET.EC8043W": 1,
-        "EDINET.EC8045W": 1,
         # The original valid05.zip (unmodified from source sample) fires EDINET.EC5700W.GFM.1.1.3
         # The version checked into this repo has been modified to not fire that error.
     },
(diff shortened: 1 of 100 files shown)


arelle/plugin/validate/EDINET/ControllerPluginData.py
```diff
@@ -48,6 +48,7 @@ def __init__(self, name: str):
         self._tocBuilder = TableOfContentsBuilder()
         self._usedFilepaths = set()
         self._uploadContents = None
+        self.jpcrpNamespace = 'http://disclosure.edinet-fsa.go.jp/taxonomy/jpcrp/2024-11-01/jpcrp_cor'
```
derekgengenbacher-wf, line 51: This namespace already exists as part of `PluginValidationDataExtension.py`. Can we bring this over from that? If not perhaps those namespaces need to be defined in `EDINET/constants.py`?

.gitignore
```diff
@@ -30,6 +30,7 @@ arelle/plugin/validate/DQC_US_Rules
 arelle/plugin/validate/DQC_test
 arelle/plugin/validate/ESEF-DQC.py
 arelle/plugin/validate/eforms.py
+arelle/plugin/xbrlus
```
derekgengenbacher-wf, line 33: I understand what this is doing, but Im confused as to why this plugin is being added at all?

arelle/plugin/validate/EDINET/ControllerPluginData.py
austinmatherne-wk: It's not accessible from the hook where this is called. We could move to a constants file or create a plugin data base class that both of the edinet plugin data classes extend. Probably best to load these from a disclosure system config file. @aaroncameron-wk thoughts?

.gitignore
austinmatherne-wk: I can move this to a separate PR if desired (it's not required for EDINET). This came out of the recent plugin loading changes for the SEC/xule which allows loading plugins when cloning the entire XBRL US plugins repo into the plugin directory.

derekgengenbacher-wf: Its fine, I was just confused.

arelle/plugin/validate/EDINET/ControllerPluginData.py
aaroncameron-wk: Agreed, setting these in a disclosure system config file seems best.

tests/integration_tests/validation/conformance_suite_configurations/edinet.py
```diff
@@ -19,13 +19,18 @@
         "EDINET.EC5700W.GFM.1.8.4": 2,
         "EDINET.EC8023W": 2,
         "EDINET.EC8027W": 1,
+        "EDINET.EC8065W": 1,
+        "EDINET.EC8067W": 1,
```
aaroncameron-wk: Are these being placed here because of uncertainty about the implementation? That was the original plan for this configuration, although I wasn't consistent about that until recently. If I'm confident that my implementation is correct, I've been putting these into the testcase xml, even for the "valid" filings. If I have doubts, I've been putting them here so it's easier to clean up once there is more clarity around that particular rule.

arelle/plugin/validate/EDINET/ControllerPluginData.py
austinmatherne-wk: I'll create a follow up ticket to move these to a config file.

derekgengenbacher-wf (approved):

PR #2022, opened 2025-11-10
Fix URI validation refactor typo

austinmatherne-wk:
#### Reason for change
Typo was causing failures in the XBRL dimensions conformance suite. 

#### Description of change
Code was checking if `UrlUtil.isValidUriReference(value)` returned `None`, which it never does.

#### Steps to Test
* CI

**review**:
@Arelle/arelle

### DIFF
--- a/tests/integration_tests/validation/conformance_suite_configurations/xbrl_dimensions_1_0.py
+++ b/tests/integration_tests/validation/conformance_suite_configurations/xbrl_dimensions_1_0.py
@@ -6,7 +6,7 @@ config = ConformanceSuiteConfig(
         ConformanceSuiteAssetConfig.conformance_suite(
             Path('XBRL-XDT-CONF-2025-09-09.zip'),
             entry_point=Path('XBRL-XDT-CONF-2025-09-09/xdt.xml'),
-            public_download_url='https://www.xbrl.org/2009/xdt-conf-cr4-2009-10-06.zip',
+            public_download_url='https://www.xbrl.org/2025/XBRL-XDT-CONF-2025-09-09.zip',
             source=AssetSource.S3_PUBLIC,
         ),
     ],


arelle/XmlValidate.py
```diff
@@ -458,9 +458,8 @@ def validateValue(
                     xValue = sValue = value
                     xValid = VALID_ID
                 elif baseXsdType == "anyURI":
-                    if value:  # allow empty strings to be valid anyURIs
-                        if UrlUtil.isValidUriReference(value) is None:
-                            raise ValueError("IETF RFC 2396 4.3 syntax")
+                    if not UrlUtil.isValidUriReference(value):
```
brettkail-wk, line 461: Out of curiosity, why is this change correct?  It looks like the [value space](https://www.w3.org/TR/xmlschema11-2/#anyURI-vs) is ~any string.  It looks like the code should just be removed.

Dimension spec [section 2.4.3](https://www.xbrl.org/specification/dimensions/rec-2012-01-25/dimensions-rec-2006-09-18+corrected-errata-2012-01-25-clean.html#sec-constraints-on-xbrldt-targetRole) has a bullet that says:
> According to the xbrldt XML Schema, the content of the @xbrldt:targetRole attribute MUST be a valid URI.

...but the xbrldt XML Schema doesn't require that.  Are XDT tests requiring URI validation?

Anyway, my primary worry is that this change breaks valid documents that are assuming anyURI accepts any string, and my secondary worry is that it rejects valid URIs (I'm not confident in what grammar `isValidUriReference` is validating).

austinmatherne-wk: > Are XDT tests requiring URI validation?

I went down a similar rabbit hole and also believe the spec is wrong, but the four dimension conformance suite tests I removed from the expected failures config require it. Additionally, I'm told that other XML validators such as Xerces do raise schema errors.

For reference, #200 is what broke the four tests. In an unfortunate series of timing that PR went in right before we included the dimensions spec in our CI, so the breaking change wasn't identified and the failing tests were wrongly assumed to have been a long standing issue.

brettkail-wk: Thanks, that context helps.  The XSD spec sure seems clear that you shouldn't validate:
> Because it is impractical for processors to check that a value is a context-appropriate URI reference, neither the syntactic constraints defined by the definitions of individual schemes nor the generic syntactic constraints defined by [[RFC 3987]](https://www.w3.org/TR/xmlschema11-2/#RFC3987) and [[RFC 3986]](https://www.w3.org/TR/xmlschema11-2/#RFC3986) and their successors are part of this datatype as defined here.

But I guess if Xerces does validate and Arelle used to, it's reasonable to restore the validation.  I'm still not sure what `isValidUriReference` is actually enforcing, but I guess the comment above `relativeUrlPattern` already says that.

PR #2023, opened 2025-11-11
Fix loadFromExcel Plugin

### LINKED ISSUE #2015: loadFromExcel plugin returns an "[error] No XBRL entry points could be preparing entrypoints from provided file in 0.00 seconds"
### What happened?

I’m getting this error when using the plug-in `loadFromExcel` to generate a taxonomy.
"[error] No XBRL entry points could be preparing entrypoints from provided file in 0.00 seconds"

The same Excel template worked fine with older versions; previously, I had used it in November 2024.

A new Excel template created using saveLoadableExcel gives the same error.

### Documents

[XII_Taxonomy_25-24_meta_data.xlsx](https://github.com/user-attachments/files/23362103/XII_Taxonomy_25-24_meta_data.xlsx)

_No response_

### Interface

GUI (arelleGUI)

### Version

2.37.69

### Download

arelle.org (bundled Python)

### Operating System

Windows

austinmatherne-wk:
#### Description of change
Resolves #2015. Recent changes broke the loadFromExcel plugin.

#### Steps to Test
* CI

**review**:
@Arelle/arelle

### DIFF
--- a/tests/integration_tests/scripts/run_scripts.py
+++ b/tests/integration_tests/scripts/run_scripts.py
@@ -81,7 +81,7 @@ def run_script_options(options: Namespace) -> list[ParameterSet]:
         if options.download_cache:
             download_and_apply_cache(
                 ALL_SCRIPTS_ZIP,
-                version_id='1gQswj8EEmCO1eV9OwC0lga34WB03Pb0'
+                version_id='CNTq_CLLvVEpcpxw9x4ipF76gD7zvZWD'
             )
     elif options.all_frozen_builds:
         scripts = [s for s in all_scripts if _for_frozen_build(s)]
--- a/tests/integration_tests/scripts/tests/load_from_excel.py
+++ b/tests/integration_tests/scripts/tests/load_from_excel.py
@@ -1,0 +1,55 @@
+from __future__ import annotations
+
+import os
+import urllib.request
+from pathlib import Path
+
+from tests.integration_tests.integration_test_util import get_s3_uri
+from tests.integration_tests.scripts.script_util import (
+    assert_result,
+    parse_args,
+    prepare_logfile,
+    run_arelle,
+    validate_log_file,
+)
+
+errors = []
+this_file = Path(__file__)
+args = parse_args(
+    this_file.stem,
+    "Smoke test load from Excel plugin.",
+    cache='load_from_excel.zip',
+    cache_version_id='oGRKSXanaohN6PKnXckFFpzNPhIr.Yq9',
+)
+arelle_command = args.arelle
+arelle_offline = args.offline
+working_directory = Path(args.working_directory)
+test_directory = Path(args.test_directory)
+arelle_log_file = prepare_logfile(test_directory, this_file)
+
+excel_taxonomy_path = test_directory / 'XII_Taxonomy.xlsx'
+excel_taxonomy_url = get_s3_uri(
+    'ci/packages/XII_Taxonomy.xlsx',
+    version_id='c5mjDshdqpmu8CvVt37Ur9yhh06fO1BB'
+)
+
+print(f"Downloading Excel file: {excel_taxonomy_url}")
+urllib.request.urlretrieve(excel_taxonomy_url, excel_taxonomy_path)
+
+run_arelle(
+    arelle_command,
+    additional_args=[
+        "--file",
+        excel_taxonomy_path,
+    ],
+    offline=arelle_offline,
+    logFile=arelle_log_file,
+    plugins=["loadFromExcel"],
+)
+
+print(f"Checking for log errors: {arelle_log_file}")
+errors += validate_log_file(arelle_log_file)
+assert_result(errors)
+
+print("Cleaning up")
+os.unlink(excel_taxonomy_path)


arelle/packages/report/ReportPackage.py
```diff
@@ -215,8 +215,15 @@ def fromFileSource(filesource: FileSource) -> ReportPackage | None:
...
         reports = getAllReportEntries(filesource, stld)
-        if reportType == Const.ReportType.UNCONSTRAINED_REPORT_PACKAGE and reportPackageJsonFile is None and reports is None:
-            # A zip file, but not a report package.
+        reportPackageSpecificFileType = reportType is not None and reportType != Const.ReportType.UNCONSTRAINED_REPORT_PACKAGE
+        containsReportPackageFile = reportPackageJsonFile is not None or reports is not None
+        if not reportPackageSpecificFileType and not containsReportPackageFile:
+            # The report package conformance suite includes validation requirements for how a future report package file
+            # extension should be handled, requiring us to parse archives with:
+            # 1. a known report package file extension and containing a recognized report package file.
+            # 2. a known report package specific file extension, but no recognized report package files.
```
brettkail-wk: What condition is the word `specific` differentiating between `2.` and `1.`?

arelle/packages/report/ReportPackage.py
```diff
@@ -215,8 +215,15 @@ def fromFileSource(filesource: FileSource) -> ReportPackage | None:
         if reportPackageJsonFile:
             reportPackageJson = getPackageJson(filesource, reportPackageJsonFile)
         reports = getAllReportEntries(filesource, stld)
-        if reportType == Const.ReportType.UNCONSTRAINED_REPORT_PACKAGE and reportPackageJsonFile is None and reports is None:
-            # A zip file, but not a report package.
+        reportPackageSpecificFileType = reportType is not None and reportType != Const.ReportType.UNCONSTRAINED_REPORT_PACKAGE
+        containsReportPackageFile = reportPackageJsonFile is not None or reports is not None
+        if not reportPackageSpecificFileType and not containsReportPackageFile:
```
brettkail-wk: The `containsReportPackageFile` var name seems confusing with the `reports is not None` condition.

It seems wrong to check `reportPackage.json` or `reports/` at all when `reportType is None`.  It seems like the report package spec only talks about `.zip`/`.ZIP`.  It seems like we should exit much earlier with `if reportType is None: return None` and then leave this condition as-is?

arelle/packages/report/ReportPackage.py
austinmatherne-wk: I was referring to `.xbri` and `.xbr` as opposed to `.zip/.ZIP`. Hopefully the updated code and comments are more clear.

arelle/packages/report/ReportPackage.py
austinmatherne-wk: Beyond `.xbr`, `.xbri`, `.zip`, and `.ZIP`, the spec and conformance suite require validation for report packages with unsupported file extensions.

If we `return None` when the archive contents indicate it's a report package, loading falls back to generic archive parsing and entry detection. This seems undesirable. We also still have to apply report package validation to the filesource to raise the unsupported file extension error.

I've updated the code to block loading reports from these packages (the unsupported report package file extension error is still raised). This prevents unhelpful cascading errors since we can't reliably determine how to load these files (they could be future report package formats that this version of Arelle doesn't support).

Open to feedback if this seems like the wrong tradeoff.

brettkail-wk: > Beyond `.xbr`, `.xbri`, `.zip`, and `.ZIP`, the spec and conformance suite require validation for report packages with unsupported file extensions.

Where are you seeing that?  Perhaps I'm misunderstanding, but I would expect [`rpe:unsupportedFileExtension`](https://www.xbrl.org/Specification/report-package/REC-2023-09-22+corrected-errata-2025-03-11/report-package-REC-2023-09-22+corrected-errata-2025-03-11.html#error-rpe-unsupportedfileextension) for any other extension.

> If we return None when the archive contents indicate it's a report package, loading falls back to generic archive parsing and entry detection.

It seems like Arelle should only do fallbacks if `--reportPackage` is unspecified.

My intuition is that `ReportPackage.fromFileSource` should implement all the spec-required behaviors by default, and if Arelle "UI" (cmdline or GUI) needs different default behavior, then it seems like we should add some `strict=True` option (caller specified via `--reportPackage`?).  I'm getting lost on whether `return None` is the right way to do that, so perhaps what I meant to say is we need something like `if reportType is None and not strict: return None` (i.e., if we're strict, then keep going and give an error about invalid report package, but otherwise let other fallbacks kick in).

austinmatherne-wk: > Where are you seeing that? Perhaps I'm misunderstanding, but I would expect [rpe:unsupportedFileExtension](https://www.xbrl.org/Specification/report-package/REC-2023-09-22+corrected-errata-2025-03-11/report-package-REC-2023-09-22+corrected-errata-2025-03-11.html#error-rpe-unsupportedfileextension) for any other extension.

We're on the same page. The spec requires that files with other file extensions be validated to trigger an `rpe:unsupportedFileExtension` validation error.

> It seems like Arelle should only do fallbacks if --reportPackage is unspecified.

I agree this would be a cleaner approach, but it runs into a practical issue with our conformance suite architecture. Currently, when running the report package conformance suite, the `--reportPackage` flag is passed to the top-level suite zip and propagates down to individual test cases. Since the conformance suite zip itself isn't a report package, implementing this change would prevent any test cases from running.

This is definitely worth addressing, but it would require refactoring how we handle conformance suite execution. I think we should wait for Aaron's conformance suite test runner refactor to land first, which should give us better control over test case execution. We can revisit this change as part of that effort.

PR #2031, opened 2025-11-13
Pass Runtime Options JSON or File via command line argument

aaroncameron-wk:
#### Reason for change
Quickly convert a JSON representation of a RuntimeOptions object into an executable configuration file.

#### Description of change
Provide a path to a JSON file via `--optionsFile`. This can be combined with other CLI arguments, which will override the JSON values.

#### Steps to Test
The testcase I've chosen happens to fail validation with `'utf-8' codec can't decode byte 0x8a`.

JSON file ([runtime_options.json](https://github.com/user-attachments/files/23535774/runtime_options.json))

Basic:
```
python arelleCmdLine.py --optionsFile runtime_options.json
```

Override an option from the file:
```
python arelleCmdLine.py --optionsFile runtime_options.json -f
"tests/resources/conformance_suites/uksef-conformance-suite-v2.0/uksef-conformance-suite-v2.0/uksef-conformance-suite-v2.0.zip/uksef-conformance-suite/tests/FRC/FRC_18/TC1_valid.zip"
```

**review**:
@Arelle/arelle

### DIFF
--- a/arelle/CntlrCmdLine.py
+++ b/arelle/CntlrCmdLine.py
@@ -11,6 +11,7 @@ import datetime
 import fnmatch
 import gettext
 import glob
+import json
 import logging
 import multiprocessing
 import os
@@ -99,6 +100,7 @@ def parseAndRun(args):
 
 PREPARSE_ARG_CONFIGS = frozenset([
     (re.compile(r'^--plugins?.*$'), 'plugins'),
+    (re.compile(r'^--options(F|f)ile.*$'), 'optionsFile'),
 ])
 
 
@@ -445,17 +447,23 @@ def parseArgs(args):
     preparsedArgs = preparseArgs(args, parser)
 
     preloadPlugins = []
+    optionsFile = preparsedArgs.get('optionsFile')
+    optionsFileOptions = {}
+    if optionsFile:
+        optionsFileOptions = _parseOptionsFile(optionsFile, parser)
+        preloadPlugins.extend(optionsFileOptions.get('plugins', '').split('|'))
+
     preloadPlugins.extend(preparsedArgs.get('plugins', '').split('|'))
 
     # install any dynamic plugins so their command line options can be parsed if present
     arellePluginModules = {}
     for pluginCmd in preloadPlugins:
-                cmd = pluginCmd.strip()
-                if cmd not in ("show", "temp") and len(cmd) > 0 and cmd[0] not in ('-', '~', '+'):
-                    moduleInfo = PluginManager.addPluginModule(cmd)
-                    if moduleInfo:
-                        arellePluginModules[cmd] = moduleInfo
-                        PluginManager.reset()
+        cmd = pluginCmd.strip()
+        if cmd not in ("show", "temp") and len(cmd) > 0 and cmd[0] not in ('-', '~', '+'):
+            moduleInfo = PluginManager.addPluginModule(cmd)
+            if moduleInfo:
+                arellePluginModules[cmd] = moduleInfo
+                PluginManager.reset()
 
     # add plug-in options
     for optionsExtender in PluginManager.pluginClassMethods("CntlrCmdLine.Options"):
@@ -467,6 +475,10 @@ def parseArgs(args):
                       help=_("Show product version, copyright, and license."))
     parser.add_option("--diagnostics", action="store_true", dest="diagnostics",
                       help=_("output system diagnostics information"))
+    parser.add_option("--optionsFile", "--optionsfile",
+                      action="store", dest="optionsFile",
+                      help=_("Provide a path to a JSON file containing runtime options. "
+                             "These options will be overridden by any command line options provided."))
 
     if not args and isGAE():
         args = ["--webserver=::gae"]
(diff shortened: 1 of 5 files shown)


arelle/CntlrCmdLine.py
```diff
@@ -626,6 +639,27 @@ def _pluginHasCliOptions(moduleInfo):
...
+        return json.loads(arg)
+    except ValueError as e:
+        # Argument is not valid JSON
+        pass
+    if os.path.exists(arg):
+        try:
+            with open(arg) as f:
+                return json.load(f)
+        except ValueError as e:
+            # File does not contain valid JSON
```
brettkail-wk: It seems like it would be preferable to print the actual exception (which probably has line/col info?) rather than the vague "Options argument is not a valid JSON or a path to a valid JSON file.".

Related, suggest `if json.startswith('{'): return json.loads else: with open(arg) as f: return json.load(f)` so that whether or not a file is opened is deterministic rather than sending a long/broken JSON string to a the OS to open as a filename.  (It might be even better to just have separate options for literal vs file.)

arelle/CntlrCmdLine.py
```diff
@@ -518,6 +522,14 @@ def parseArgs(args):
...
             pluginOptionDestinations.add(groupOption.dest)
     baseOptions = {}
+    if options.optionsJson:
+        # If provided, options JSON sets the initial base options.
+        # These will be overwritten if options are also provided by another source.
+        jsonOptions = _parseRuntimeOptionsArg(options.optionsJson)
+        if jsonOptions is None or not isinstance(jsonOptions, dict):
+            parser.error(_("Options argument is not valid JSON or a path to a valid JSON file."))
+        else:
+            baseOptions.update(_parseRuntimeOptionsArg(options.optionsJson))
```
austinmatherne-wk: Could you help me understand the use case this addresses? I want to make sure we're solving the right problem. I have a few concerns:
* Making a typo in a JSON option key will trigger an unhandled exception that logs a stacktrace
* Passing JSON from the CLI for the iXDS plugin and `--file` argument has caused issues for Windows users
* `RuntimeOptions.optionsJson` could be confusing (users might think it's a full serialization of `RuntimeOptions`)

aaroncameron-wk: The test engine I'm working on logs the runtime options passed to Arelle for each testcase variation. Being able to copy/paste those runtime options to quickly recreate the individual testcase variation run is a nice convenience, rather than needing to convert those options to CLI args.

> Making a typo in a JSON option key will trigger an unhandled exception that logs a stacktrace

Good point. I'll make some updates to have have more user-friendly error handling.

> Passing JSON from the CLI for the iXDS plugin and --file argument has caused issues for Windows users

I don't mind limiting this feature to passing a JSON file rather than directly in the argument.

> RuntimeOptions.optionsJson could be confusing (users might think it's a full serialization of RuntimeOptions)

If I limit the feature to only JSON files, I'll rename this to `optionsFile` or similar and it should avoid that confusion.

arelle/CntlrCmdLine.py
aaroncameron-wk: I think my changes based on [this comment](https://github.com/Arelle/Arelle/pull/2031#discussion_r2528417432) will also address your concerns.

arelle/CntlrCmdLine.py
austinmatherne-wk: Limiting to json files with key validation addresses my concerns. Also, using these json files to handle user defaults (internet connectivity, etc.) is generally useful 👍

arelle/CntlrCmdLine.py
brettkail-wk: Yes, those changes look plausible to me, thanks.

tests/integration_tests/scripts/tests/options_file.py
```diff
@@ -29,11 +30,14 @@
...
 
+def _escape_json_path(path: Path) -> str:
+    return json.dumps(str(path))[1:-1]
+
 test_cases: list[tuple[str, str, list[str], dict[regex.Pattern[str], int]]] = [
     # Entry point only set in options file
     (
         'options_only',
-        f'{{ "validate": true, "entrypointFile": "{instance_files[0]}" }}',
+        f'{{ "validate": true, "entrypointFile": "{_escape_json_path(instance_files[0])}" }}',
```
brettkail-wk: Using an f-string to build JSON seems fragile with needing to remember to escape strings.  It seems preferable to just use `json.dumps` on the whole object:
```
json.dumps({ 'validate': True, 'entrypointFile': instance_files[0] }),
```
Or perhaps even better, `dict[str, Any]` and `f.write(json.dumps(options))` below.

If that's unacceptable for some reason, it seems preferable to escape the whole string rather than string manipulation on the result of `json.dumps`:
```
f'{{ "validate": true, "entrypointFile": {json.dumps(instance_files[0])} }}',
```

arelle/CntlrCmdLine.py
```diff
@@ -625,6 +648,27 @@ def _pluginHasCliOptions(moduleInfo):
...
+    """
+    if not os.path.exists(optionsFile):
+        parser.error(_("Options file path does not exist: {}").format(optionsFile))
+        return {}
+    try:
+        with open(optionsFile) as f:
+            jsonOptions = json.load(f)
+    except Exception as e:
+        parser.error(_("Unable to parse options JSON file: {}").format(e))
+        return {}
```
austinmatherne-wk: nit: avoid TOCTOU by catching OSError instead of using a separate exists check.

PR #2032, opened 2025-11-14
Use StartDateForPeriodCoveredByReport and EndDateForPeriodCoveredByReport for context checking for mandatory facts

derekgengenbacher-wf:
#### Description of change
Update mandatory fact check for `validate/UK` to use `StartDateForPeriodCoveredByReport` and `EndDateForPeriodCoveredByReport` for context date checking.

#### Steps to Test
CI

**review**:
@Arelle/arelle

### DIFF
--- a/arelle/plugin/validate/UK/__init__.py
+++ b/arelle/plugin/validate/UK/__init__.py
@@ -330,7 +330,7 @@ def validateXbrlFinally(val, *args, **kwargs):
             startDate = None
             endDate = None
             _missingItems = []
-            for fact in mandatoryFacts.get('EndDateForPeriodCoveredByReport'):
+            for fact in mandatoryFacts.get('EndDateForPeriodCoveredByReport', set()):
                 if (fact is not None and
                         fact.xValid >= VALID and
                         fact.context is not None and
@@ -338,10 +338,10 @@ def validateXbrlFinally(val, *args, **kwargs):
                         dateUnionEqual(fact.context.instantDate, fact.xValue)):
                     endDate = fact.xValue
                     break
-            if not endDate:
+            else:
                 _missingItems.append('EndDateForPeriodCoveredByReport')
 
-            for fact in mandatoryFacts.get('StartDateForPeriodCoveredByReport'):
+            for fact in mandatoryFacts.get('StartDateForPeriodCoveredByReport', set()):
                 if (fact is not None and
                         fact.xValid >= VALID and
                         fact.context is not None and
@@ -349,7 +349,7 @@ def validateXbrlFinally(val, *args, **kwargs):
                         dateUnionEqual(fact.context.instantDate, endDate)):
                     startDate = fact.xValue
                     break
-            if not startDate:
+            else:
                 _missingItems.append('StartDateForPeriodCoveredByReport')
 
             if startDate is not None and endDate is not None:
@@ -364,7 +364,8 @@ def validateXbrlFinally(val, *args, **kwargs):
                     for fact in mandatoryFacts.get(mandatoryConcept, set()):
                         if (fact is not None and fact.context is not None and fact.xValid >= VALID and
                                 ((fact.context.isInstantPeriod and dateUnionEqual(fact.context.instantDate, endDate) or
-                                  (fact.context.isStartEndPeriod and dateUnionEqual(fact.context.startDatetime, startDate) and dateUnionEqual(fact.context.endDate, endDate))))):
+                                  (fact.context.isStartEndPeriod and dateUnionEqual(fact.context.startDatetime, startDate) and
+                                   dateUnionEqual(fact.context.endDate, endDate, instantEndDate=True))))):
                                 foundFact = True
                                 break
                     if not foundFact:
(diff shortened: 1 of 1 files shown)


arelle/plugin/validate/UK/__init__.py
```diff
@@ -326,17 +327,68 @@ def checkFacts(facts):
         checkFacts(modelXbrl.facts)
 
         if val.isAccounts:
-            _missingItems = MANDATORY_ITEMS[val.txmyType] - mandatoryFacts.keys()
-            if hasCompaniesHouseContext and "UKCompaniesHouseRegisteredNumber" not in mandatoryFacts:
-                _missingItems.add("UKCompaniesHouseRegisteredNumber")
+            startDate = None
+            endDate = None
+            _missingItems = []
+            for fact in mandatoryFacts.get('EndDateForPeriodCoveredByReport'):
```
stevenbronson-wk: This will crash if the key is missing.

arelle/plugin/validate/UK/__init__.py
```diff
@@ -326,17 +327,68 @@ def checkFacts(facts):
...
+                if (fact is not None and
+                        fact.xValid >= VALID and
+                        fact.context is not None and
+                        fact.context.isInstantPeriod and
+                        dateUnionEqual(fact.context.instantDate, fact.xValue)):
+                    endDate = fact.xValue
+                    break
+            if not endDate:
```
stevenbronson-wk: Could use "else" here and below.

arelle/plugin/validate/UK/__init__.py
```diff
@@ -326,17 +327,68 @@ def checkFacts(facts):
...
+                _missingItems.append('EndDateForPeriodCoveredByReport')
```
stevenbronson-wk, line 342: Different validation message?  Probably confusing to see "fact missing" because the date doesn't match the fact value.

arelle/plugin/validate/UK/__init__.py
```diff
@@ -326,17 +327,68 @@ def checkFacts(facts):
...
+                        fact.xValid >= VALID and
+                        fact.context is not None and
+                        fact.context.isInstantPeriod and
+                        dateUnionEqual(fact.context.instantDate, endDate)):
+                    startDate = fact.xValue
+                    break
+            if not startDate:
+                _missingItems.append('StartDateForPeriodCoveredByReport')
+
+            if startDate is not None and endDate is not None:
```
stevenbronson-wk, line 355: A bit odd to switch from `if not startDate` to `startDate is [not] None`.

arelle/plugin/validate/UK/__init__.py
```diff
@@ -326,17 +327,68 @@ def checkFacts(facts):
...
+                    mandatoryConceptsToCheck = MANDATORY_ITEMS[val.txmyType] | {"UKCompaniesHouseRegisteredNumber"}
+
+                for mandatoryConcept in mandatoryConceptsToCheck:
+                    if mandatoryConcept in ['StartDateForPeriodCoveredByReport', 'EndDateForPeriodCoveredByReport']:
+                        continue
+                    foundFact = False
+                    for fact in mandatoryFacts.get(mandatoryConcept, set()):
+                        if (fact is not None and fact.context is not None and fact.xValid >= VALID and
+                                ((fact.context.isInstantPeriod and dateUnionEqual(fact.context.instantDate, endDate) or
+                                  (fact.context.isStartEndPeriod and dateUnionEqual(fact.context.startDatetime, startDate) and dateUnionEqual(fact.context.endDate, endDate))))):
```
stevenbronson-wk: Differs a bit from https://github.com/Arelle/Arelle/blob/f47b1c6ca315a7a04d55da956219bd369c43a007/arelle/ModelXbrl.py#L597-L598
(.endDate, instantEndDate)

arelle/plugin/validate/UK/__init__.py
```diff
@@ -326,17 +327,68 @@ def checkFacts(facts):
...
             if _missingItems:
                 modelXbrl.error("JFCVC.3312",
                     _("Facts are MANDATORY: %(missingItems)s"),
                     modelObject=modelXbrl, missingItems=", ".join(sorted(_missingItems)))
-            if not atLeastOneFacts and val.txmyType in MUST_HAVE_ONE_ITEM:
-                modelXbrl.error("JFCVC.3312.atLeastOne",
-                                _("At least one of the facts is MANDATORY: %(missingItems)s"),
-                                modelObject=modelXbrl, missingItems=", ".join(sorted(MUST_HAVE_ONE_ITEM[val.txmyType])))
+
+            if val.txmyType in MUST_HAVE_ONE_ITEM:
```
stevenbronson-wk: We skip the other check if we don't find start/end date.  Should we do the same here?

arelle/plugin/validate/UK/__init__.py
derekgengenbacher-wf: I think you are saying that this could be bumped in underneath the for loop? So my argument against that will be with that in mind. It seems to me that if we bump this `if` into the for loop and make it an `else`, then we are making an assumption that there will only ever be 0 or 1 fact tagged with the mandatory date concepts. While this may be the case it seems like we should handle it as if there may be many of these facts tagged where only one aligns with the mandatory rule check?

austinmatherne-wk: I think Steven was referring to [for/else](https://docs.python.org/3/tutorial/controlflow.html#else-clauses-on-loops)

> In a for or while loop the break statement may be paired with an else clause. If the loop finishes without executing the break, the else clause executes.

aaroncameron-wk (dismissed):

austinmatherne-wk (dismissed):

arelle/plugin/validate/UK/__init__.py
stevenbronson-wk: Still using .startDatetime for one comparison but .endDate for the other.

allenbross-wf:
QA +10
All tests passed.

PR #2041, opened 2025-11-20
Add NL.NL-KVK.6.1.3.1, NL.NL-KVK.6.1.3.2 and NL.NL-KVK.6.1.3.3

derekgengenbacher-wf:
#### Description of change
- Added NL.NL-KVK.6.1.3.1
- Added NL.NL-KVK.6.1.3.2
- Added NL.NL-KVK.6.1.3.3

#### Steps to Test
CI

**review**:
@Arelle/arelle

### DIFF
--- a/arelle/plugin/validate/NL/rules/nl_kvk.py
+++ b/arelle/plugin/validate/NL/rules/nl_kvk.py
@@ -1827,6 +1827,7 @@ def rule_nl_kvk_6_1_3_3(
             msg=_('The target attribute `filing-information` MUST be used for the content of the required '
                   'elements for filing with the Business Register.'),
             modelObject=targetElements
+        )
 
 
 @validation(


arelle/plugin/validate/NL/resources/config.xml
```diff
@@ -15,6 +15,12 @@
             validationType="NL"
             exclusiveTypesPattern="EFM|GFM|FERC|HMRC|SBR.NL|EBA|EIOPA|ESEF"
     />
+    <DisclosureSystem
+            names="NL-INLINE-MULTI-TARGET|nl-inline-multi-target|kvk-inline-multi-target-preview"
```
austinmatherne-wk, line 25: Should this disclosure system include a year?

derekgengenbacher-wf: I added a 2025 disclosure system for IFRS and GAAP here, but that disclosure system is ultimately going to go away because these rules apply to both 2024 and 2025 in some cases. The 2025 disclosure system was more a way to make sure the 2024 conformance suite can still pass while we wait for the 2025 conformance suite. So I would think at this time that it would be more prudent to have a generic multi-target disclosure system rather than a year specific one.

austinmatherne-wk: When you say the 2025 disclosure system will ultimately go away, do you mean the 2024/2025 versions will be combined without a year/version identifier and only the 2025 conformance suite will be expected to pass? I'm not sure if the intent of this validation plugin is to support multiple years, or latest stable. If multiple years, I'd expect all disclosure systems to include a year/version/range, if latest/stable, than I'd expect none of them to include versions except for draft disclosure systems (while we wait for a conformance suite).

derekgengenbacher-wf: Yes your sentiment is correct. The new rules for 2025 will also be applied to 2024, so we will need to remove the dates from the currently existing disclosures so that they are generic.

PR #2057, opened 2025-12-06
Python 3.14 Support

austinmatherne-wk:
#### Reason for change
Our dependencies now support Python 3.14. 3.9 is EOL and no longer receiving security updates.

#### Description of change
* Drop support for EOL Python 3.9.
* Add support for Python 3.14.
* Update frozen builds to use Python 3.14.
* Update Linux build to use latest OpenSSL (not required, but easy to test at the same time).
* Remove Python upper range limit from Python package (it's [not recommended](https://discuss.python.org/t/requires-python-upper-limits/12663) to declare an upper range).

#### Steps to Test
* CI
* [x] Test frozen [Linux build](https://github.com/Arelle/Arelle/actions/runs/19983471007).
* [x] Test frozen [macOS arm build](https://github.com/Arelle/Arelle/actions/runs/19983466146).
* [x] Test frozen [macOS Intel build](https://github.com/Arelle/Arelle/actions/runs/19983465042).
* [x] Test frozen [Windows builds](https://github.com/Arelle/Arelle/actions/runs/19983472566).
* [x] Test [Python package](https://github.com/Arelle/Arelle/actions/runs/19983475824).

**review**:
@Arelle/arelle

### DIFF
--- a/distro.py
+++ b/distro.py
@@ -106,15 +106,6 @@ if sys.platform == LINUX_PLATFORM:
     includeLibs.append("Crypto.Cipher")
     includeLibs.append("Crypto.Cipher.AES")
     includeFiles.append(("arelle/scripts-unix", "scripts"))
-    if os.path.exists("/etc/redhat-release"):
-        includeFiles.append(("/usr/lib64/libexslt.so.0", "libexslt.so"))
-        includeFiles.append(("/usr/lib64/libxml2.so", "libxml2.so"))
-        includeFiles.append(("/usr/lib64/libxml2.so.2", "libxml2.so.2"))
-        includeFiles.append(("/usr/lib64/libxslt.so.1", "libxslt.so"))
-        includeFiles.append(("/lib64/libz.so.1", "libz.so.1"))
-        includeFiles.append(("/usr/lib64/liblzma.so.5", "liblzma.so.5"))
-        includeFiles.append(("/usr/local/lib/tcl8.6", "tcl8.6"))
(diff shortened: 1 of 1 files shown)


distro.py
```diff
@@ -141,12 +148,12 @@
...
     guiExecutable = Executable(
         script="arelleGUI.pyw",
-        base="Win32GUI",
+        base="gui",
         icon="arelle\\images\\arelle16x16and32x32.ico",
     )
     includeFiles.append(("arelle\\scripts-windows", "scripts"))
     if "arelle.webserver" in packages:
-        includeFiles.append("QuickBooks.qwc")
+        includeFiles.append(("QuickBooks.qwc", "QuickBooks.qwc"))
```
brettkail-wk, line 147: Out of curiosity, what does this do?  Based on other usages, it looks like it creates a `QuickBooks.qwc` directory with a same-named file inside.

austinmatherne-wk: It doesn't create a directory, it just copies the file, same as the old format in prior cx_Freeze versions. You can confirm by unzipping the [Windows zip distribution](https://github.com/Arelle/Arelle/actions/runs/19983472566/artifacts/4784340068) linked in the PR description.

Most of our other usages use a source directory, when the source is a file, it handles things differently. The Red Hat shared objects were the only other file specific inputs we had, which reminds me, we can delete that code since we haven't built for Red Hat in a long time.

brettkail-wk: Ah, it's a source, thanks!

PR #2058, opened 2025-12-08
Update lxml and boto3 type libraries

### LINKED ISSUE #2043: Type hints package for lxml needs switching
### What happened?

This is only of interest to those developing Arelle rather than those running it.

lxml-stubs is listed in requirements-dev.txt but [lxml-stubs](https://github.com/lxml/lxml-stubs) has [been abandoned](https://github.com/lxml/lxml-stubs/blob/83f445da9ee27aacf4302a204baaede9b605c50e/README.md?plain=1#L12) in favour of [types-lxml](https://github.com/abelcheung/types-lxml)

A simple switch of one for the other in #2042 made the mypy type checking tests sad so it will need a bit more work.

FWIW, the newer type hints do get more object details correct and cause fewer false-positives in terms of red-wavy lines in VS Code when I've been using them.

**review**:
@Arelle/arelle

### Documents

_No response_

_No response_

### Interface

Python library (pip install)

### Version

598e13d4e2a96e62c61e0bf5010cbda44d5a4a19

### Download

From source (git checkout)

### Operating System

Windows

austinmatherne-wk:
#### Reason for change
resolves #2043 

#### Description of change
* Use updated boto3 types including S3 API.
* Replace deprecated lxml stubs with new types package.
* Resolve mypy errors from updated type libraries.

#### Steps to Test
* CI

**review**:
@Arelle/arelle

### DIFF
--- a/arelle/plugin/validate/EDINET/TableOfContentsBuilder.py
+++ b/arelle/plugin/validate/EDINET/TableOfContentsBuilder.py
@@ -323,7 +323,7 @@ class TableOfContentsBuilder:
     def _getTextParts(
             self,
             elt: ModelObject
-    ) -> tuple[str, str | None, str | None, list[_Element], list[_Element]]:
+    ) -> tuple[str, str | None, str | None, list[ModelObject], list[ModelObject]]:
         """
         Determines the TOC number, label, and tail of a given element (if set correctly)
         based on the text nodes directly beneath the element.


arelle/ModelObjectFactory.py
```diff
@@ -122,7 +122,7 @@ def lookup(self, node_type: str, document: etree._Document, ns: str | None, ln:
             elif ns == XbrlConst.link:
                 if self.type is None:
                     self.type = LINKBASE
-                if modelObjectClass := LINK_LOCALNAME_TO_MODEL_CLASS.get(ln):
+                if ln is not None and (modelObjectClass := LINK_LOCALNAME_TO_MODEL_CLASS.get(ln)):
```
brettkail-wk: This is `if node_type == "element"`, so this `is not None` check "looks" unnecessary.  Would `typing.cast` or `assert` be clearer?

arelle/ModelObject.py
```diff
@@ -160,9 +159,10 @@ def slottedAttributesNames(self) -> set[Any]:
         return emptySet
 
     def setNamespaceLocalName(self) -> None:
-        ns, sep, self._localName = self.tag.rpartition("}")
+        tag = cast(str, self.tag)
```
brettkail-wk: Out of curiosity, is it clearer to add `tag: str | bytes` to the top of `ModelObject` rather than change all use-sites?

arelle/ModelObject.py
```diff
@@ -308,7 +308,7 @@ def elementAttributesStr(self) -> str:
         # Note 2022-09-09:
         # Mypy raises the following error. Not sure why this is the case, this returns a str not binary data?
         # On Python 3 formatting "b'abc'" with "{}" produces "b'abc'", not "abc"; use "{!r}" if this is desired behavior
-        return ', '.join(["{0}='{1}'".format(name, value) for name, value in self.items()]) # type: ignore[str-bytes-safe]
```
brettkail-wk, line 311: Should the comment be removed?

arelle/ModelObject.py
austinmatherne-wk: Most of the use sites are the base class `_Element`, so it doesn't help those. I added `tag: str` as common usage of things like `b'{foo}bar'.rpartition("}")` will throw a type error and it's not clear to me when it would ever be something other than a `str`.

PR #2068, opened 2025-12-09
Convert registry conformance suite configs into test scripts

aaroncameron-wk:
#### Reason for change
These function registry testcases are more like unit tests than integration tests in that they're testing individual formula statements rather than end-to-end XBRL processsing/validation.

Because of this, these conformance suites require special handling in Arelle's current internal testcase variation runner. Similar customizations would be needed in the upcoming test engine overhaul. Instead, we can run these unique conformance suites through their own pathway to keep the standard conformance suite runner clean.

#### Description of change

- Convert function registry conformance suite configs to test scripts
- Update discovery and grouping of integration test scripts (partially in order to support test scripts with private assets)
- Remove unused `test-scripts-parallel.yml`

#### Steps to Test
CI

**review**:
@Arelle/arelle

### DIFF
--- a/tests/integration_tests/scripts/discover_tests.py
+++ b/tests/integration_tests/scripts/discover_tests.py
@@ -56,7 +56,7 @@ def get_all_scripts() -> list[Path]:
 
 def get_frozen_build_scripts() -> list[Path]:
     """
-    Returns absolute paths of runnable scripts based on the operating system.
+    Returns absolute paths of scripts that frozen builds should be tested against.
     :return: Tuple of runnable scripts.
     """
     return [p for p in get_all_scripts() if _for_frozen_build(p)]


tests/integration_tests/scripts/discover_tests.py
```diff
@@ -0,0 +1,103 @@
...
+
+def _for_frozen_build(path: Path) -> bool:
+    if path.stem.startswith("python_api_"):
+        return False
+    if _is_private(path):
+        return False
+    return True
+
+def _is_private(path: Path) -> list[str]:
+    return path.stem in FUNCTION_REGISTRY_TESTS
```
austinmatherne-wk: Here and in a few other places the types hints are incorrect.

tests/integration_tests/scripts/discover_tests.py
```diff
@@ -0,0 +1,103 @@
...
+    Returns absolute paths of runnable scripts based on the operating system.
+    :return: Tuple of runnable scripts.
+    """
+    return [x for x in Path(TESTS_PATH).glob('**/*.py')]
+
+def get_frozen_build_scripts() -> list[Path]:
+    """
+    Returns absolute paths of runnable scripts based on the operating system.
+    :return: Tuple of runnable scripts.
+    """
```
austinmatherne-wk, line 61: copy paste from `get_all_scripts()`?

tests/integration_tests/scripts/tests/xbrl_transformation_registry_4.py
```diff
@@ -0,0 +1,69 @@
...
+
+from tests.integration_tests.scripts.script_util import run_arelle, parse_args, validate_log_text, assert_result, prepare_logfile
+from tests.integration_tests.validation.conformance_suite_config import ConformanceSuiteAssetConfig
+from tests.integration_tests.validation.download_assets import download_assets
+
+errors = []
+this_file = Path(__file__)
+args = parse_args(
+    this_file.stem,
+    "Confirm malformed UTR fires the expected errors.",
```
austinmatherne-wk: Not all of the tests in the suites are for malformed values?

tests/integration_tests/scripts/tests/xbrl_formula_1_0_function_registry.py
```diff
@@ -0,0 +1,78 @@
...
+    plugins=['formulaXPathChecker', 'functionsMath'],
+    offline=arelle_offline,
+    logFile=arelle_log_file,
+)
+
+errors += validate_log_text(arelle_report_file, expected_results=expected_results)
+assert_result(errors)
+
+print("Cleaning up")
+rmtree(working_directory / 'xbrl_formula_1_0_function_registry' / 'suite')
```
austinmatherne-wk, line 79: The other test scrips delete the suite zip, but this one doesn't. Is that intentional?

aaroncameron-wk: None of them should actually, they're each using the existing conformance suite download process which places it in the shared `resources` location.

tests/integration_tests/scripts/discover_tests.py
aaroncameron-wk: I was separately working on configuring type hint checks in the tests module, might as well pull that in here.

derekgengenbacher-wf (approved):

austinmatherne-wk (approved):

PR #2072, opened 2025-12-12
EDINET.EC5611W

aaroncameron-wk:
#### Reason for change
EDINET.EC5611W: The calculated totals for the accounts in the main financial statements, when using the appropriate combination of context and extended link roles, should be within the range of differences due to rounding.

#### Description of change
This implementation makes use of the existing internal calculation validations with some adjustments that allow filtering of inconsistencies before actually registering an error. This isn't optimal since a non-negligible amount of calculations are validated that will ultimately be discarded, but I found it preferable to (a) attempting to duplicate the internal process into the EDINET plugin or (b) implementing the EDINET-specific style of filtering into the internal process.

#### Steps to Test
CI

**review**:
@Arelle/arelle


arelle/Cntlr.py
```diff
@@ -298,6 +299,11 @@ def __init__(
         self.startLogging(logFileName, logFileMode, logFileEncoding, logFormat)
         self.errorManager = ErrorManager(self.modelManager, logging._checkLevel("INCONSISTENCY")) # type: ignore[attr-defined]
 
+    def validation(self, val: Validation, fileSource: FileSource | None = None) -> None:
+        """Same as log, but parameters passed in from Validation object
```
austinmatherne-wk: Cntlr doesn't have a log function? Just an error method that takes a level for some reason

arelle/plugin/validate/EDINET/rules/edinet.py
```diff
@@ -178,6 +177,59 @@ def rule_EC5002E(
...
+                  "<roleUri=%(linkrole)s> <contextID=%(contextID)s> <unit=%(unitID)s>."),
+            reportedSum=_validation.args.get("reportedSum"),
+            concept=_validation.args.get("concept"),
+            computedSum=_validation.args.get("computedSum"),
+            count=len(_validation.args.get("modelObject", [])) - 1,
+            linkrole=_validation.args.get("linkrole"),
+            contextID=_validation.args.get("contextID"),
+            unitID=_validation.args.get("unitID"),
+            modelObject=_validation.args.get("modelObject"),
+        )
```
austinmatherne-wk: `warning()` is a static method, so calling it on `_validation` is misleading (it doesn't use any data from the instance). Either call `Validation.warning(...)` directly, or add an instance method like `derive()` that merges `self.args` with new values to avoid manually copying args.

arelle/plugin/validate/EDINET/rules/edinet.py
```diff
@@ -4,15 +4,14 @@
 from __future__ import annotations
 
 from collections import defaultdict
-from collections.abc import Iterable
 from pathlib import Path
-from typing import Any
+from typing import Any, Iterable
```
austinmatherne-wk: `typing.Iterable` is deprecated in favor of `collections.abc.Iterable`.

arelle/ValidateXbrlCalcs.py
```diff
@@ -142,9 +143,15 @@ def validate(self):
...
+                    "calc11e:tuplesInReportWarning",
+                    _("Validating of calculations ignores tuples."),
+                )
             if any(e in oimXbrlxeBlockingErrorCodes for e in oimErrs if e != "xbrlxe:unsupportedTuple"):
-                modelXbrl.warning("calc11e:oimIncompatibleReportWarning","Validating of calculations is skipped due to OIM errors.")
+                yield Validation.warning(
+                    "calc11e:oimIncompatibleReportWarning",
+                    _("Validating of calculations is skipped due to OIM errors."),
+                )
                 return;
```
austinmatherne-wk: preexisting: trailing semicolon

austinmatherne-wk (approved):

PR #2082, opened 2025-12-17
LRR suite config uses fixup plugin instead of url_replace

aaroncameron-wk:
#### Reason for change
Part of the motivation of the new test engine is that test cases can be recreated by running Arelle directly, and that Arelle is essentially a black box to the test engine. This means patching `normalizeUrl` is not an option. 

#### Description of change
Instead, we can use the existing `WebCache.TransformURL` hook and a plugin to implement the same conformance-suite-specific URL rewriting.

#### Steps to Test
CI

**review**:
@Arelle/arelle

### DIFF
--- a/tests/plugin/lrrUrlRewriter.py
+++ b/tests/plugin/lrrUrlRewriter.py
@@ -15,7 +15,7 @@ def webCacheTransformUrl(cntlr: Cntlr, url: str | None, base: str | None = None)
     return url, False
 
 __pluginInfo__ = {
-    'name': 'LLR Conformance Suite URL Rewriter',
+    'name': 'LRR Conformance Suite URL Rewriter',
     'version': '1.0.0',
     'description': "Maps \"file:///c:/temp/conf/\" URLs to path within LRR conformance suite.",
     'license': 'Apache-2',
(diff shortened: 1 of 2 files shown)


tests/plugin/lrrUrlRewriter.py
```diff
@@ -0,0 +1,26 @@
...
+"""
+from pathlib import Path
+
+from arelle.Cntlr import Cntlr
+from arelle.Version import authorLabel, copyrightLabel
+
+PREFIX = str(Path("/c:/temp/conf/"))
+REPLACE = str(Path("tests/resources/conformance_suites/lrr-conf-pwd-2005-06-21.zip/"))
+
+def webCacheTransformUrl(cntlr: Cntlr, url: str | None, base: str | None = None) -> tuple[str | None, bool]:
```
stevenbronson-wk, line 12: Opinions on making it a command line argument?

aaroncameron-wk: I explored that route, and actually have another branch for it, but re-using an existing feature seemed preferable to adding a CLI argument just for this conformance suite use case. If it turns out to be necessary elsewhere or it is requested in the future, we can go that route.

aaroncameron-wk: But I do see the benefit of not relying on a conformance suite specific plugin as well. I can go either way. I'll do whatever anyone feels strongly about.

stevenbronson-wk: I don't have a strong opinion.  It just seemed potentially convenient to remap paths for development or testing of taxonomies or conformance suites.

PR #2096, opened 2026-01-06
Add `--taxonomyPackage` to force validation as taxonomy package

aaroncameron-wk:
#### Reason for change
Some key adjustments are necessary to have the report/taxonomy package conformance suites run correctly and pass using the new test engine. These changes get us most of the way there without getting lost in the larger test engine PR to come.

#### Description of change
- Incorporate taxonomy package validation directly in `ValidateFileSource`, similar to report package validation.
- Add CLI and GUI options for forcing taxonomy package validation
- Update package validation to add error to controller using `Cntlr.error` (which prevents the need for some error bubbling in the testcase variation runner)
- Add a couple of new expected errors as a result of these changes

#### Steps to Test
[V-701-zip-with-no-taxonomy.zip](https://github.com/user-attachments/files/24457469/V-701-zip-with-no-taxonomy.zip)
- Validate via GUI without "Validate all files as Taxonomy Packages" checked.
  - No `tpe:metadataFileNotFound`.
- Validate via GUI with "Validate all files as Taxonomy Packages" checked.
  - `tpe:metadataFileNotFound` fired.
- `python arelleCmdLine.py -f "V-701-zip-with-no-taxonomy.zip" --validate`
  - No `tpe:metadataFileNotFound`.
- `python arelleCmdLine.py -f "V-701-zip-with-no-taxonomy.zip" --validate --taxonomyPackage`
  - `tpe:metadataFileNotFound` fired.

**review**:
@Arelle/arelle

### DIFF
--- a/arelle/Validate.py
+++ b/arelle/Validate.py
@@ -220,7 +220,8 @@ class Validate:
                                         modelObject=testcaseVariation,
                                         variationId=testcaseVariation.id)
 
-            self.modelXbrl.modelManager.cntlr.testcaseVariationReset()
+            # Clear errors added to the controller during testcase loading.
+            self.modelXbrl.modelManager.cntlr.errorManager.clear()
             for modelTestcaseVariation in testcaseVariations:
                 self._validateTestcaseVariation(testcase, modelTestcaseVariation)
 


tests/integration_tests/validation/conformance_suite_configurations/xbrl_report_packages_1_0.py
```diff
@@ -16,6 +16,14 @@
...
     ],
     expected_additional_testcase_errors={f"report-package-conformance/index.csv:{s}": val for s, val in {
+        # Invalid zip also fires FileSourceError.
+        "V-000-invalid-zip": {
+            "FileSourceError": 1,
+        },
+        # Triggers a scenario where the testcase variation runner performs taxonomy package validation twice.
+        "V-012-encrypted-zip": {
+            "tpe:invalidArchiveFormat": 1,
+        },
```
austinmatherne-wk: Why does it run twice? Is this temporary until the rest of the changes are committed?

arelle/PackageManager.py
```diff
@@ -559,11 +576,13 @@ def packageInfo(cntlr, URL, reload=False, packageManifestName=None, errors=[]):
...
-                                           file=os.path.basename(URL),
-                                           level=logging.ERROR)
+                            cntlr.error(
+                                codes="tpe:packageDuplicateMapping",
+                                msg=_("Package mapping duplicate rewrite start string %(rewriteStartString)s"),
+                                fileSource=filesource,
+                                rewriteStartString=prefix,
+                                file=os.path.basename(URL),
+                            )
                             errors.append("arelle.packageDuplicateMapping")
```
austinmatherne-wk, line 586: The logged error code prefix changed from `arelle.packageDuplicateMapping` to `tpe:packageDuplicateMapping`, but the code appended to errors still uses the `arelle` prefix. The package spec doesn't define this error code, so it shouldn't use `tpe`.

arelle/Validate.py
```diff
@@ -219,6 +219,8 @@ def validateTestcase(self, testcase):
                     self.modelXbrl.info("info", "Skipped testcase variation %(variationId)s.",
                                         modelObject=testcaseVariation,
                                         variationId=testcaseVariation.id)
+
+            self.modelXbrl.modelManager.cntlr.testcaseVariationReset()
             for modelTestcaseVariation in testcaseVariations:
                 self._validateTestcaseVariation(testcase, modelTestcaseVariation)
```
austinmatherne-wk: Out of curiosity, why does the reset need to be called prior to variation validation? Doesn't the variation validation function already call this after each variation?

tests/integration_tests/validation/conformance_suite_configurations/xbrl_report_packages_1_0.py
aaroncameron-wk: Since `validateTaxonomyPackage` is called within `loadTaxonomyPackageMappings`, validation was running twice. My latest commit ensures taxonomy package mappings only attempt loading once, and I've swapped out the direct call to `validateTaxonomyPackage` within `ValidateFileSource` with `loadTaxonomyPackageMappings`

arelle/Validate.py
aaroncameron-wk: Errors from the loading of the conformance suite itself can exist on the controller at this point (and they do now that those taxonomy package errors are using `Cntlr.error`). There's probably a way to get away with only calling it once but I'd rather not mess with it when this code will be removed in the near future anyway.

austinmatherne-wk: Does clearing the error manager directly (no plugin data reset) instead cover everything that's required? It's not intuitive why testcase variation reset is called (resetting errors from loading the suite, not a variation). Either way, a comment above this line explaining why it's here would be useful.

PR #2097, opened 2026-01-08
Retain invalid facts when partitioning on context hash. 

stevenbronson-wk:
**review**:
@Arelle/arelle

### DIFF
--- a/arelle/ValidateDuplicateFacts.py
+++ b/arelle/ValidateDuplicateFacts.py
@@ -530,11 +530,13 @@ def getFactValueEqualityKey(fact: ModelFact) -> TypeFactValueEqualityKey:
 def getHashEquivalentFactGroups(facts: list[ModelFact]) -> list[list[ModelFact]]:
     """
     Given a list of facts in an instance, returns a list of lists of facts
-    that are concept/context/unit hash-equivalent.
+    that are concept/context/unit hash-equivalent.  Return unhashable facts
+    in singleton lists.
     :param facts:
     :return: List of hash-equivalent fact lists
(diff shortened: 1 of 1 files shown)


austinmatherne-wk (approved):

arelle/ValidateDuplicateFacts.py
```diff
@@ -547,7 +549,9 @@ def getHashEquivalentFactGroups(facts: list[ModelFact]) -> list[list[ModelFact]]
             and f.concept.type is not None
         ):
             hashDict[f.conceptContextUnitHash].append(f)
-    return list(hashDict.values())
+        else:
+            unhashableGroups.append([f])
+    return list(hashDict.values()) + unhashableGroups
```
brettkail-wk, line 554: This seems ok, but do you have any additional details/motivation?  I guess `saveDeduplicatedInstance` is dropping the invalid facts but should keep them?

(It's perhaps a little unfortunate that the order of the facts isn't maintained.)

brettkail-wk (approved):

arelle/ValidateDuplicateFacts.py
stevenbronson-wk: Right, saveDeduplicatedInstance excludes "duplicates", defined as the complement of getDeduplicatedFacts ~ getDuplicateFactSets(includeSingles=True), which prior to this included invalid facts in "duplicates".  "Should keep them" I guess is a matter of opinion, but I'm claiming yes.

aaroncameron-wk (approved):

PR #2101, opened 2026-01-12
Centralize validation extension functions

allenbross-wf:
#### Description of change

Moved extension concept validation functions to a centralized location.

#### Steps to Test

CI

**review**:
@Arelle/arelle

### DIFF
--- a/arelle/utils/validate/Concepts.py
+++ b/arelle/utils/validate/Concepts.py
@@ -1,0 +1,21 @@
+from __future__ import annotations
+
+from arelle.ModelDtsObject import ModelConcept
+from arelle.ModelXbrl import ModelXbrl
+
+
+def isExtensionUri(uri: str, modelXbrl: ModelXbrl, taxonomyUrlPrefixes: frozenset[str]) -> bool:
+    if uri.startswith(modelXbrl.uriDir):
+        return True
+    return not any(uri.startswith(taxonomyUri) for taxonomyUri in taxonomyUrlPrefixes)
+
+def getExtensionConcepts(modelXbrl: ModelXbrl, taxonomyUrlPrefixes: frozenset[str]) -> list[ModelConcept]:
+    """
+    Returns a list of extension concepts in the DTS.
+    """
+    extensionConcepts = []
+    for concepts in modelXbrl.nameConcepts.values():
+        for concept in concepts:
+            if isExtensionUri(concept.document.uri, modelXbrl, taxonomyUrlPrefixes):
+                extensionConcepts.append(concept)
+    return extensionConcepts
--- a/arelle/plugin/validate/EDINET/Constants.py
+++ b/arelle/plugin/validate/EDINET/Constants.py
@@ -84,6 +84,21 @@ NUMERIC_LABEL_ROLES = frozenset({
     'http://www.xbrl.org/2009/role/negatedTerseLabel',
 })
 
+STANDARD_TAXONOMY_URL_PREFIXES = frozenset((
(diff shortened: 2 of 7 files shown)


arelle/utils/validate/Concepts.py
```diff
@@ -0,0 +1,24 @@
...
+
+@lru_cache(1)
+def getExtensionConcepts(modelXbrl: ModelXbrl, taxonomyUrlPrefixes: frozenset[str]) -> list[ModelConcept]:
+    """
+    Returns a list of extension concepts in the DTS.
+    """
+    extensionConcepts = []
+    for concepts in modelXbrl.nameConcepts.values():
+        for concept in concepts:
+            if isExtensionUri(concept.document.uri, modelXbrl, taxonomyUrlPrefixes):
```
austinmatherne-wk: The two previous implementations were inconsistent (one used `concept.document.uri`, the other `concept.qname.namespaceURI`). While I expect they'll return the same results most of the time, using the document uri can return incorrect results in some edge cases.

arelle/utils/validate/Concepts.py
```diff
@@ -0,0 +1,24 @@
...
+from arelle.ModelXbrl import ModelXbrl
+
+
+def isExtensionUri(uri: str, modelXbrl: ModelXbrl, taxonomyUrlPrefixes: frozenset[str]) -> bool:
+    if uri.startswith(modelXbrl.uriDir):
+        return True
+    return not any(uri.startswith(taxonomyUri) for taxonomyUri in taxonomyUrlPrefixes)
+
+@lru_cache(1)
+def getExtensionConcepts(modelXbrl: ModelXbrl, taxonomyUrlPrefixes: frozenset[str]) -> list[ModelConcept]:
```
austinmatherne-wk: Moving this function made me realize we have a memory leak. The `lru_cache` holds a strong reference to `modelXbrl`, preventing it from being garbage collected until another call evicts it. There is similar usage across the plugin data classes, so a proper fix is more extensive. The plugin data usage is more problematic, as both the model and plugin data class instances [leak](https://docs.python.org/3/faq/programming.html#faq-cache-method-calls):

> The lru_cache approach works with methods that have hashable arguments. It creates a reference to the instance unless special efforts are made to pass in weak references.
>
> The advantage of the least recently used algorithm is that the cache is bounded by the specified maxsize. The disadvantage is that instances are kept alive until they age out of the cache or until the cache is cleared.

Can you remove the `lru_cache` from this new function (I doubt the performance benefit justifies managing a cache) and create a follow-up ticket to audit/fix the other usages?

allenbross-wf: XT-5404

austinmatherne-wk (approved):

arelle/utils/validate/Concepts.py
```diff
@@ -0,0 +1,21 @@
...
+    return not any(uri.startswith(taxonomyUri) for taxonomyUri in taxonomyUrlPrefixes)
+
+def getExtensionConcepts(modelXbrl: ModelXbrl, taxonomyUrlPrefixes: frozenset[str]) -> list[ModelConcept]:
+    """
+    Returns a list of extension concepts in the DTS.
+    """
+    extensionConcepts = []
+    for concepts in modelXbrl.nameConcepts.values():
+        for concept in concepts:
+            if isExtensionUri(concept.qname.namespaceURI, modelXbrl, taxonomyUrlPrefixes):
```
aaroncameron-wk: We want to use the document URI, not the namespace URI.

PR #2104, opened 2026-01-14
Change accepted domain item types for the DBA in ESEF.

stevenbronson-wk:
Denmark seems to reject some DTR versions.

**review**:
@Arelle/arelle

### DIFF
--- a/arelle/plugin/validate/ESEF/resources/authority-validations.json
+++ b/arelle/plugin/validate/ESEF/resources/authority-validations.json
@@ -31,7 +31,8 @@
         "authorityRequiredTaxonomyURLs: authority-required standard URL in addition to ESEF effectiveTaxonomyURLs",
         "standardTaxonomyURIs: set of taxonomy URIs which are deemed base taxonomies",
         "formulaMessagesAdditionalURLs: for base URL list of additional formula messages URLs to load out of DTS",
-        "formulaRunIDs: if not null, regex of formula IDs to run (or with negative lookahead regex, to suppress)"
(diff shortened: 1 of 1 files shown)


arelle/plugin/validate/ESEF/ESEF_Current/DTS.py
```diff
@@ -101,6 +102,19 @@ def lc3wordAdjust(word: str) -> str:
         calcRelSet = val.modelXbrl.relationshipSet(XbrlConst.summationItems)
         dimensionDefaults = val.modelXbrl.relationshipSet(dimensionDefault, DefaultDimensionLinkroles)
         labelsRelationshipSet = val.modelXbrl.relationshipSet(XbrlConst.conceptLabel)
+
+        if (authorityDtrNamespaces := val.authParam.get("dtrNamespaces")) is not None:
+            esefDomainItemTypes = frozenset(qname(ns, "domainItemType") for ns in authorityDtrNamespaces.keys())
```
brettkail-wk, line 107: Is it significant that the `QName` doesn't have a `prefix`?  Assuming not, I wonder if it's simpler to just remove the prefix from the constants.

It's a little odd to assume that the item type is `domainItemType` in all versions.  I guess the intent of `dtrNamespaces` is that it's one of the namespaces already known.

stevenbronson-wk: It's not significant.

If domainItemType changes we have other places to update, like isDomainItemType, so I'm sure we'll update this, too.  I can put the list of item types in the configuration if people feel strongly.

PR #2116, opened 2026-01-22
DBA.FR116, DBA.FR118

aaroncameron-wk:
#### Reason for change
DBA.FR116: It is not permitted to declare numeric fields with multiple values in the same period.
DBA.FR118: Fields that appear in the other rendering dimension must also appear in the reported dimension with a different scale.

#### Steps to Test
CI

**review**:
@Arelle/arelle


arelle/plugin/validate/DBA/rules/fr.py
```diff
@@ -1252,6 +1255,8 @@ def rule_fr87(
         ixTargets = set()
         for ixdsHtmlRootElt in modelXbrl.ixdsHtmlElements:
             for elt in ixdsHtmlRootElt.iter():
+                if not isinstance(elt, ModelObject):
+                    continue
```
stevenbronson-wk: Assuming you're avoiding non-elements like comments, you can use `.iter(etree.Element)`.

arelle/plugin/validate/DBA/rules/fr.py
```diff
@@ -1494,3 +1499,68 @@ def findOverlappingPeriods(groupedFacts: dict[ContextHashKey, list[ModelFact]],
...
+            dimQname=pluginData.reportedValueOtherRenderingOfReportedValueDimensionQn,
+            memQname=pluginData.otherRenderingOfReportedValueMemberQn,
+    ):
+        if fact.xValid < VALID or fact.context is None:
+            continue
+        otherRenderingFactsMap[(
+            fact.qname,
+            fact.context.periodHash,
+            fact.context.entityIdentifierHash,
+            fact.context.nonDimHash,
```
stevenbronson-wk: I think this is for XBRL that uses dimensions but not "XBRL Dimensions".  I don't know the validation details, but I'd guess you might want to compare dimensions and members except for the ReportedValueOtherRenderingOfReportedValueDimension.

arelle/plugin/validate/DBA/rules/fr.py
```diff
@@ -1494,3 +1499,68 @@ def findOverlappingPeriods(groupedFacts: dict[ContextHashKey, list[ModelFact]],
...
+        )][
+            fact.scaleInt # type: ignore[attr-defined]
+        ].append(fact)
+
+    errorFacts = []
+    for key, scaleMap in otherRenderingFactsMap.items():
+        for scale, facts in scaleMap.items():
+            matchingReportedFactsMap = reportedFactsMap[key]
```
stevenbronson-wk: I expected to see some equality checks corresponding to the hashes in the key.

arelle/plugin/validate/DBA/rules/fr.py
aaroncameron-wk: The intent of the rule is to apparently give a pathway to allow otherwise duplicate facts by adding this dimension member.  Without refactoring the existing duplicate fact detection I don't think we'll be able to get an exact implementation of that, which is why I've hopefully built a "close enough" check here that now matches on period, entity, and dimensions other than the dimension in question.

arelle/plugin/validate/DBA/rules/fr.py
aaroncameron-wk: The key from one map being used to get a value from the other map is the equality check. I can't see a reason to break out the individual hashes for further comparison.

brettkail-wk: Hashes aren't guaranteed equality.  See [this code](https://github.com/Arelle/Arelle/blob/556f0acf72d90606ceac96848f0078b8eb499aa4/arelle/archive/plugin/validate/calc2.py#L137-L140) or elsewhere: after doing the hash comparison, we need an equality check.

(e.g., `ContextHashKey` avoids that problem by making the dict key have the full value rather than just a hash.)

stevenbronson-wk: I see this has been merged, but I don't see the equality checks.

aaroncameron-wk: [Swapped out hashes](https://github.com/Arelle/Arelle/pull/2128) in favor of underlying values in the key.

PR #2118, opened 2026-01-23
Test Engine Implementation

aaroncameron-wk:
#### Description of change
Implementation of a new "Test Engine" for running conformance suites. The intent is for this to replace the existing conformance suite and testcase loading pathway currently in Arelle.

Key Points:
* Each testcase run is an independent execution of Arelle via the Session API.
  * The previous testcase validation ran through a separate validation pathway specific to testcase registries, diminishing the the accuracy of the tests and forcing validation authors to take both pathways into consideration. The test engine runs each testcase as an independent execution of Arelle through the Session API.
  * Each execution logs its runtime options, log output, and result to a dedicated file.
  * The runtime options JSON logged for each testcase can be quickly copied into a `runtime_options.json` and reproduced via `python arelleCmdLine.py —optionsFile runtime_options.json`.
* The previous testcase validation required us to pre-arrange a predetermined number of groups of testcases to take advantage of multiprocessing. With the test engine, multiprocessing has full control over executing the pool of testcases, taking full advantage of the resources on the machine by default. Batching testcases is now only necessary when dividing testcases onto different machines.

Follow-up Work:
* Running testcases via GUI should use the test engine
* The test engine’s initial XML loader depends on the existing implementation. Rework to use plain lxml for loading XML testcase registries.
* Clear out code specific to the old testcase validation pathway.
* `arelleTestEngine.py` for command line execution and bundling with frozen builds.
* Most conformance suites transitioned seamlessly, but others revealed some follow-up may be needed to address new "expected" errors/failures.
* There's some complexity involved in running supporting selection of multiple shards in our CI run configuration. I don't believe there's a meaningful use case for running more than one shard at a time, so this can probably be simplified.

Note: Overall, the test engine execution is slower than the previous implementation. This is most apparent in CI environments where the CPU cores were already being maxed out, but in practice people will find conformance suite's faster due to the default parallelization taking full advantage of their computer's resources. For our CI, I've increased the number of runners allocated to the EFM conformance suite to partially offset this longer runtime.

#### Steps to Test
CI

**review**:
@Arelle/arelle

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -171,7 +171,7 @@ Include a license at the top of new Python files ([example][python-license-examp
 
 #### Python Coding Style
 
-Stick to the [PEP8 guidelines][pep-0008] and use `mixedCase` for variable and function
-names.
+Stick to the [PEP8 guidelines][pep-0008]. 
+The project historically uses `mixedCase` which needs to be retained for backwards compatibility due to external code integrations, but new features can use current best practices.
 
 [pep-0008]: https://www.python.org/dev/peps/pep-0008/


arelleTestEngine.py
austinmatherne-wk, line 1: I'd prefer not to include this file until it's implemented.

arelle/testengine/ErrorLevel.py
```diff
@@ -0,0 +1,17 @@
...
+See COPYRIGHT.md for copyright information.
+"""
+from __future__ import annotations
+
+from enum import Enum
+
+
+class ErrorLevel(Enum):
+    OK = "OK"
+    SATISIFED = "SATISIFED"
```
austinmatherne-wk: Typo

arelle/testengine/TestEngine.py
```diff
@@ -0,0 +1,409 @@
...
+            assert key_val is not None
+            constraint_key = (key_val, constraint.level)
+            match_count = 0
+            for actual_key, count in actual_error_counts.items():
+                actual_error, level = actual_key
+                if level != constraint.level:
+                    continue
+                if compare_context.compare(constraint, actual_error):
+                    match_count += count
+                    actual_error_counts[actual_key] -= count
```
austinmatherne-wk: It's surprising for a method called `get_x` to mutate arguments. The caller doesn't reuse `actual_error_counts` so it's fine, but fragile. It would be better to create a copy of `actual_error_counts` (with a different name, since it's no longer the actual errors counts after modifying) inside of this function for it to mutate.

arelle/testengine/TestEngine.py
```diff
@@ -0,0 +1,409 @@
...
+        for uri in uris
+    ]
+    if len(uris) > 1:
+        if all(uri.suffix in ('.htm', '.html', '.xhtml') for uri in uris):
+            docset_surrogate_path = norm_path(uris[0].parent) + os.sep + IXDS_SURROGATE
+            return [docset_surrogate_path + IXDS_DOC_SEPARATOR.join(norm_path(uri) for uri in uris)]
+    return [norm_path(uri) for uri in uris]
+
+
+def _collect_errors(session: Session) -> list[str | None]:
```
austinmatherne-wk: Is this return type right? The caller asserts that it's `list[str | dict]`.

arelle/testengine/TestEngine.py
```diff
@@ -0,0 +1,409 @@
...
+    XmlTestcaseLoader()
+]
+
+
+def _block_codes(actual_errors: list[ActualError], pattern: str) -> tuple[list[ActualError], dict[str, int]]:
+    results = []
+    blocked_codes: dict[str, int] = defaultdict(int)
+    if not pattern:
+        return actual_errors, blocked_codes
+    compiled_pattern = re.compile(re.sub(r'\\(.)', r'\1', pattern))
```
austinmatherne-wk: I'm probably overlooking something basic, but what is `re.sub` used for here? A comment would be helpful.

arelle/testengine/CompareContext.py
```diff
@@ -0,0 +1,64 @@
...
+        return False
+
+    @staticmethod
+    def _compare_qname(constraint: Constraint, qname: QName) -> bool:
+        if qname is None:
+            return False
+        if constraint.qname is None:
+            if constraint.pattern == qname.localName:
+                return True
+        if constraint.qname == qname:
```
austinmatherne-wk, line 38: nit: We could use an early `return` or `elif` for when `constraint.qname is None` and `constraint.pattern != qname.localName` this falls through to a redundant `constraint.qname == qname` check when we already know `qname` is not `None` and `constraint.qname` is.

```suggestion
        if constraint.qname is None:
            return constraint.pattern == qname.localName
        if constraint.qname == qname:
```

arelle/testengine/TestEngine.py
```diff
@@ -0,0 +1,409 @@
...
+CWD = Path.cwd()
+DEFAULT_PLUGIN_OPTIONS = {
+    'EDGAR/render': {
+        'keepFilingOpen': True,
+    },
+    'xule': {
+        "xule_time": 2.0,
+        "xule_rule_stats_log": True,
+    }
+}
```
austinmatherne-wk, line 47: nit: mix of single and double quotes.

arelle/Cntlr.py
```diff
@@ -519,6 +521,9 @@ def close(self, saveConfig: bool = False) -> None:
             PackageManager.save(self)
         if saveConfig:
             self.saveConfig()
+        if self.modelManager is not None:
+            for modelXbrl in self.modelManager.loadedModelXbrls:
+                self.modelManager.close(modelXbrl)
```
austinmatherne-wk: `modelManager.close(modelXbrl)` removes the model from `self.modelManager.loadedModelXbrls`. Directly iterating over the collection while mutating it can cause models to be skipped. It's safer to iterate over a copy of the collection instead.

tests/integration_tests/validation/conformance_suite_config.py
```diff
@@ -214,6 +215,7 @@ class ConformanceSuiteConfig:
     expected_additional_testcase_errors: dict[str, dict[str, int]] = field(default_factory=dict)
     expected_failure_ids: frozenset[str] = frozenset()
     expected_missing_testcases: frozenset[str] = frozenset()
+    ignore_levels: frozenset[ErrorLevel] = frozenset({ErrorLevel.OK, ErrorLevel.WARNING})
```
austinmatherne-wk: Out of curiosity, how many configs would need `ErrorLevel.WARNING` added if you removed it from this default ignore list?

arelle/conformance/CSVTestcaseLoader.py
```diff
@@ -1,102 +0,0 @@
...
-    "rpe": "https://xbrl.org/2023/report-package/error",
-    "tpe": "http://xbrl.org/2016/taxonomy-package/errors",
-}
-
-
-class CSVTestcaseException(Exception):
-    pass
-
-
-def loadCsvTestcase(
```
austinmatherne-wk, line 40: Mentioned at standup, but I think we need to retain this plugin (for now) so conformance suites can continue to be loaded from the GUI/standard CLI.

arelle/packages/report/ReportPackage.py
```diff
@@ -27,9 +27,9 @@ def pathsContainReportsDirectory(paths: set[str], topLevelDir: str) -> bool:
     return any(path.startswith(reportDirectory) for path in paths)
 
 def getReportPackageTopLevelDirectory(filesource: FileSource) -> str | None:
-    packageEntries = set(filesource.dir or [])
+    packageEntries = PackageUtils.getPackageEntries(filesource)
```
austinmatherne-wk, line 30: Curious. What initiated this change?

arelle/CntlrCmdLine.py
```diff
@@ -1360,6 +1362,19 @@ def monitorParentProcess():
             for pluginXbrlMethod in PluginManager.pluginClassMethods("Validate.Complete"):
                 pluginXbrlMethod(self, filesource)
 
+        if options.reportCount is not None:
+            reportCount = len([
+                model for model in self.modelManager.loadedModelXbrls
+                if model.modelDocument is not None and (model.fileSource.isReportPackage or not model.fileSource.isTaxonomyPackage)
```
austinmatherne-wk: If a user opens a taxonomy directly (no instance) without using a taxonomy package, this will count that as a report, won't it? Also, I wasn't quite sure what this argument was intended to do based on the name. Maybe call this `expectedInstanceCount` to help clarify how multi-doc and/or multi-target inline document sets are counted?

.github/actions/test_conformance_suite/action.yml
```diff
@@ -10,6 +10,9 @@ inputs:
   shard:
     description: Optional selection of shards to run (e.g. "0", "0-3")
     required: false
+  shard_count:
```
austinmatherne-wk, line 13: arbitrary line:

Good work. There are a lot of nice improvements here. Admittedly, it's difficult to do a deep review of the entire changeset at once.

While I anticipated some new "expected" errors, the volume here gives me some pause. I don't think it should hold up the PR, but investigating them should be prioritized as follow-up.

One of the most persistent challenges with testing Arelle is the number of different ways the same files can be opened (multiple GUI options, the iXDS json syntax, zipped vs. unzipped archives, etc.) each of which exercises at least some unique loading code. Combined with cross-platform inconsistencies, the testing burden is significant.

I believe the long-term goal of this work is to reduce the number of code paths through which files are loaded, so that conformance suite paths all use the same code as opening the test file directly. However, in the current state of this PR, it actually increases the surface area: conformance suites can now go through the standard CLI, the GUI, or the test engine. You mentioned follow-up work to migrate the GUI to the test engine, and I want to make sure we don't end up in an extended intermediate state where we're running conformance suites through the test engine (such as the EFM suite) while other users (such as the SEC) are running them differently and potentially getting different results.

arelleTestEngine.py
aaroncameron-wk: Agreed. I have it on another branch and the empty file slipped in here somehow.

arelle/testengine/TestEngine.py
aaroncameron-wk: I think the goal is that it was unescaping regex escape characters, turning `\\.` into `\.`. But I'm not sure why I have that in place because the inputs are already `\.`. I'll remove and see if I reveals a reason for that to be there.

arelle/packages/report/ReportPackage.py
aaroncameron-wk: The problem boils down to invalid paths (leading slash, backslash) preventing the report package from being detected and loading.

This prevents the actual validation rules that yield the appropriate errors from firing.

The options for solutions would be:
- Attempt to find the report package JSON and load the report package despite the invalid paths
- Change the controller logic so that the validations still run if the report package JSON could not be found (this works to get the expected validations firing, but yields an `IOerror` since the report ultimately fails to load in the end).

The former (implemented in this PR) matches the expected conformance suite output exactly, which is why I chose it. But an argument could be made that an `IOerror` is also appropriate since the report package is, in fact, invalid. I don't feel strongly so I'll defer to you if you have a preference. Switching between the two solutions is a minor change.

.github/actions/test_conformance_suite/action.yml
aaroncameron-wk: Agreed. I think the majority of changes in expected errors are things that were either caused by the old validation pathway or were hidden by the old validation pathway. I'll do a review to see what work might need to be done. 

As far as replacing the old pathway entirely, what are your thoughts on the `arelleTestEngine.py` approach versus auto-detecting conformance suites within `arelleCmdLine.py` and rerouting to the test engine? I prefer the former, as my overall philosophy is that the test engine is a wrapper for core Arelle executions and not something that Arelle should even be aware of internally.

Although as far as the GUI is concerned, our hands are kind of tied there as far as needing to have crossover between the Test Engine and internal Arelle logic. We should be able to insulate that pretty well though. I was able to put together a proof of concept where the GUI is able to run the test engine in multiprocessing mode and was pretty surprised with the results.

tests/integration_tests/validation/conformance_suite_config.py
aaroncameron-wk: Turns out, none.

PR #2119, opened 2026-01-23
Add disclosure_system_by_prefix.

stevenbronson-wk:
It doesn't make sense to run the Netherlands iXBRL conformance suite twice but then mark almost all tests as expected failures.  Run the relevant tests under the other disclosure system.

**review**:
@Arelle/arelle


tests/integration_tests/validation/conformance_suite_configurations/nl_inline_2024.py
```diff
@@ -18,6 +18,10 @@
         *NL_PACKAGES['NL-INLINE-2024'],
     ],
     base_taxonomy_validation='none',
+    disclosure_system_by_prefix=[(f'tests/{s}', 'NL-INLINE-2024-GAAP-OTHER') for s in [
+        'G5-1-3_1/index.xml',
```
brettkail-wk, line 22: Would trailing `:` be more accurate? `f'tests/{s}:'`

stevenbronson-wk: It's compared against the path without the `:$variation`, like [additional_plugins_by_prefix](https://github.com/Arelle/Arelle/blob/556f0acf72d90606ceac96848f0078b8eb499aa4/tests/integration_tests/validation/conformance_suite_configurations/efm_current.py#L12-L18).  I think (agree?) that having those two kinds of IDs floating about with configuration checking one or the other is confusing.

brettkail-wk: Ah, ok.  Would trailing `/` be more accurate then, or is it used to check prefixes of both forms?

stevenbronson-wk: It's only checked against one form.  An example of the check that's eventually performed is:
```py
for prefix, candidate_disclosure_system in config.disclosure_system_by_prefix:
    if testcase_path.startswith(prefix):
    ->
    if 'tests/G5-1-3_1/index.xml'.startswith('tests/G5-1-3_1/index.xml'):
```
If you meant for the specific EFM configuration of additional_plugins_by_prefix linked above, then yes a trailing slash would be more accurate there, since it's checked against, for example, `conf/626-rendering-syntax/626-01-all-facts-filtered/626-01-all-facts-filtered-testcase.xml`, making `conf/626-rendering-syntax` a bit loose.

PR #2135, opened 2026-01-29
Update NL.NL-KVK.3.2.7.1 for 2025

aaroncameron-wk:
#### Reason for change
Update NL.NL-KVK.3.2.7.1 for 2025: Starting in 2025, non-textBlockItemType facts containing special characters like '&' or '<' must also be assigned escape="true"

#### Steps to Test
CI

**review**:
@Arelle/arelle

### DIFF
--- a/arelle/plugin/validate/NL/rules/nl_kvk.py
+++ b/arelle/plugin/validate/NL/rules/nl_kvk.py
@@ -412,10 +412,8 @@ def rule_nl_kvk_3_2_7_1 (
         for fact in facts:
             if not fact.isEscaped and (
                 fact.concept.isTextBlock or
-                len(fact) > 0 or
-                (
-                    fact.text and escapeWorthyStr.match(fact.text)
-                )
+                len(fact) > 0 or # Has child XML elements
+                (fact.text and escapeWorthyStr.match(fact.text)) # Has special characters
             ):
                 improperlyEscapedFacts.append(fact)
     if len(improperlyEscapedFacts) > 0:


arelle/plugin/validate/NL/rules/nl_kvk.py
```diff
@@ -388,16 +388,41 @@ def rule_nl_kvk_3_2_7_1 (
...
+            if fact.isEscaped != fact.concept.isTextBlock:
+                improperlyEscapedFacts.append(fact)
+    else:
+        msg = _('Ensure that any block-tagged facts of type textBlockItemType or facts '
+                'containing special characters like \'&\' or \'<\' are assigned @escape="true". '
+                'All other facts are assigned @escape="false".')
+        for fact in facts:
+            if not fact.isEscaped and (
+                fact.concept.isTextBlock or
+                len(fact) > 0 or
```
brettkail-wk: Out of curiosity, what does `len(fact) > 0` mean?

arelle/plugin/validate/NL/rules/nl_kvk.py
```diff
@@ -388,16 +388,41 @@ def rule_nl_kvk_3_2_7_1 (
...
+                'while other data types (e.g., xbrli:stringItemType) are assigned @escape="false".')
+        for fact in facts:
+            if fact.isEscaped != fact.concept.isTextBlock:
+                improperlyEscapedFacts.append(fact)
+    else:
+        msg = _('Ensure that any block-tagged facts of type textBlockItemType or facts '
+                'containing special characters like \'&\' or \'<\' are assigned @escape="true". '
+                'All other facts are assigned @escape="false".')
+        for fact in facts:
+            if not fact.isEscaped and (
```
brettkail-wk, line 413: This doesn't seem to implement the "All other" case (i.e., where `fact.isEscaped` is true).

arelle/plugin/validate/NL/rules/nl_kvk.py
aaroncameron-wk: lxml elements are treated as a list of their children when cast to an iterable, so this discovers valid XML children within the unescaped fact, which is expected to trigger this validation based on one of the testcases. Added a comment for clarity.

arelle/plugin/validate/NL/rules/nl_kvk.py
aaroncameron-wk: That appears to have been text I added to the message based on the assumption that that part of the rule carried over from 2024, but that's not the case. 2025 has a valid testcase where `Inline XBRL document with fact of xbrli:stringItemType datatype using attribute escape="true" and no special characters used in the tagged human readable content`. The implementation is correct but the message is wrong.

arelle/plugin/validate/NL/rules/nl_kvk.py
brettkail-wk: > lxml elements are treated as a list of their children when cast to an iterable

:bulb:

> XML children within the unescaped fact, which is expected to trigger this validation based on one of the testcases.

That seems wrong for overlapping tags (e.g., `<ix:nonNumeric name="tag1"><ix:nonNumeric name="tag2">Some value</ix:nonNumeric></ix:nonNumeric>`).  Can you share the snippet that is causing a problem?  I'm curious whether there's something else special about that filing, and I somewhat wonder whether the conformance suite is overaggressive and should be fixed instead.

aaroncameron-wk: For sure:
https://www.sbr-nl.nl/sites/default/files/2026-01/conformance-suite-2025-sbr-domein-handelsregister.zip
tests/G3-2-7_1/TC3_invalid.xbri
Fact ID ID0009010
`<ix:nonNumeric contextRef="context_3" name="bw2-titel9:BasisOfPreparation" id="ID0009010" escape="false">Deze grondslagen voor de jaarverslaggeving staan ook bekend als <i>Dutch GAAP</i> of <i>NL-GAAP</i>.</ix:nonNumeric>`

brettkail-wk: I wonder if the validation should exclude non-iXBRL children and also include continuations (matching [{relevant content}](https://www.xbrl.org/specification/inlinexbrl-part1/rec-2013-11-18/inlinexbrl-part1-rec-2013-11-18.html#d1e5997:~:text=order%C2%A0attribute.-,%7Brelevant%20content%7D,-The%20concatenation%20of%20the) from the iXBRL spec.)

PR #2136, opened 2026-01-29
Update NL.NL-KVK.RTS_Art_6_a for 2025

aaroncameron-wk:
#### Reason for change
The 2025 suite added a new valid testcase: "Annual report with multiple files, some with inline XBRL embedded". This suggests that a filing with inline facts in _any_ of the inline documents passes this validation, whereas the current implementation will fire for each individual inline document that does not contain facts. This correction aligns with the ESEF implementation.

This correction revealed that both the 2024 and 2025 invalid testcases for this rule have errors. Both include a hidden inline XBRL fact (making them logically identical to the valid testcase in the context of this rule), and so are not actually invalid.

#### Steps to Test
CI

**review**:
@Arelle/arelle

### DIFF
--- a/tests/integration_tests/validation/conformance_suite_configurations/nl_inline_2025.py
+++ b/tests/integration_tests/validation/conformance_suite_configurations/nl_inline_2025.py
@@ -63,6 +63,7 @@ config = ConformanceSuiteConfig(
         'G3-5-3_1/index.xml:TC2_invalid': {
             'arelle:ixdsTargetNotDefined': 1,
             'extensionTaxonomyWrongFilesStructure': 2,
+            'noInlineXbrlTags': 1,
             # This test is looking at the usage of the target attribute and does not import the correct taxonomy urls
             'requiredEntryPointNotImported': 1,
             'incorrectKvkTaxonomyVersionUsed': 1,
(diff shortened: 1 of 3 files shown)


tests/integration_tests/validation/conformance_suite_configurations/nl_inline_2025.py
```diff
@@ -263,6 +248,7 @@
         'G5-1-3_2/index.xml:TC3_invalid',  # Expects incorrectVersionEntryPointOtherGaapReferenced, should be incorrectVersionEntryPointOtherReferenced.
         'RTS_Annex_III_Par_1/index.xml:TC3_invalid',  # Expects invalidInlineXbrl, but this is valid.
         'RTS_Annex_IV_Par_12_G3-2-4_1/index.xml:TC4_invalid',  # Expects inconsistentDuplicateNonnumericFactInInlineXbrlDocumentSet, should be inconsistentDuplicateNumericFactInInlineXbrlDocument.
+        'RTS_Art_6_a/index.xml:TC3_invalid',  # Expects noInlineXbrlTags, but kvk-2025-12-31-nl.xhtml has a hidden nonNumeric fact.
```
brettkail-wk, line 251: Perhaps "embedded in the annual reports" means the check should be excluding hidden facts (i.e., those facts aren't embedded in the narrative, they're just part of the iXBRL header).

aaroncameron-wk: I had the same thought, but Lisa found it more likely to be a conformance suite error. The fact in question is a mandatory fact that isn't supposed to be allowed as hidden anyway.

allenbross-wf:
QA +10
All tests passed.

PR #2144, opened 2026-02-03
For NL inline, check for prohibited uses of dimensions.

stevenbronson-wk:
#### Reason for change
People shouldn't use dimensions on "NonDimensional" line items.

**review**:
@Arelle/arelle


derekgengenbacher-wf:

PR #2152, opened 2026-02-06
Support capturing warnings as errors via Session API

aaroncameron-wk:
#### Reason for change
While `testcaseResultsCaptureWarnings` is primarily intended for the testcase validation process to collect `WARNING` logs as errors in the `ErrorManager`, we should apply that logic consistently so that Session API users can also collect warnings as errors.

### Description of change
The [testcase validation logic](https://github.com/Arelle/Arelle/blob/bee125f3b600d68ca036675778c4f31c137223c9/arelle/Validate.py#L254-L269) results in `testcaseResultsCaptureWarnings` passing `WARNING` as `errorCaptureLevel` during instance loading, but the `CntlrCmdLine` logic [does not](https://github.com/Arelle/Arelle/blob/bee125f3b600d68ca036675778c4f31c137223c9/arelle/CntlrCmdLine.py#L1041-L1107).

#### Steps to Test
CI

**review**:
@Arelle/arelle

### DIFF
--- a/arelle/CntlrCmdLine.py
+++ b/arelle/CntlrCmdLine.py
@@ -1038,8 +1038,10 @@ class CntlrCmdLine(Cntlr.Cntlr):
             fo.testcaseExpectedErrors = options.testcaseExpectedErrors
         if options.testcaseFilters:
             fo.testcaseFilters = options.testcaseFilters
+        errorCaptureLevel = None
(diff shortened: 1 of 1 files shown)


austinmatherne-wk (approved):

derekgengenbacher-wf (approved):

derekgengenbacher-wf:

PR #2158, opened 2026-02-09
Merge PR branch into master for integration tests

aaroncameron-wk:
#### Reason for change
Using `pull_request_target` means we don't automatically get a merged version of the code to test. 

#### Description of change
We can manually merge the PR head into the base branch to ensure we are testing a merged version of the code. 

Note: This doesn't fix the gap where CI does not re-run when master changes.

#### Steps to Test

- CI on this PR will not reflect these changes.
- Tested on a PR within my fork here: https://github.com/aaroncameron-wk/arelle-public/actions/runs/21997318797?pr=172
- This doesn't test the case of a PR between a fork and a base repo, but we'll know very quickly and can revert this change if that breaks.

**review**:
@Arelle/arelle

### DIFF
--- a/.github/actions/merge/action.yml
+++ b/.github/actions/merge/action.yml
@@ -26,5 +26,5 @@ runs:
         git merge FETCH_HEAD --no-edit || {
           echo "::error::Merge conflict detected."
           git merge --abort
-          exit 1s
+          exit 1
         }


derekgengenbacher-wf (approved):

.github/actions/merge/action.yml
```diff
@@ -0,0 +1,30 @@
...
+  using: "composite"
+  steps:
+    - name: Fetch Remote SHA
+      shell: bash
+      run: |
+        git config user.name "github-actions"
+        git config user.email "github-actions@github.com"
+        
+        echo "Fetching from ${{ inputs.repository_full_name }}..."
+        git fetch https://github.com/${{ inputs.repository_full_name }}.git ${{ inputs.commit_sha }}
```
austinmatherne-wk: Both inputs are interpolated directly into a bash run block. A maliciously crafted fork name could theoretically inject shell commands. GitHub naming restrictions likely prevent anything bad from happening here, but would still be good to guard against. [GitHub says](https://docs.github.com/en/actions/reference/security/secure-use#use-an-intermediate-environment-variable) to use intermediate environment variables. Also good bash hygiene would be to quote the values.

.github/actions/merge/action.yml
```diff
@@ -11,20 +11,20 @@ inputs:
...
-    - name: Merge
-      shell: bash
-      run: |
-        echo "Merging SHA ${{ inputs.commit_sha }}..."
-        git merge FETCH_HEAD || {
+        git merge FETCH_HEAD --no-edit || {
           echo "::error::Merge conflict detected."
-          exit 1
+          git merge --abort
+          exit 1s
```
austinmatherne-wk: Is the `s` a typo?

allenbross-wf:
QA +10

PR #2165, opened 2026-02-13
EDINET: Update illegal character validations

aaroncameron-wk:
#### Reason for change
EC5003E, EC8073E, EC8073W and EC8074W attempted to use an exclusive approach to detecting illegal characters. This errantly included legal characters (The `HALF_KANA` constant we copied from `jaconv` contains full-width characters). It also looked at all documents rather than just HTML.

#### Description of change
Add a character sheet document (sourced from EDINET documentation) that includes all allowed characters. Generate a regex pattern based on these characters and apply as necessary for the relevant validation rules.

#### Steps to Test
CI

**review**:
@Arelle/arelle

### DIFF
--- a/arelle/plugin/validate/EDINET/ControllerPluginData.py
+++ b/arelle/plugin/validate/EDINET/ControllerPluginData.py
@@ -83,11 +83,11 @@ class ControllerPluginData(PluginData):
     @lru_cache(1)
     def getIllegalCharactersPattern(self) -> regex.Pattern[str]:
         allowedCharacters = set()
-        with open(self._allowedCharacterSheetPath, 'r') as file:
+        with open(self._allowedCharacterSheetPath, 'r', encoding='utf-8') as file:
             for line in file:
                 part = line.strip().split(' ')[0]
                 assert 2 <= len(part) <= 6, f"Invalid line in allowed character sheet: {line}"
-                char = bytes.fromhex(part).decode()
+                char = bytes.fromhex(part).decode(encoding='utf-8')
                 assert char not in allowedCharacters, f"Duplicate character in allowed character sheet: {line}"
                 allowedCharacters.add(char)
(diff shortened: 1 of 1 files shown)


arelle/plugin/validate/EDINET/ControllerPluginData.py
```diff
@@ -75,6 +80,18 @@ def getDeiRequirements(self, csvPath: Path, deiItems: tuple[QName, ...], filingF
     def getDeiValue(self, localName: str) -> TypeXValue:
         return self._deiValues.get(localName)
 
+    @lru_cache(1)
+    def getIllegalCharactersPattern(self) -> regex.Pattern[str]:
+        allowedCharacters = set()
+        with open(self._allowedCharacterSheetPath, 'r') as file:
```
austinmatherne-wk: Would be good to include an explicit utf-8 encoding.

arelle/plugin/validate/EDINET/rules/gfm.py
```diff
@@ -30,7 +30,7 @@
 from arelle.utils.Units import getDuplicateUnitGroups
 from arelle.utils.validate.Decorator import validation
 from arelle.utils.validate.Validation import Validation
-from ..Constants import domainItemTypeQname, HALF_KANA, JAPAN_LANGUAGE_CODES, NUMERIC_LABEL_ROLES, LC3_NAME_PATTERN, STANDARD_TAXONOMY_URL_PREFIXES
+from ..Constants import domainItemTypeQname, JAPAN_LANGUAGE_CODES, NUMERIC_LABEL_ROLES, LC3_NAME_PATTERN, STANDARD_TAXONOMY_URL_PREFIXES
```
austinmatherne-wk, line 33: Is `HALF_KANA` in Constants.py dead code that can be removed?

allenbross-wf:
QA +10
All tests passed.

PR #2166, opened 2026-02-16
Forbid multiple top-level schemas for a namespace

austinmatherne-wk:
#### Reason for change
New tests have been proposed for the XBRL 2.1 conformance suite clarifying that multi-file schemas must not have more than one top level schema.

#### Description of change
This resolves some of the proposed test cases. More test cases exist which still fail either because there are issues with the test cases themselves or the test cases contradict other 2.1 requirements. Feedback for the remaining failing test cases will be reported to the working group.

#### Steps to Test
* CI

**review**:
@Arelle/arelle

### DIFF
--- a/arelle/ValidateXbrlDTS.py
+++ b/arelle/ValidateXbrlDTS.py
@@ -1486,16 +1486,18 @@ def checkNamespaceSchemaConnectivity(val: ValidateXbrl) -> None:
                     includeNeighbors[refDoc].add(schema)
                     includedSchemas.add(refDoc)
         # Walk the undirected graph to find connected components,
-        # then count top-level schemas (not included by anything) in each.
+        # then count effective entry points per component. Each component
+        # contributes max(non-included count, 1) so that disjoint
+        # components (even fully circular ones) are detected.
         #
         # Examples:
-        #   N2->N1         : 1 component, 1 top-level (N2)          -> valid
-        #   N2->N1<-N3     : 1 component, 2 top-levels (N2, N3)     -> error
-        #   N4->N5->N6->N4 : 1 component, 0 top-levels (circular)   -> valid
-        #   A<->B, C<->D   : 2 components, 0 top-levels each        -> error
+        #   N2->N1         : 1 component, 1 non-included (N2)       -> valid
+        #   N2->N1<-N3     : 1 component, 2 non-included (N2, N3)   -> error
+        #   N4->N5->N6->N4 : 1 component, 0 non-included (circular) -> valid
+        #   A<->B, C<->D   : 2 components, 0 non-included each      -> error
         visited: set[ModelDocument] = set()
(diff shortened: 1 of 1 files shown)


arelle/ValidateXbrlDTS.py
```diff
@@ -1444,3 +1444,43 @@ def _shouldValidateBaseTaxonomyDoc(val: ValidateXbrl, modelDocument: ModelDocume
...
+    # We need to count distinct URIs to detect schemas loaded from separate
+    # paths (e.g., local linkbase ref vs HTTP xsi:schemaLocation).
+    schemaUrisByNamespace: defaultdict[str, list[str]] = defaultdict(list)
+    for uri, doc in val.modelXbrl.urlDocs.items():
+        if doc.type == ModelDocument.Type.SCHEMA and doc.targetNamespace:
+            schemaUrisByNamespace[doc.targetNamespace].append(uri)
+    for targetNamespace, uris in schemaUrisByNamespace.items():
+        if len(uris) <= 1:
+            continue
+        schemaDocs = {val.modelXbrl.urlDocs[uri] for uri in uris}
```
brettkail-wk, line 1470: Out of curiosity, why not `defaultdict[str,list[ModelDocument]]` so that this step isn't needed?

arelle/ValidateXbrlDTS.py
```diff
@@ -1444,3 +1444,43 @@ def _shouldValidateBaseTaxonomyDoc(val: ValidateXbrl, modelDocument: ModelDocume
...
+        includedSchemas = {
+            refDoc
+            for schema in schemaDocs
+            for refDoc, docRef in schema.referencesDocument.items()
+            if "include" in docRef.referenceTypes and refDoc in schemaDocs
+        }
+        topLevelSchemas = schemaDocs - includedSchemas
```
brettkail-wk: This seems to wrongly allow [A<->B, C<->D] (two disjoint sets of documents mutually including eachother).  If that's not caught by the testsuite, it seems like a testcase is missing.

arelle/ValidateXbrlDTS.py
austinmatherne-wk: I started with that approach, but switched because the URI strings are needed for both the count (Arelle's loading deduplication logic can cause a single ModelDocument to have multiple URIs) and the error message. Collecting URIs and deriving the doc set from them seemed simpler than tracking both.

arelle/ValidateXbrlDTS.py
austinmatherne-wk: Good catch. I pushed an update to address and left a comment on the draft test cases pointing out the gap in coverage.

aaroncameron-wk (dismissed):

derekgengenbacher-wf:

PR #2172, opened 2026-02-16
Break some circular references.

stevenbronson-wk:
Circular references are a bit of a problem.  Break a few that go the "wrong" way.

**review**:
@Arelle/arelle

### DIFF
--- a/arelle/typing.py
+++ b/arelle/typing.py
@@ -47,6 +47,8 @@ class LocaleDict(TypedDict):
     thousands_sep: str  # ','
     grouping: list[int]  # [3, 3, 0]
 
+# Use these for isinstance checks when importing, say, ModelDocument
+# seems excessive or might lead to circular references.
 class ModelDocumentBase: pass
 class ModelFactBase: pass
 class ModelObjectBase: pass


arelle/ModelDocumentType.py
aaroncameron-wk, line 1: Arbitrary line.

We have a handful of instances of the below that we could probably update as well.
```
from arelle.ModelDocument import ModelDocument, Type as ModelDocumentType
```

stevenbronson-wk: I missed those.  Updated, thanks.

allenbross-wf (dismissed):

arelle/typing.py
```diff
@@ -46,3 +46,11 @@ class LocaleDict(TypedDict):
     decimal_point: str  # '.'
     thousands_sep: str  # ','
     grouping: list[int]  # [3, 3, 0]
+
+class ModelDocumentBase: pass
```
austinmatherne-wk, line 52: A brief comment above these classes for why they exist and what they're used for would be appreciated.

stevenbronson-wk: Sure, added.

austinmatherne-wk (approved):

PR #2173, opened 2026-02-16
Cache ModelRelationships.

stevenbronson-wk:
Forgetting them seems wasteful.

**review**:
@Arelle/arelle

### DIFF
--- a/arelle/LinkRelationships.py
+++ b/arelle/LinkRelationships.py
@@ -1,0 +1,131 @@
+from __future__ import annotations
+
+from collections import defaultdict
+from itertools import chain
+from typing import TYPE_CHECKING, Callable, cast
+
+from arelle import XbrlConst
+from arelle.typing import ModelResourceBase, LocPrototypeBase
+
+if TYPE_CHECKING:
+    from arelle.ModelDtsObject import ModelLink, ModelRelationship, ModelResource
+    from arelle.ModelValue import QName
+    from arelle.PrototypeDtsObject import LinkPrototype, LocPrototype
+
+ModelRelationship = None  # type: ignore[assignment,misc]
+
+
+class LinkRelationships:
+    _relationships: tuple[ModelRelationship, ...] | None
+    _relationshipsByArcrole: dict[str, tuple[ModelRelationship, ...]] | None
+    _relationshipsByArcroleArcqname: dict[str, dict[QName, tuple[ModelRelationship, ...]]] | None
+    _dimensionRelationships: tuple[ModelRelationship, ...] | None
+    _formulaRelationships: tuple[ModelRelationship, ...] | None
+    _tableRenderingRelationships: tuple[ModelRelationship, ...] | None
+
+    def initRelationships(self) -> None:
+        self._relationships = None
+        self._relationshipsByArcrole = None
+        self._relationshipsByArcroleArcqname = None
+        self._dimensionRelationships = None
+        self._formulaRelationships = None
+        self._tableRenderingRelationships = None
+
+    @property
+    def relationships(self) -> tuple[ModelRelationship, ...]:
+        if self._relationships is None:
+            self._collectRelationships()
+            assert self._relationships is not None
+        return self._relationships
+
+    @property
+    def relationshipsByArcrole(self) -> dict[str, tuple[ModelRelationship, ...]]:
+        if self._relationshipsByArcrole is None:
+            self._collectRelationships()
+            assert self._relationshipsByArcrole is not None
+        return self._relationshipsByArcrole
+
+    @property
+    def relationshipsByArcroleArcqname(self) -> dict[str, dict[QName, tuple[ModelRelationship, ...]]]:
+        if self._relationshipsByArcroleArcqname is None:
+            self._collectRelationships()
+            assert self._relationshipsByArcroleArcqname is not None
+        return self._relationshipsByArcroleArcqname
+
+    @property
+    def dimensionRelationships(self) -> tuple[ModelRelationship, ...]:
+        # "XBRL-dimensions"
+        if self._dimensionRelationships is None:
+            self._dimensionRelationships = self._relationshipsForArcrolePredicate(XbrlConst.isDimensionArcrole)
+            assert self._dimensionRelationships is not None
+        return self._dimensionRelationships
+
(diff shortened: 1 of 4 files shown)


austinmatherne-wk (approved):

arelle/ModelRelationshipSet.py
```diff
@@ -144,52 +146,41 @@ def __init__(self, modelXbrl, arcrole, linkrole=None, linkqname=None, arcqname=N
...
-                            if otherRel is not USING_EQUIVALENCE_KEY: # move equivalentRel to use key instead of hasn
-                                if modelRel.isIdenticalTo(otherRel):
-                                    continue # skip identical arc
-                                relationships[otherRel.equivalenceKey] = otherRel
-                                relationships[modelRelEquivalenceHash] = USING_EQUIVALENCE_KEY
-                            modelRelEquivalenceKey = modelRel.equivalenceKey    # this is a complex tuple to compute, get once for below
-                            if modelRelEquivalenceKey not in relationships or \
-                                modelRel.priorityOver(relationships[modelRelEquivalenceKey]):
-                                relationships[modelRelEquivalenceKey] = modelRel
+            for modelRel in chain.from_iterable(linkRelationships):
```
brettkail-wk, line 169: This is fine, but I am somewhat surprised that the element traversal was a hotspot; I would have guessed the `relationships` dict.  I thought the use case we were looking at were two code paths that were calling both `modelRelationshipSet(linkrole=None)` and `modelRelationshipSet(linkrole=x)` (for each link role), where it seems like the former could be implemented by simply concatenating the latter (since `equivalenceKey` already contains the link role).  Perhaps that's a followup improvement.

arelle/PrototypeDtsObject.py
```diff
@@ -56,6 +57,7 @@ def __init__(self, modelDocument, parent, qname, role, sourceElement=None):
         if role:
             self.attributes["{http://www.w3.org/1999/xlink}role"] = role
         self.labeledResources = defaultdict(list)
+        self.initRelationships()
 
     def clear(self):
         self.__dict__.clear() # dereference here, not an lxml object, don't use superclass clear()
```
brettkail-wk, line 63: Out of curiosity, does this need to change at all?

arelle/LinkRelationships.py
```diff
@@ -0,0 +1,131 @@
...
+            for fromResource in labeledResources[fromLabel]:
+                if not isinstance(fromResource, (ModelResourceBase, LocPrototypeBase)):
+                    continue
+                fromResource = cast('ModelResource | LocPrototype', fromResource)
+                for toResource in labeledResources[toLabel]:
+                    if not isinstance(toResource, (ModelResourceBase, LocPrototypeBase)):
+                        continue
+                    toResource = cast('ModelResource | LocPrototype', toResource)
+                    fromResourceElement = fromResource.dereference()  # type: ignore [no-untyped-call]
+                    toResourceElement = toResource.dereference()  # type: ignore [no-untyped-call]
```
brettkail-wk, line 116: Do we need to do anything to handle multi-target to avoid referencing "stale" model objects (XSD data types for concepts or resources)?  I think the `dereference` is implicitly based on `modelXbrl`, and I thought there was some multi-target code path that resets `modelXbrl` so that `ModelConcept` can be updated.  (Perhaps that multi-target code calls `clear`, and this question is the same as the other.)

arelle/LinkRelationships.py
aaroncameron-wk, line 1: @austinmatherne-wk Since this is new API, does it make sense to have this implemented in `arelle.api` with the appropriate code style for the public API? Or we could just wrap this later like everything else.

Separately, since this is used via inheritance, I think a name like `RelatableLink` or simply `Link` would be more appropriate. `LinkRelationships` suggests more of a composition relationship (`ModelLink` _has a_ `LinkRelationships` vs. `ModelLink` _is a_ `Link`).

austinmatherne-wk: > Or we could just wrap this later like everything else.

Wrapper makes sense to me.

arelle/ModelDtsObject.py
```diff
@@ -1738,10 +1738,10 @@ class ModelRelationship(ModelObject):
     """
     _equivalenceHash: int | None
 
-    def __init__(self, modelDocument, arcElement, fromModelObject, toModelObject):
+    def __init__(self, modelDocument, arcElement, fromModelObject, toModelObject, linkrole=None):
         # copy model object properties from arcElement
         self.arcElement = arcElement
-        self.linkrole = arcElement.getparent().get("{http://www.w3.org/1999/xlink}role")
+        self.linkrole = linkrole if linkrole is not None else arcElement.getparent().get("{http://www.w3.org/1999/xlink}role")
```
volodymyrtur-wk, line 1744: `self.linkrole = arcElement.getparent().get("{http://www.w3.org/1999/xlink}role") if linkrole is None else linkrole`
The result is the same, but as for me, this boolean logic is easier to understand

arelle/ModelDtsObject.py
```diff
@@ -1602,14 +1603,13 @@ class ModelLink(ModelObject):
     :type modelDocument: ModelDocument
     """
```
volodymyrtur-wk, line 1604: You forgot to update the docstring

arelle/ModelRelationshipSet.py
stevenbronson-wk: It was and is primarily constructing new ModelRelationships and accessing the attributes/elements required for equivalenceHash.

The scenario was multiple arcroles without a linkrole and then specific arcroles or linkroles, but extending your suggestion to per base set: yes, that would be an improvement, and we could improve combining other ModelRelationshipSet fields like modelRelationshipsFrom, but the profile was less excited about those.

arelle/PrototypeDtsObject.py
stevenbronson-wk: No, it's the same "self" in LinkRelationships, so those attributes are cleared as well.

arelle/LinkRelationships.py
stevenbronson-wk: No, new ModelLinks and a new ModelXbrl are created for each target.

Issue #720, opened 2023-06-06
Set exit code to reflect validation status when running with -v on command line


paulwarren-wk:
### What should we change and why?

The `-v` command line option allows you to validate XBRL files, but there's no easy way to tell the overall validation result (i.e. was the file valid or not?)

It would be very helpful if Arelle set the exit code to reflect the overall validation status, and also if the log output recorded this too.  For example, here is some typical output:

```
[info] loaded in 3.09 secs at 2023-06-06T16:22:21 - /builds/taxonomies/iso-code-lists/build/country-WGWD-YYYY-MM-DD/countryList.xsd 
[xbrl.3.5.4:hrefIdNotFound] Href countryList-elts.xsd#country_CountryDomain not located - country-definition.xml 39
[xbrl.5.2.6.1:definitionLinkLocTarget] definitionLink loc href countryList-elts.xsd#country_CountryDomain must identify a concept - country-definition.xml 39
[info] validated in 1.10 secs - /builds/taxonomies/iso-code-lists/build/country-WGWD-YYYY-MM-DD/countryList.xsd 
```

This is clearly invalid, but the last line doesn't state this.

Setting the exit code would make it easy to use Arelle for validating outputs in a CI process.

It would be good if this differentiated between the different validation levels (i.e. spec violations vs formula/calculation/filing rule errors)

richieeeeeeee:
That would be a great improvement.

PR #2191, opened 2026-02-24
Add validation exit code option and update exit behavior based on validation results

volodymyrtur-wk:
#### Reason for change
The GitHub [issue](https://github.com/Arelle/Arelle/issues/720)

#### Description of change
This pull request adds a new command-line option to Arelle for returning a specific exit code based on the validation result, and refactors the codebase to consistently track validation success or failure. The main change is the introduction of the `--validationExitCode` option, which enables the CLI to exit with code 3 when validation errors or warnings are found, improving integration with automated systems.

**New CLI Feature:**

* Added the `--validationExitCode` command-line option, which causes the CLI to exit with code 3 if validation errors or warnings are detected, and 0 if the validation is successful. This helps external scripts and CI systems detect validation failures programmatically.

**Refactoring for Validation State Tracking:**

* Introduced `validation_exit_code` and `is_successful` attributes to the `CntlrCmdLine` class, and refactored the validation logic throughout the codebase to use `self.is_successful` instead of a local `success` variable. This ensures the validation state is consistently tracked and accessible for determining the exit code. 

**Other Technical Improvements:**

* Updated the `RuntimeOptions` dataclass to include the new `validationExitCode` attribute, ensuring type safety and correct propagation of the option.
* Minor formatting and whitespace adjustments for code clarity.

These changes make it easier to automate validation workflows and improve the maintainability of the CLI validation logic.

#### Steps to Test
CI
**review**:
@Arelle/arelle

### DIFF
--- a/tests/integration_tests/scripts/tests/validation_exit_code.py
+++ b/tests/integration_tests/scripts/tests/validation_exit_code.py
@@ -78,6 +78,7 @@ result = run_arelle_cmd(
         "--validate",
         "--validationExitCode",
         "--webserver",
+        "locahost:8080",
         ],
     offline=arelle_offline,
     logFile=arelle_log_file,


arelle/CntlrCmdLine.py
```diff
@@ -84,7 +84,11 @@ def main():
     args = shlex.split(envArgs) if envArgs else sys.argv[1:]
     setApplicationLocale()
     gettext.install("arelle")
-    parseAndRun(args)
+    ctrl = parseAndRun(args)
+
+    if ctrl.validation_exit_code:
+        if ctrl.errors or not ctrl.is_successful:
```
austinmatherne-wk: I don't think this quite works for all errors. There's [an `ErrorManager` on the controller](https://github.com/Arelle/Arelle/blob/a0de19275145c9ce5b9ea8559491ccb184e93129/arelle/Cntlr.py#L300) for filesource level errors, but there's also [an `ErrorManager` on the models](https://github.com/Arelle/Arelle/blob/a0de19275145c9ce5b9ea8559491ccb184e93129/arelle/ModelXbrl.py#L322) themselves for model level errors.

arelle/CntlrCmdLine.py
```diff
@@ -723,7 +736,9 @@ class CntlrCmdLine(Cntlr.Cntlr):
 
     def __init__(self, logFileName=None, uiLang=None, disable_persistent_config=False):
         super().__init__(hasGui=False, uiLang=uiLang, disable_persistent_config=disable_persistent_config, logFileName=logFileName)
-        self.preloadedPlugins =  {}
+        self.preloadedPlugins = {}
+        self.validation_exit_code = False
+        self.is_successful = False
```
austinmatherne-wk: This codebase historically uses `camelCase` throughout. Per [PEP 8](https://peps.python.org/pep-0008/#a-foolish-consistency-is-the-hobgoblin-of-little-minds), consistency within a module matters more than the style guide or standard Python conventions, so please use `camelCase` for new attributes within existing modules (e.g. `validationExitCode`, `isSuccessful`). We're happy to use `snake_case` in brand new modules and APIs, but want to keep each file internally consistent. Thanks!

arelle/CntlrCmdLine.py
```diff
@@ -491,6 +495,14 @@ def parseArgs(args):
                       action="store", dest="optionsFile",
                       help=_("Provide a path to a JSON file containing runtime options. "
                              "These options will be overridden by any command line options provided."))
+    parser.add_option("--validationExitCode",
```
austinmatherne-wk, line 512: I think if a user attempted to use both `--webserver` and `--validationExitCode` together it would cause the server to exit upon processing a filing with an error? Using them together doesn't really make sense though. Can you add a call to `parser.error` if a user tries to use both of them together informing them that `--validationExitCode` can't be used with `--webserver`?

arelle/CntlrCmdLine.py
```diff
@@ -734,6 +749,8 @@ def run(self, options: RuntimeOptions, sourceZipStream=None, responseZipStream=N
         :param options: OptionParser options from parse_args of main argv arguments (when called from command line) or corresponding arguments from web service (REST) request.
         :type options: optparse.Values
         """
+        self.validation_exit_code = options.validationExitCode
```
austinmatherne-wk: This codebase historically uses `camelCase` throughout. Per [PEP 8](https://peps.python.org/pep-0008/#a-foolish-consistency-is-the-hobgoblin-of-little-minds), consistency within a module matters more than the style guide or standard Python conventions, so please use `camelCase` for new attributes within existing modules (e.g. `validationExitCode`, `isSuccessful`). We're happy to use `snake_case` in brand new modules and APIs, but want to keep each file internally consistent. Thanks!

arelle/CntlrCmdLine.py
austinmatherne-wk: `is_successful` doesn't get reset between runs does it? If the same controller is [reused for multiple runs](https://github.com/Arelle/Arelle/blob/a0de19275145c9ce5b9ea8559491ccb184e93129/arelle/CntlrWebMain.py#L380), this field would carry over stale state from a previous run?

arelle/RuntimeOptions.py
```diff
@@ -157,6 +157,7 @@ class RuntimeOptions:
     viewFile: Optional[str | FileNamedStringIO] = None
     webserver: Optional[str] = None
     xdgConfigHome: Optional[str] = None
+    validationExitCode: bool = False
```
austinmatherne-wk, line 160: arbitrary line:

It's not required for this PR, but if you want to dig into how our integration tests run, you could setup a test to verify the exit codes similar to [the test here](https://github.com/Arelle/Arelle/blob/a0de19275145c9ce5b9ea8559491ccb184e93129/tests/integration_tests/scripts/tests/ixbrl-viewer_cli.py).

arelle/CntlrCmdLine.py
aaroncameron-wk: Unfortunately those models will be [closed](https://github.com/Arelle/Arelle/blob/a0de19275145c9ce5b9ea8559491ccb184e93129/arelle/CntlrCmdLine.py#L1358-L1362) at this point without `--keepOpen`, and it doesn't appear that `success` (or `self.is_successful`) is affected by the presence of model errors. We may need `if modelXbrl.errors: success = False` or similar.

arelle/CntlrCmdLine.py
volodymyrtur-wk: Based on the link, it looks like the web server uses the same instance of `CntlrCmdLine`, similar to the `Singleton` pattern. In this case, all `self` will share the state from a previous run

arelle/CntlrCmdLine.py
volodymyrtur-wk: I ran some tests, and in my case, the `modelManager` doesn't contain an `ErrorManager`. The `CntlrCmdLine.errors` is the same as `ErrorManager.errors`.

arelle/CntlrCmdLine.py
austinmatherne-wk: Yep. The controller can be reused across multiple runs, so storing run state on `self` is problematic (there's config state that's intentionally shared). `is_successful` is ambiguous (last run was successful? all runs have been successful?) and will bleed between invocations. The existing return value already signaled success, so we should be able to use that instead of adding new class state.

arelle/CntlrCmdLine.py
austinmatherne-wk: The `ModelManager` doesn't include an `ErrorManager` itself, only the specific `ModelXbrl` classes do. You should be able to get to the XBRL models from model manager (while they're still open/loaded to Aaron's point) using `modelManager.loadedModelXbrls`

volodymyrtur-wk: Aaron's suggestion `if modelXbrl.errors: success = False` is already implemented in a few places. In this case, I think we can rely on current `success` calculations, and the option `--keepOpen` shouldn't impact the result

aaroncameron-wk: I think most of where you're seeing `success = False` set in `CntlrCmdLine` are more edge cases. The happy path makes it to the main validation step [here](https://github.com/Arelle/Arelle/blob/master/arelle/CntlrCmdLine.py#L1209). That validation module is where most errors are added to the model. After that, there is no check that sets `success = False` based on the presence of model errors.

We could either add one immediately after that validation call, or it might be more robust to add it immediately before anywhere the model is closed.

arelle/CntlrCmdLine.py
```diff
@@ -491,6 +495,14 @@ def parseArgs(args):
                       action="store", dest="optionsFile",
                       help=_("Provide a path to a JSON file containing runtime options. "
                              "These options will be overridden by any command line options provided."))
+    parser.add_option("--validationExitCode",
+                      action="store_true",
+                      default=False,
+                      dest="validationExitCode",
+                      help=_("Display a code indicating validation result: "
```
austinmatherne-wk: Missed this on the first pass, but the help text says `Display a code` but this option doesn't display anything, it sets the process exit code.

arelle/CntlrCmdLine.py
```diff
@@ -656,7 +672,12 @@ def configAndRunCntlr(options, arellePluginModules):
                            logXmlMaxAttributeLength=options.logXmlMaxAttributeLength,
                            logPropagate=options.logPropagate)
         cntlr.postLoggingInit()  # Cntlr options after logging is started
-        cntlr.run(options)
+        result = cntlr.run(options)
+
+        if validationExitCode := options.validationExitCode:
+            cntlr.validationExitCode = validationExitCode
+            cntlr.isSuccessful = result
```
austinmatherne-wk: This still populates run specific state on the controller, which has a different lifecycle. The webserver check keeps this out of the current reuse paths, but there aren't any tests preventing a refactor or new feature from changing that expectation. We can avoid the issue entirely by threading the result through to `main()` using new internal functions. `parseAndRun` and `configAndRunCntlr` are "public" and may have external callers ([such as XuleServer](https://github.com/xbrlus/xule/blob/40f774ced269ee6637a96d51db244efd6337e689/plugin/xule/XuleServer.py#L61)), so we should preserve their signatures to avoid introducing a breaking change:

```python
def _configAndRunCntlr(options, arellePluginModules):
    # existing configAndRunCntlr code
    return cntlr, result

def configAndRunCntlr(options, arellePluginModules):
    cntlr, _ = _configAndRunCntlr(options, arellePluginModules)
    return cntlr

def _parseAndRun(args):
    runtimeOptions, arellePluginModules = parseArgs(args)
    cntlr, result = _configAndRunCntlr(runtimeOptions, arellePluginModules)
    return cntlr, runtimeOptions, result

def parseAndRun(args):
    cntlr, _, _ = _parseAndRun(args)
    return cntlr

def main():
    ...
    cntlr, options, result = _parseAndRun(args)
    if options.validationExitCode and not result:
        sys.exit(3)
```

This keeps the exit code logic in `main()` using local variables, avoids adding run-specific state to the controller, and preserves backward compatibility.

arelle/CntlrCmdLine.py
```diff
@@ -1355,6 +1377,13 @@ def monitorParentProcess():
                 modelXbrl.profileStat(_("total"), time.time() - firstStartedAt)
                 if options.collectProfileStats and modelXbrl:
                     modelXbrl.logProfileStats()
+
+                if modelDiffReport and modelDiffReport.errors:
+                    success = False
+
+                if modelXbrl and modelXbrl.errors:
+                    success = False
+
```
austinmatherne-wk, line 1407: This sets `success` to `False` unconditionally, even without `--validationExitCode` set. That means for all users, not just those opting into the new flag:

1. Since it's inside the entrypoint loop, a validation error in an earlier entrypoint will set `success = False` for all subsequent entrypoints, changing their processing behavior (anywhere `if success` gates logic within the loop).
2. After the loop, the `CntlrCmdLine.Filing.Validate` and `CntlrCmdLine.Filing.End` plugin hooks are gated by `if success`, this will now prevent those hooks from running on validation errors.
3. The return value of `run()` changes for existing callers who don't opt in to the new `validationExitCode` option.

Track validation errors separately so existing behavior is untouched:
```python
# initialize near top of run() alongside success
hasValidationErrors = False

# inside the loop (replacing the new success = False assignments)
if modelDiffReport and modelDiffReport.errors:
    hasValidationErrors = True

if modelXbrl and modelXbrl.errors:
    hasValidationErrors = True

# at the end
return success and not (options.validationExitCode and hasValidationErrors)
```

arelle/CntlrCmdLine.py
volodymyrtur-wk: I'll change it. I wrote it when I thought the exit code needed to be printed out.

arelle/CntlrCmdLine.py
volodymyrtur-wk: I've got your point. I changed the controller like a monkey patch. It only affects that specific controller instance that was modified and has no impact on other instances. I tried to make the minimal changes possible to avoid issues if someone had used these functions.

arelle/CntlrCmdLine.py
```diff
@@ -1406,7 +1456,7 @@ def showStatusOnPipe(self, message, clearAfter=None):
             except Exception as ex:
                 #with open("Z:\\temp\\trace.log", "at", encoding="utf-8") as fh:
                 #    fh.write("Status pipe exception {} {}\n".format(type(ex), ex))
-                system.exit()
+                sys.exit()
```
austinmatherne-wk, line 1459: Good catch!

arelle/CntlrCmdLine.py
```diff
@@ -84,19 +85,30 @@ def main():
     args = shlex.split(envArgs) if envArgs else sys.argv[1:]
     setApplicationLocale()
     gettext.install("arelle")
-    parseAndRun(args)
+    cntlr, options, result = _parseAndRun(args)
```
austinmatherne-wk, line 88: arbitrary line:

Looks good! Mypy is failing in CI (can run locally using `mypy arelle tests --namespace-packages`) the` # type: ignore[no-untyped-call]` [on this line](https://github.com/Arelle/Arelle/blob/c572e7f4771f6ac37927921dce89eaf9057e8182/tests/integration_tests/integration_test_util.py#L95) is now unused since `parseAndRun` has a return type annotation. Just need to remove that type ignore comment.

@aaroncameron-wk any idea why the new merge action [is failing](https://github.com/Arelle/Arelle/actions/runs/22459098924/job/65048197906?pr=2191) with `Error: Merge conflict detected` when the PR doesn't show a merge conflict?

tests/integration_tests/scripts/tests/validation_exit_code.py
```diff
@@ -0,0 +1,93 @@
...
+    errors.append(f"Errors found. Validation finished with exit code {result.returncode}. Expected 3.")
+
+result = run_arelle_cmd(
+    arelle_command,
+    additional_args=[
+        "-f", str(warning_target_path),
+        "--validate",
+        "--validationExitCode",
+        "--webserver",
+        ],
```
austinmatherne-wk, line 82: `--webserver` isn't provided with a value (such as `localhost:8080`). Maybe why ci is failing on windows?

volodymyrtur-wk: Yes, you're right. For Windows, there's a separate functionality which prepares `args` and puts after `--webserver` an empty string, which is equal to `False`. For the rest of the OSs, we take the next parameter right after `--webserver`, and it allows us to pass the condition `if options.webserver and options.validationExitCode:`, which causes the error that we expected

austinmatherne-wk (dismissed):

PR #2193, opened 2026-02-25
Adding EDINET validation EC8070W.

allenbross-wf:
#### Description of change

Added EDINET validation EC8070W.

#### Steps to Test

**review**:
@Arelle/arelle


arelle/plugin/validate/EDINET/rules/edinet.py
```diff
@@ -2083,6 +2073,55 @@ def rule_EC8069W_2(
...
+        if fact.context is None:
+            continue
+        if len(fact.context.qnameDims) > 0:
+            if fact.context.qnameDims.keys() != {pluginData.directorsAndOtherOfficersAxisQn}:
+                continue
+            memberUsed = fact.context.dimMemberQname(pluginData.directorsAndOtherOfficersAxisQn)
+            for rel in relationships.fromModelObject(parentMember):
+                if rel.toModelObject.qname == memberUsed:
+                    factFound = True
+                    break
```
aaroncameron-wk: `fromModelObject` caches data internally so this isn't a huge deal, but it might be cleaner to collect a set of the valid members once earlier on rather than performing this iteration every time. i.e.:
```
directorsAndOtherOfficersMembers = {
	rel.toModelObject.qname for rel in relationships.fromModelObject(parentMember)
}
```
Then here, just:

```suggestion
			if memberUsed in directorsAndOtherOfficersMembers:
				break
```

arelle/plugin/validate/EDINET/rules/edinet.py
```diff
@@ -2083,6 +2073,55 @@ def rule_EC8069W_2(
         )
 
 
+@validation(
+    hook=ValidationHook.XBRL_FINALLY,
+    disclosureSystems=[DISCLOSURE_SYSTEM_EDINET],
+)
+def rule_EC8070E(
```
austinmatherne-wk: Function should be named `rule_EC8070W`?

arelle/plugin/validate/EDINET/rules/edinet.py
```diff
@@ -2083,6 +2073,55 @@ def rule_EC8069W_2(
...
+    relationships = val.modelXbrl.relationshipSet(XbrlConst.domainMember)
+    for fact in facts:
+        if fact.context is None:
+            continue
+        if len(fact.context.qnameDims) > 0:
+            if fact.context.qnameDims.keys() != {pluginData.directorsAndOtherOfficersAxisQn}:
+                continue
+            memberUsed = fact.context.dimMemberQname(pluginData.directorsAndOtherOfficersAxisQn)
+            for rel in relationships.fromModelObject(parentMember):
+                if rel.toModelObject.qname == memberUsed:
```
austinmatherne-wk: `rel.toModelObject` could be `None`

arelle/plugin/validate/EDINET/rules/edinet.py
```diff
@@ -2083,6 +2073,55 @@ def rule_EC8069W_2(
...
+        return
+    facts = pluginData.iterValidNonNilFacts(val.modelXbrl, pluginData.totalAmountOfRemunerationEtcPaidByGroupRemunerationEtcPaidByGroupToEachDirectorOrOtherOfficerQn)
+    factFound = False
+    parentMember = val.modelXbrl.qnameConcepts.get(pluginData.directorsAndOtherOfficersMemberQn)
+    relationships = val.modelXbrl.relationshipSet(XbrlConst.domainMember)
+    for fact in facts:
+        if fact.context is None:
+            continue
+        if len(fact.context.qnameDims) > 0:
+            if fact.context.qnameDims.keys() != {pluginData.directorsAndOtherOfficersAxisQn}:
```
austinmatherne-wk: Does the rule specifically require this single dimension, or would it be valid to tag against this and another dimension?

arelle/plugin/validate/EDINET/rules/edinet.py
austinmatherne-wk: This also only breaks out of the inner loop. The outer loop continues unnecessarily checking the remaining facts.

Instead of tracking `factFound` as a var, we could return as soon as a matching fact is found?

arelle/plugin/validate/EDINET/rules/edinet.py
```diff
@@ -2083,6 +2074,60 @@ def rule_EC8069W_2(
...
+        return None
+
+    if not pluginData.shouldContainExecutiveDetails(val.modelXbrl):
+        return
+    facts = pluginData.iterValidNonNilFacts(val.modelXbrl, pluginData.totalAmountOfRemunerationEtcPaidByGroupRemunerationEtcPaidByGroupToEachDirectorOrOtherOfficerQn)
+    parentMember = val.modelXbrl.qnameConcepts.get(pluginData.directorsAndOtherOfficersMemberQn)
+    relationships = val.modelXbrl.relationshipSet(XbrlConst.domainMember)
+    directorsAndOtherOfficersMembers = {
+        rel.toModelObject.qname for rel in relationships.fromModelObject(parentMember)
+    }
```
austinmatherne-wk, line 2123: `rel.toModelObject` can be `None`:
```suggestion
    directorsAndOtherOfficersMembers = {
        rel.toModelObject.qname for rel in relationships.fromModelObject(parentMember)
        if rel.toModelObject is not None
    }
```

arelle/plugin/validate/EDINET/rules/edinet.py
```diff
@@ -2083,6 +2074,60 @@ def rule_EC8069W_2(
...
+    DirectorsAndOtherOfficersAxis and that member needs to be a child of DirectorsAndOtherOfficersMember
+    """
+    def matchingFact(facts: Iterable[ModelFact], dimensionQn: QName, validMembers: set[QName]) -> ModelFact | None:
+        for fact in facts:
+            context = fact.context
+            if context is None:
+                continue
+            if len(context.qnameDims) == 0:
+                continue
+            if set(context.qnameDims.keys()) != {dimensionQn}:
```
austinmatherne-wk: set wrapping is unnecessary, `dict.keys()` natively supports set comparison:
```suggestion
            if context.qnameDims.keys() != {dimensionQn}:
```

But my earlier question is still open. Does the rule require exactly this single dimension, or would a fact tagged with this dimension and additional dimensions also be valid? If so, this should be:
```suggestion
            if dimensionQn not in context.qnameDims:
```

allenbross-wf: I'm sorry, I had thought I responded.  The rule requires a single dimension.

arelle/plugin/validate/EDINET/rules/edinet.py
allenbross-wf: I have confirmed it requires a single dimension.

PR #2209, opened 2026-03-01
Optimize Solvency II Formula Processing

austinmatherne-wk:
#### Reason for change
Formula processing for Solvency II reports is slow. These changes reduce formula validation time on an internal solvency II report by ~60% (6 min → 2:25 min on M1 MacBook Pro).

#### Description of change
 - Cache `ModelTypedDimension.dimQname` and `dimQnameExpression` to avoid repeated XML lookups.
  - Skip context aspect comparisons when two facts share the same context object.

#### Steps to Test
* CI

**review**:
@Arelle/arelle

### DIFF
--- a/arelle/formula/FormulaEvaluator.py
+++ b/arelle/formula/FormulaEvaluator.py
@@ -12,7 +12,7 @@ from math import fabs, isinf, isnan, log10
 import regex as re
 
 from arelle import XbrlConst, XbrlUtil, XmlUtil
-from arelle.Aspect import Aspect, aspectModels, aspectModelAspect
+from arelle.Aspect import Aspect, aspectContextAspects, aspectModels, aspectModelAspect
 from arelle.FunctionXs import xsString
 from arelle.Locale import format_string
 from arelle.ModelFormulaObject import (
@@ -934,6 +934,17 @@ def aspectsMatch(xpCtx, fact1, fact2, aspects):
                 return False
             if matches is None:
                 uncachedAspects.append(aspect)
+    if uncachedAspects and fact1 is not None and fact2 is not None:
+        c1 = fact1.context
+        if c1 is not None and c1 is fact2.context:
+            uncachedAspects = [
+                a for a in uncachedAspects
+                if a not in aspectContextAspects
+                # context derived dimension aspects are QNames
+                and not isinstance(a, QName)
+            ]
+            if not uncachedAspects:
+                return True
     prioritizedAspectCache = factAspectsCache.prioritizedAspects
     prioritizedAspects = sorted(uncachedAspects, key=lambda a: a not in prioritizedAspectCache)
     for aspect in prioritizedAspects:
--- a/arelle/Aspect.py
+++ b/arelle/Aspect.py
@@ -58,6 +58,16 @@ aspectRuleAspects = {   # aspect correspondence to rule-retrievable aspects
     Aspect.UNIT: (Aspect.UNIT_MEASURES, Aspect.MULTIPLY_BY, Aspect.DIVIDE_BY)
 }
 
+aspectContextAspects = frozenset({
+    Aspect.ENTITY_IDENTIFIER, Aspect.VALUE, Aspect.SCHEME,
+    Aspect.PERIOD, Aspect.PERIOD_TYPE, Aspect.START, Aspect.END, Aspect.INSTANT, Aspect.INSTANT_END,
+    Aspect.COMPLETE_SEGMENT,
+    Aspect.COMPLETE_SCENARIO,
+    Aspect.NON_XDT_SEGMENT,
+    Aspect.NON_XDT_SCENARIO,
+    Aspect.DIMENSIONS, Aspect.OMIT_DIMENSIONS,
+})
+
 aspectModels = {
     "dimensional": {  # order by likelyhood of short circuting aspect match tests
         Aspect.CONCEPT, Aspect.PERIOD, Aspect.UNIT, Aspect.LOCATION, Aspect.ENTITY_IDENTIFIER,


arelle/formula/FormulaEvaluator.py
```diff
@@ -784,21 +777,15 @@ def trialFilterFacts(xpCtx, vb, facts, filterRelationships, filterType, var=None
         facts = factSet
 
     # handle now simple model type complement (at the end since it appears to reduce much less)
-    if filterInfo is not None:
-        if filterInfo and len(complHandledFilterRels) > 0:
-            for varFilterRel, dimQname in complHandledFilterRels:
+    if filterInfo and complHandledFilterRels:
+        if not noComplHandledFilterRels:
+            facts = {f for f in facts if f.isItem}
```
volodymyrtur-wk: A more Pythonic way to write this is:
```facts = {fact for fact in facts if fact.isItem}```

File: arelle/formula/FormulaEvaluator.py [hunk 2]
```diff
@@ -732,23 +732,16 @@ def trialFilterFacts(xpCtx, vb, facts, filterRelationships, filterType, var=None
...
-                if fact.isItem:
-                    for varFilterRel, dimQname in noComplHandledFilterRels:
-                        dim = fact.context.qnameDims.get(dimQname)
-                        if dim is not None:
-                            outFacts.add(fact)
-            facts = outFacts
-            if len(facts) == 0:
+                vb.aspectsCovered |= _filter.aspectsCovered(vb)
+            dimedFacts = set.union(*[inst.factsByDimMemQname(dimQname) for inst in vb.instances])
+            facts = facts & dimedFacts
```
brettkail-wk: Out of curiosity, why is this faster?  My guess would have been that `noComplHandledFilterRels` is small, so this is typically iterating the set of facts once and removing the non-matches.  Pre-existing, I guess the code could have a `break` after `outFacts.add`?

The new code seems like it would repeatedly rebuild `set.union(*[...])`.  It seems like that might be faster if that `set.union` is often empty (or much smaller than `facts`), so the `&` can iterate the smaller of the two sets, but it seems like it would be worse if `factsByDimMemQname` is large since it would have to repeatedly convert ModelXbrl set->list->set.union?

I wonder if it's better to only apply these special cases in the PR only when `len(vb.instances) == 1`, and then you can directly use the `inst.factsByDimMemQname` `set` rather than needing the `set.union`.

arelle/formula/FormulaEvaluator.py
```diff
@@ -934,6 +921,16 @@ def aspectsMatch(xpCtx, fact1, fact2, aspects):
                 return False
             if matches is None:
                 uncachedAspects.append(aspect)
+    if uncachedAspects and fact1 is not None and fact2 is not None:
+        c1 = fact1.context
+        if c1 is not None and c1 is fact2.context:
+            uncachedAspects = [
+                a for a in uncachedAspects
+                if a not in aspectContextAspects
+                and not isinstance(a, QName)
```
brettkail-wk, line 944: Out of curiosity, why this `isinstance`?

arelle/ModelFormulaObject.py
```diff
@@ -2127,17 +2127,21 @@ def aspectsCovered(self, varBinding):
 
     @property
     def dimQname(self):
-        dimQname = XmlUtil.descendant(self, XbrlConst.df, "qname")
-        if dimQname is not None:
-            return qname( dimQname, XmlUtil.text(dimQname) )
-        return None
+        try:
+            return self._dimQname
+        except AttributeError:
```
brettkail-wk: This is completely fine.  I know this pattern is pervasive in Arelle, but out of curiosity, why do we use a separate `_x` property rather than directly assigning `self.dimQname` (I think(?) `@cached_property` does effectively that) so that the next access can directly use the property rather than invoke the func and fetch the `_x` property?

arelle/formula/FormulaEvaluator.py
austinmatherne-wk: The `factsByDimMemQname` index is built once per dimension on first access and cached on `ModelXbrl`. After that, each call is an O(1) dict lookup returning the prebuilt set which can then leverage the benefits from set operations running at the C level instead of interpreted Python.

The `set.union(*[...])` syntax unpacks a list of references to those existing cached sets as positional args (ModelXbrl `set[ModelFact]`->`list[set[ModelFact]]`->`set[ModelFact]` union). I replaced the list with a generator. It doesn't make a measurable performance difference, but good practice anyway.

I added the suggested `len(vb.instances) == 1` guard to skip the union call for single sets entirely in the common instance case (maybe ~0.2% improvement, but within the margin of noise).

arelle/formula/FormulaEvaluator.py
austinmatherne-wk: For context derived dimensions. I added a comment.

arelle/ModelFormulaObject.py
austinmatherne-wk: I'm not sure if that'll work for other classes which define `__slots__`, but there's no good reason here. I've switched to cached_property.

arelle/formula/FormulaEvaluator.py
brettkail-wk: I'm still worried about scenarios like `len(facts) == 1` but `len(factsByDimMemQname) > 10000` or whatever, where that set->list->set has far worse performance then the original code.  It still seems like we should do something like this?
```
if len(vb.instances) == 1:
    foreach dimQName:
        facts = facts & vb.instances[0].factsByDimMemQname(dimQname)
else:
    outFacts = {}
    foreach fact:
         if fact has any of the dims:
              outFacts.add(fact);
facts = outFacts
```
Or perhaps something like this works?
```
foreach dimQName:
    foreach instance:
        facts = facts & instance.factsByDimMemQname(dimQname)
```

austinmatherne-wk: These changes only accounted for 3% of the performance improvements, so not worth the complexity and testing for performance regressions with other taxonomies and reports. I might revisit this in the future, but backed them out of this PR.

PR #2215, opened 2026-03-02
Integration tests checkout latest master and fetch appropriate history for merge

aaroncameron-wk:
#### Reason for change
- `actions/checkout` doesn't necessarily checkout the latest commit of the base branch, in many scenarios it will checkout the commit that the PR branch originally branched from.
- The shallow fetch is causing merge failures due to unrelated histories.

#### Description of change
- Pass the base branch name to `actions/checkout` so it checks out the HEAD of the base branch.
- Incrementally fetch additional history until a shared ancestor is found (or a high limit is reached).

#### Steps to Test
Difficult to test all scenarios, but basic merge was successful when testing on my fork.
CI failing due to issue that this resolves.

**review**:
@Arelle/arelle

### DIFF
--- a/.github/actions/merge/action.yml
+++ b/.github/actions/merge/action.yml
@@ -4,6 +4,10 @@ inputs:
   commit_sha:
     description: 'The specific commit SHA to merge'
     required: true
+  fetch_depth:
+    description: 'The depth of commit history to fetch'
+    required: true
+    default: 100
   repository_full_name:
     description: 'The full name of the repository to fetch from (e.g., owner/repo)'
     required: true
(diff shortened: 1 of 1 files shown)


.github/workflows/integration-tests.yml
```diff
@@ -43,6 +44,7 @@ jobs:
       - uses: actions/checkout@v6.0.2
         with:
           persist-credentials: false
+          ref: ${{ github.base_ref }}  # Forces the tip of the branch, not the event SHA
```
austinmatherne-wk, line 47: The run jobs also need this?

.github/actions/merge/action.yml
```diff
@@ -20,11 +20,28 @@ runs:
...
+        STEP=32
+        TOTAL_FETCHED=0
+        MAX_HISTORY=1000 
+      
+        # Attempt merge. On failure, fetch more history and try again until we either succeed or reach our history limit
+        until git merge FETCH_HEAD --no-edit; do
+          EXIT_CODE=$?
+        
+          # If exit is 128, we are likely missing the merge base
+          if [ $EXIT_CODE -eq 128 ] && [ $TOTAL_FETCHED -le $MAX_HISTORY ]; then
```
austinmatherne-wk: 128 isn't specific to missing merge base, is it?

Would we be better off using `fetch-depth: 0` (or something reasonably deep) on checkout instead of this?

aaroncameron-wk: > 128 isn't specific to missing merge base, is it?

It is not, but inspecting the error message for the exact cause adds complexity to something that already feels overly complex.

> Would we be better off using fetch-depth: 0 (or something reasonably deep) on checkout instead of this?

Fetching full history adds a significant amount of time. More than a minute, which doesn't seem like a ton but since we need to approve some of our workflows it's a bit of a pain to sit and wait for that. Similarly, fetching a fixed amount of history is a balance between adding a baseline delay to every PR vs. how old a PR can be before merging breaks. 

Thinking about it more, expecting contributors to rebase old PR's doesn't seem unreasonable. So I'm OK with maybe 100? A single PR shouldn't have 100 commits and maybe we should expect contributors to rebase their branches if they are >100 behind?

austinmatherne-wk: Agreed. 100 is entirely reasonable.

Issue #2192, opened 2026-02-24
Update Command Line Operation Docs


austinmatherne-wk:
### What should we change and why?

The CLI [reference page](https://arelle.readthedocs.io/en/2.38.17/command_line.html) is [hand maintained](https://github.com/Arelle/Arelle/blob/master/docs/source/command_line.md) and has drifted from the actual `arelleCmdLine` `--help` output, some options are missing, others have stale or incorrect descriptions.

Minimum goal
Audit the current page against `--help` output and correct any missing or wrong entries.

Stretch goal
Autogenerate the CLI reference during the docs build by introspecting Arelle's parser config to keep docs permanently in sync without manual effort.

PR #2222, opened 2026-03-06
Enhance command-line documentation formatting and argument parsing

volodymyrtur-wk:
#### Reason for change
The GitHub [issue](https://github.com/Arelle/Arelle/issues/2192)

#### Description of change
This pull request updates the command line documentation to clarify argument descriptions, add new validation and output options, and improve formatting for readability.

#### Steps to Test
CI
**review**:
@Arelle/arelle

### DIFF
--- a/arelle/CntlrCmdLine.py
+++ b/arelle/CntlrCmdLine.py
@@ -1304,7 +1304,7 @@ def parseArgs(args: list[str]) -> tuple[RuntimeOptions, dict]:
         args = []
         namedOptions = set()
         optionsWithArg = set()
-        optionList = parser.option_list
+        optionList = parser.option_list.copy()
         [optionList.extend(group.option_list) for group in parser.option_groups]
         for option in optionList:
             names = str(option).split("/")


arelle/CntlrCmdLine.py
```diff
@@ -156,340 +156,1116 @@ def parseArgs(args):
...
+        action="store",
+        dest="saveOIMToXMLReport",
+        help=_("Save a report loaded from OIM into this file XML file name.")
+        )
+    validationGroup.add_option(
+        "--deduplicateFacts",
+        "--deduplicatefacts",
+        choices=[a.value for a in ValidateDuplicateFacts.DeduplicationType],
+        dest="deduplicateFacts",
+        default=ValidateDuplicateFacts.DeduplicationType.COMPLETE.value,
```
austinmatherne-wk: Lots of nice improvement in this PR!

One thing I'd like to adjust before I dig into a full review, right now formatting/structural changes and functional changes are all in the same commit, which makes it tough to review with confidence. As an example, I happened to noticed that `--deduplicateFacts` changed to have a default value in a7b6e6b36e26ededc49f86a002bf35bb4a203195. With 1,000+ lines of changes it's hard to be sure what else might have shifted in that commit.

Can you split these into separate commits, something like one for the pure reformatting, one for option reorganization, and one for any functional changes (new defaults, help text edits, etc.)? That way the team can quickly confirm the formatting is purely structural and then focus on reviewing the functional pieces.

We slip up from time to time, but this is a habit we try to follow for all PRs. Keeping formatting separate from behavioral changes and making each functional change it's own commit makes reviews way smoother. Happy to chat more about this if you have questions!

arelle/CntlrCmdLine.py
```diff
@@ -156,340 +156,1026 @@ def parseArgs(args):
...
+        help=_("Ignore detected file type and validate all files as Taxonomy Packages.")
+        )
+    parser.add_option(
+        "--deduplicateFacts",
+        "--deduplicatefacts",
+        choices=[a.value for a in ValidateDuplicateFacts.DeduplicationType],
+        dest="deduplicateFacts",
+        help=_(
+            "When using `--saveDeduplicatedInstance` to save a deduplicated instance, check for duplicates of this type"
+            " (`complete`, `consistent-pairs`, `consistent-sets`)."
```
aaroncameron-wk: If we're going to include these in the help message, we should have them be generated dynamically (see `choices` above).

Alternatively, maybe there's a way to configure the parser to include the `choices` values in the help message automatically?

arelle/CntlrCmdLine.py
```diff
@@ -156,340 +156,1026 @@ def parseArgs(args):
...
+            )
+        )
+    parser.add_option(
+        "--validateDuplicateFacts",
+        "--validateduplicatefacts",
+        choices=[a.value for a in ValidateDuplicateFacts.DUPLICATE_TYPE_ARG_MAP],
+        dest="validateDuplicateFacts",
+        help=_(
+            "Select which types of duplicates should trigger warnings "
+            "(`none`, `inconsistent`, `consistent`, `incomplete`, `complete`, `all`)"
```
aaroncameron-wk: See my comment on `--deduplicateFacts`

arelle/CntlrCmdLine.py
```diff
@@ -156,340 +156,1026 @@ def parseArgs(args):
...
+        )
+    parser.add_option(
+        "--logXmlMaxAttributeLength",
+        "--logxmlmaxattributelength",
+        action="store",
+        dest="logXmlMaxAttributeLength",
+        type="int",
+        help=_(
+            "Truncate XML log file attribute values at length. "
+            "The default is 4_096_000 for JSON content and 128 for everything else."
```
aaroncameron-wk: Nit: `4,096,000` would be more appropriate.

arelle/CntlrCmdLine.py
```diff
@@ -156,340 +156,1026 @@ def parseArgs(args):
...
+            )
+        )
+    parser.add_option(
+        "--testcaseResultOptions",
+        choices=("match-any", "match-all"),
+        action="store",
+        dest="testcaseResultOptions",
+        help=_(
+            "For testcase results, default is `match-any` expected result, "
+            "options to `match-any` or `match-all` expected result(s).  "
```
aaroncameron-wk: This was a bit clunky previously. Maybe: 

> For testcase results, default configuration is to pass if any expected result is matched (`match-any`). Alternatively, require that all expected results are matched (`match-all`).

arelle/CntlrCmdLine.py
```diff
@@ -156,340 +156,1026 @@ def parseArgs(args):
...
+        help=_("Specify formula are to be compiled but not executed.")
+        )
+    parser.add_option(
+        "--formulaCacheSize",
+        "--formulacachesize",
+        action="store",
+        dest="formulaCacheSize",
+        help=_(
+            "Specify the number of fact aspect combinations to cache during formula evaluations. "
+            "Negative numbers have no limit."
```
aaroncameron-wk: The note about the default value was removed: `Default: 10,000,000`

(although I don't mind not including it since it is an arbitrarily high limit)

arelle/CntlrCmdLine.py
```diff
@@ -156,340 +156,1026 @@ def parseArgs(args):
...
+        "--xdgConfigHome",
+        action="store",
+        dest="xdgConfigHome",
+        help=_(
+            "Specify non-standard location for configuration and cache files "
+            "(overrides environment parameter XDG_CONFIG_HOME)."
+            )
+        )
+    parser.add_option(
+        "--plugin",
```
aaroncameron-wk, line 1246: `--plugins` is a bit of a special argument, it get's "preparsed" and thus has some special handling (see `PREPARSE_ARG_CONFIGS`). I understand the motivation for adding a singular version but I think that's a separate discussion and PR.

arelle/CntlrCmdLine.py
```diff
@@ -222,13 +253,27 @@ def parseArgs(args: list[str]) -> tuple[RuntimeOptions, dict]:
...
+    inputGroup.add_option(
+        "--optionsFile",
+        "--optionsfile",
+        action="store",
+        dest="optionsFile",
+        help=_(
+            "Provide a path to a JSON file containing runtime options. "
+            "These options will be overridden by any command line options provided."
+            )
+        )
```
aaroncameron-wk, line 301: I think this belongs in a different category or without a category. This argument loads an options file as an alternative to providing other command line arguments, so it's not like the other "input file" arguments here that are actually passed into Arelle for processing.

arelle/CntlrCmdLine.py
```diff
@@ -177,7 +177,40 @@ def parseArgs(args: list[str]) -> tuple[RuntimeOptions, dict]:
...
+        dest="outputAttribution",
+        help=SUPPRESS_HELP
+        )
+    parser.add_option("--showOptions", action="store_true", dest="showOptions", help=SUPPRESS_HELP)
+    parser.add_option("--gfm", action="store", dest="disclosureSystemName", help=SUPPRESS_HELP)
+    # special option for web interfaces to suppress closing an opened modelXbrl
+    parser.add_option("--keepOpen", dest="keepOpen", action="store_true", help=SUPPRESS_HELP)
+
+    # ====== GENERAL ARGUMENTS ======
+    generalGroup = OptionGroup(parser, _("General Arguments"))
```
aaroncameron-wk, line 195: Based on contents, maybe "Informational Arguments"?

arelle/CntlrCmdLine.py
```diff
@@ -191,19 +224,17 @@ def parseArgs(args: list[str]) -> tuple[RuntimeOptions, dict]:
...
         "--username",
         dest="username",
         help=_("user name if needed (with password) for web file retrieval")
         )
-    parser.add_option(
+    inputGroup.add_option(
         "--password",
         dest="password",
         help=_("password if needed (with user name) for web retrieval")
         )
```
aaroncameron-wk: These seem more appropriate to be in the same group as things like `--proxy`, `--httpUserAgent`, `--internetConnectivity` etc. They don't configure the input file themselves but may influence how those files are accessed.

arelle/CntlrCmdLine.py
```diff
@@ -596,45 +653,50 @@ def parseArgs(args: list[str]) -> tuple[RuntimeOptions, dict]:
...
+    outputFileGroup.add_option(
+        "--renderedTableLinkbase",
         action="store",
-        dest="skipLoading",
+        dest="renderedTableLinkbaseFile",
+        help=_("Write rendered table linkbase grid into `FILE` (format determined by extension: `.html`, `.xlsx`)")
+        )
+    outputFileGroup.add_option(
+        "--validationExitCode",
+        action="store_true",
```
aaroncameron-wk: In my opinion, this is better located in the validation group than the output file group.

arelle/CntlrCmdLine.py
```diff
@@ -414,39 +392,114 @@ def parseArgs(args: list[str]) -> tuple[RuntimeOptions, dict]:
...
+        )
+    validationGroup.add_option(
+        "--taxonomyPackage",
+        "--taxonomypackage",
+        action="store_true",
+        dest="taxonomyPackage",
+        help=_("Ignore detected file type and validate all files as Taxonomy Packages.")
+        )
+    validationGroup.add_option(
+        "--saveOIMToXMLReport",
```
aaroncameron-wk: Output file group?

arelle/CntlrCmdLine.py
```diff
@@ -414,39 +392,114 @@ def parseArgs(args: list[str]) -> tuple[RuntimeOptions, dict]:
...
+            "When using `--saveDeduplicatedInstance` to save a deduplicated instance, check for duplicates of this type"
+            " (`complete`, `consistent-pairs`, `consistent-sets`)."
+            )
+        )
+    validationGroup.add_option(
+        "--saveDeduplicatedInstance",
+        "--savededuplicatedinstance",
+        dest="saveDeduplicatedInstance",
+        help=_("Save an instance document with duplicates of the provided type (`--deduplicateFacts`) deduplicated.")
+        )
```
aaroncameron-wk: These work together to generate an output file, so I think both belong in that group.

arelle/CntlrCmdLine.py
```diff
@@ -414,39 +392,114 @@ def parseArgs(args: list[str]) -> tuple[RuntimeOptions, dict]:
...
-    parser.add_option(
+    validationGroup.add_option(
         "--utr",
         action="store_true",
         dest="utrValidate",
         help=_("Select validation with respect to Unit Type Registry.")
         )
-    parser.add_option(
+    validationGroup.add_option(
         "--utrUrl",
```
aaroncameron-wk, line 439: I don't think this is specific enough to validation to merit being in this group. I believe it can influence how instances are loaded even without validation. I could be wrong though, I'll let others weigh in.

arelle/CntlrCmdLine.py
```diff
@@ -723,30 +785,31 @@ def parseArgs(args: list[str]) -> tuple[RuntimeOptions, dict]:
...
+
+    formulaGroup.add_option(
         "--parameters",
         action="store",
         dest="parameters",
         help=_("Specify parameters for formula and validation (name=value[,name=value]).")
         )
-    parser.add_option(
+    formulaGroup.add_option(
         "--parameterSeparator",
```
aaroncameron-wk: If we move these to the validation group, then this group can become simply "Formula Arguments".

arelle/CntlrCmdLine.py
```diff
@@ -964,31 +1055,18 @@ def parseArgs(args: list[str]) -> tuple[RuntimeOptions, dict]:
...
-            "Negative numbers have no limit."
+    parser.add_option_group(testGroup)
+
+    # ====== CONFIGURATION OPTIONS ======
+    configGroup = OptionGroup(
+        parser,
+        _("Configuration Options"),
+        _(
+            "It is possible to specify options to be defaults for the web server, "
+            "such as disclosureSystem and validations, but not including file names."
```
aaroncameron-wk, line 1099: I don't think this message is an appropriate description for this group. These options aren't specific to the web server.

arelle/CntlrCmdLine.py
```diff
@@ -1010,83 +1088,82 @@ def parseArgs(args: list[str]) -> tuple[RuntimeOptions, dict]:
...
         dest="internetRecheck",
         help=_(
-            "Specify rechecking for newer cache files 'quarter-hourly', "
-            "`hourly`, `daily`, `weekly` or `never`."
+            "Specify rechecking for newer cache files 'quarter-hourly', `hourly`, `daily`, `weekly` or `never`."
             )
         )
-    parser.add_option(
+    configGroup.add_option(
         "--internetLogDownloads",
```
aaroncameron-wk: To "Logging Arguments"

arelle/CntlrCmdLine.py
```diff
@@ -1010,83 +1088,82 @@ def parseArgs(args: list[str]) -> tuple[RuntimeOptions, dict]:
...
+    configGroup.add_option(
         "--cacheDirectory",
         "--cachedirectory",
         action="store",
         dest="cacheDirectory",
         help=_("Override the default location of the cache directory")
         )
-    parser.add_option(
+    configGroup.add_option(
+        "--xdgConfigHome",
```
aaroncameron-wk: This should go in the same place that `--optionsFile` ends up.

arelle/CntlrCmdLine.py
```diff
@@ -1010,83 +1088,82 @@ def parseArgs(args: list[str]) -> tuple[RuntimeOptions, dict]:
...
+    configGroup.add_option(
         "--httpUserAgent",
         "--httpuseragent",
         action="store",
         dest="httpUserAgent",
         help=_("Specify non-standard http header User-Agent value")
         )
-    parser.add_option(
+    configGroup.add_option(
         DISABLE_PERSISTENT_CONFIG_OPTION,
```
aaroncameron-wk: This should go in the same place that `--optionsFile` ends up.

arelle/CntlrCmdLine.py
```diff
@@ -1115,7 +1192,7 @@ def parseArgs(args: list[str]) -> tuple[RuntimeOptions, dict]:
             "Cmd line package settings are not persistent.)"
             )
         )
-    parser.add_option(
+    configGroup.add_option(
         "--packageManifestName",
```
aaroncameron-wk: This basically specifies where to look for a manifest file within a taxonomy package, so I think it belongs in `Input File Arguments` with `--packages`.

arelle/CntlrCmdLine.py
```diff
@@ -1103,9 +1180,9 @@ def parseArgs(args: list[str]) -> tuple[RuntimeOptions, dict]:
             "Packaged plug-in urls are their directory's url (e.g., `--plugins EDGAR/render` or `--plugins xbrlDB`).  "
             )
         )
-    parser.add_option(
-        "--packages",
+    configGroup.add_option(
         "--package",
+        "--packages",
```
aaroncameron-wk: Packages are another form of input file, in my opinion, so `Input File Arguments` seems appropriate.

arelle/CntlrCmdLine.py
```diff
@@ -1126,27 +1203,39 @@ def parseArgs(args: list[str]) -> tuple[RuntimeOptions, dict]:
             "(Replaces search for either .taxonomyPackage.xml or catalog.xml)."
             )
         )
-    parser.add_option(
+    configGroup.add_option(
         "--abortOnMajorError",
```
aaroncameron-wk: This is a kind of "application lifecycle" control argument, similar to `--keepOpen`. I don't know if they necessitate their own group, but they seem like they should be together.

arelle/CntlrCmdLine.py
```diff
@@ -1126,27 +1203,39 @@ def parseArgs(args: list[str]) -> tuple[RuntimeOptions, dict]:
...
         action="store_true",
         dest="abortOnMajorError",
         help=_("Abort process on major error, such as when load is unable to find an entry or discovered file.")
         )
-    parser.add_option(
+    configGroup.add_option(
+        "--skipDTS",
```
aaroncameron-wk: I could see `--skipDTS` being a part of the "application lifecycle" group (see above comment).

arelle/CntlrCmdLine.py
```diff
@@ -1126,27 +1203,39 @@ def parseArgs(args: list[str]) -> tuple[RuntimeOptions, dict]:
...
+        "--skipdts",
+        action="store_true",
+        dest="skipDTS",
+        help=_("Skip DTS activities (loading, discovery, validation), useful when an instance needs only to be parsed.")
+        )
+    configGroup.add_option(
+        "--skipLoading",
```
aaroncameron-wk: This is maybe more of an "Input File" argument since it filters out files from the input/loading process.

arelle/CntlrCmdLine.py
```diff
@@ -1126,27 +1203,39 @@ def parseArgs(args: list[str]) -> tuple[RuntimeOptions, dict]:
...
+        "--skiploading",
+        action="store",
+        dest="skipLoading",
+        help=_(
+            "Skip loading discovered or schemaLocated files matching pattern "
+            "(unix-style file name patterns separated by `|`), useful when not all linkbases are needed."
+            )
+        )
+    configGroup.add_option(
         "--showEnvironment",
```
aaroncameron-wk: This belongs with `--about` and `--diagnostics`

arelle/CntlrCmdLine.py
```diff
@@ -1126,27 +1203,39 @@ def parseArgs(args: list[str]) -> tuple[RuntimeOptions, dict]:
...
-        "--collectProfileStats",
-        action="store_true",
-        dest="collectProfileStats",
-        help=_("Collect profile statistics, such as timing of validation activities and formulae.")
-        )
+
     if hasWebServer():
-        parser.add_option(
+        configGroup.add_option(
             "--webserver",
```
aaroncameron-wk: Maybe "application lifecycle" group mentioned above?

arelle/CntlrCmdLine.py
```diff
@@ -177,7 +177,40 @@ def parseArgs(args: list[str]) -> tuple[RuntimeOptions, dict]:
...
+    parser.add_option("--monitorParentProcess", action="store", dest="monitorParentProcess", help=SUPPRESS_HELP)
     parser.add_option(
+        "--outputAttribution",
+        "--outputattribution",
+        action="store",
+        dest="outputAttribution",
+        help=SUPPRESS_HELP
+        )
+    parser.add_option("--showOptions", action="store_true", dest="showOptions", help=SUPPRESS_HELP)
+    parser.add_option("--gfm", action="store", dest="disclosureSystemName", help=SUPPRESS_HELP)
```
aaroncameron-wk, line 190: Even though this is SUPPRESS_HELP, this is sort of an alias for `--disclosureSystem`, so can we move it near that? I think we should try and group these even if they are suppressed from the `--help` output.

arelle/CntlrCmdLine.py
```diff
@@ -964,31 +1055,18 @@ def parseArgs(args: list[str]) -> tuple[RuntimeOptions, dict]:
...
-        dest="formulaCacheSize",
-        help=_(
-            "Specify the number of fact aspect combinations to cache during formula evaluations. "
-            "Negative numbers have no limit."
+    parser.add_option_group(testGroup)
+
+    # ====== CONFIGURATION OPTIONS ======
+    configGroup = OptionGroup(
+        parser,
+        _("Configuration Options"),
```
aaroncameron-wk, line 1096: This group is currently sort of a catch all for a bunch of different args. I suggest:
```
"Internet Connectivity Arguments"
    --proxy
    --internetConnectivity
    --internetTimeout
    --internetRecheck
    --noCertificateCheck
    --httpsRedirectCache
    --cacheDirectory
    --httpUserAgent
```
```
"Configuration Arguments"
    --optionsFile
    --xdgConfigHome
    --disablePersistentConfig
    --plugins
```
I'll try and make suggestions for the rest below.

arelle/CntlrCmdLine.py
volodymyrtur-wk: I removed default values from help text in some options, because the help text defines a default value, but in the option builder, we don't define this value.

arelle/CntlrCmdLine.py
```diff
@@ -531,15 +1291,15 @@ def parseArgs(args):
         namedOptions = set()
         optionsWithArg = set()
         for option in parser.option_list:
```
austinmatherne-wk: I'm not sure if this code is actually still required, but it needs to be updated to handle both the base options list and group options (or tested on windows to see if it can be removed).

arelle/CntlrCmdLine.py
```diff
@@ -156,340 +156,1117 @@ def parseArgs(args):
...
-    parser.add_option("--saveDeduplicatedInstance", "--savededuplicatedinstance",
-                      dest="saveDeduplicatedInstance",
-                      help=_("Save an instance document with duplicates of the provided type ('--deduplicateFacts') deduplicated."))
-    parser.add_option("--noValidateTestcaseSchema", "--novalidatetestcaseschema", action="store_false", dest="validateTestcaseSchema", default=True,
-                      help=_("Validate testcases against their schemas."))
-    betaGroup = OptionGroup(parser, "Beta Features",
-                        "Caution: these are beta features, use these options at your own risk.")
+
+    # ====== GENERAL ARGUMENTS ======
+    generalGroup = OptionGroup(parser, _("General Arguments"))
```
austinmatherne-wk, line 195: Using groups is a good refactor, but does change the state of the parser that's passed to plugins (what's exposed via `parser.option_list` for example). Can you check if any of the 3 major external plugins ([EDGAR](https://github.com/Arelle/EDGAR), [XULE](https://github.com/xbrlus/xule), and [ixbrl-viewer](https://github.com/Arelle/ixbrl-viewer)) will be impacted by the use of groups?

arelle/CntlrCmdLine.py
austinmatherne-wk: It's a bit sloppy, but while the option configuration may not define a default value, the application [applies a default elsewhere](https://github.com/Arelle/Arelle/blob/3a34b5d5c61383b3c5a6788b2d5e042eabf56acc/arelle/ModelFormulaObject.py#L26) if the option was unset. I think that's what the prior message was trying to capture.

volodymyrtur-wk: Yes, you are right. In this case, we can set a default value in the option builder to avoid misunderstanding, and it doesn't break anything.  But in the case of `--deduplicateFacts`, the original help text said `When using '--saveDeduplicatedInstance' to save a deduplicated instance, check for duplicates of this type. Defaults to 'complete'`. `Defaults to 'complete'` is true only if we use the `--saveDeduplicatedInstance`. Using `--deduplicateFacts` without `--saveDeduplicatedInstance` causes the error, and there is no information about it in the help text, which can lead to misunderstanding and confusion among users. I think it would be a good idea to slightly change the help text to something like: `Use only with '--saveDeduplicatedInstance' to save a deduplicated instance, check for duplicates of this type. Defaults to 'complete'.`

arelle/CntlrCmdLine.py
```diff
@@ -530,16 +1304,18 @@ def parseArgs(args):
         args = []
         namedOptions = set()
         optionsWithArg = set()
-        for option in parser.option_list:
-            names = str(option).split('/')
+        optionList = parser.option_list
+        [optionList.extend(group.option_list) for group in parser.option_groups]
```
austinmatherne-wk, line 1308: This is mutating `parser.option_list` by extending it with all group options which impacts any other code that accesses this later. You should use `optionList = list(parser.option_list)` to create a copy then extend the copy, or build a separate list.

volodymyrtur-wk: :facepalm: Yes, you're right)

PR #2224, opened 2026-03-06
Centralizing common fact checks

allenbross-wf:
#### Description of change
Centralized some fact checks
Added empty set returns for factsByQname.gets.
camelCased some snake_cased variables.

#### Steps to Test

CI

**review**:
@Arelle/arelle

### DIFF
--- a/arelle/utils/validate/Facts.py
+++ b/arelle/utils/validate/Facts.py
@@ -58,7 +58,7 @@ def factHasNegativeNumericValue(fact: ModelFact) -> bool:
     """
     Returns True if fact is valid and has a negative numeric value.
     """
-    return fact.xValid >= VALID and fact.xValue is not None and fact.isNumeric and cast(int, fact.xValue) < 0
+    return fact.xValid >= VALID and fact.xValue is not None and isinstance(fact.xValue, int) and fact.xValue < 0
 
 
 def getDuplicateFactGroupsByConceptContextUnit(


arelle/plugin/validate/DBA/rules/fr.py
```diff
@@ -1372,7 +1373,7 @@ def rule_fr89(
                 pluginData.independentAuditorsReportDanish,
                 pluginData.independentAuditorsReportEnglish,
             ]
-        auditorFacts = val.modelXbrl.factsByQname.get(pluginData.typeOfAuditorAssistanceQn, set())
+        auditorFacts = val.modelXbrl.factsByQname.get (pluginData.typeOfAuditorAssistanceQn, set())
```
austinmatherne-wk: typo:
```suggestion
        auditorFacts = val.modelXbrl.factsByQname.get(pluginData.typeOfAuditorAssistanceQn, set())
```

arelle/utils/validate/Facts.py
```diff
@@ -0,0 +1,137 @@
...
+    Returns True if fact is valid (xValid >= VALID) and non-nil.
+    """
+    return fact.xValid >= VALID and not fact.isNil
+
+
+def factHasNegativeNumericValue(fact: ModelFact) -> bool:
+    """
+    Returns True if fact is valid and has a negative numeric value.
+    """
+    return fact.xValid >= VALID and fact.xValue is not None and cast(int, fact.xValue) < 0
```
austinmatherne-wk: This should check that the value is numeric instead of casting. If it's not, the comparison operator will throw a type error.

arelle/utils/validate/Facts.py
```diff
@@ -0,0 +1,137 @@
...
+
+
+def getUsedConceptsFromFacts(modelXbrl: ModelXbrl) -> set[ModelConcept]:
+    """
+    Returns the set of concepts used on facts in the instance.
+    """
+    return {fact.concept for fact in modelXbrl.facts if fact.concept is not None}
+
+
+def getRequiredFactMissing(
```
austinmatherne-wk: This is fine, but get is out of place as a function name prefix here, it's returning true/false, not returning an object. Also, non-negative functions are generally preferred. For example:
```python
if hasNonNillFact(...):
if not hasNonNillFact(...):
```
Is easier to understand at a quick glance than:
```python
if not getRequiredFactMissing(...):
if getRequiredFactMissing(...):
```

arelle/utils/validate/Facts.py
```diff
@@ -0,0 +1,137 @@
...
+    byQname: bool = True,
+) -> bool:
+    """
+    Returns True if no fact with the given concept key is tagged with a non-nil value.
+    conceptKey is QName when byQname=True, else str (localName).
+    """
+    if byQname:
+        facts = modelXbrl.factsByQname.get(cast(QName, conceptKey), set())
+    else:
+        facts = modelXbrl.factsByLocalName.get(cast(str, conceptKey), set())
```
austinmatherne-wk: Instead of using a `byQname` param (that the caller is required to keep in sync with the `conceptKey` type), use an `isinstance` check on `conceptKey` itself:
```suggestion
    if isinstance(conceptKey, QName):
        facts = modelXbrl.factsByQname.get(conceptKey, set())
    else:
        facts = modelXbrl.factsByLocalName.get(conceptKey, set())
```

arelle/utils/validate/Facts.py
```diff
@@ -0,0 +1,137 @@
...
+    for fact in facts:
+        if not fact.isNil:
```
austinmatherne-wk: Should this also be checking `fact.xValid >= VALID`?

arelle/utils/validate/Facts.py
```diff
@@ -0,0 +1,137 @@
...
+            return False
+    return True
+
+
+def getNegativeFacts(
+    modelXbrl: ModelXbrl,
+    conceptKey: Union[QName, str],
+    byQname: bool = True,
```
austinmatherne-wk: Similar `byQname` comment as above.

arelle/utils/validate/Facts.py
```diff
@@ -0,0 +1,137 @@
...
+    Returns True if fact is valid (xValid >= VALID) and non-nil.
+    """
+    return fact.xValid >= VALID and not fact.isNil
+
+
+def factHasNegativeNumericValue(fact: ModelFact) -> bool:
+    """
+    Returns True if fact is valid and has a negative numeric value.
+    """
+    return fact.xValid >= VALID and fact.xValue is not None and fact.isNumeric and cast(int, fact.xValue) < 0
```
austinmatherne-wk: Following up on my prior comment, this should still check the type of `fact.xValue` to avoid type errors at runtime (the cast is masking a real type issue that mypy is trying to warn us about).

PR #2240, opened 2026-03-16
Consolidating context checks

allenbross-wf:
#### Description of change

Consolidating context checks

#### Steps to Test

CI

**review**:
@Arelle/arelle


arelle/utils/validate/Contexts.py
```diff
@@ -0,0 +1,62 @@
...
+    result: list[ModelContext | None] = []
+    for context in modelXbrl.contexts.values():
+        if context.hasSegment:
+            result.append(context)
+    return result
+
+
+def getContextsWithScenarioContent(modelXbrl: ModelXbrl) -> list[ModelContext | None]:
+    """Return contexts that contain scenario content."""
+    result: list[ModelContext | None] = []
```
aaroncameron-wk: While I think values in `modelXbrl.contexts` can be `None`, I think we should simply skip those values and have the return type be `list[ModelContext]`. Applies to other functions as well.

arelle/plugin/validate/EBA/__init__.py
```diff
@@ -200,11 +203,11 @@ def validateFacts(val, factsToCheck):
             # cannot pass context object to final() below, for error logging, if streaming mode
             val.cntxDates[cntx.instantDatetime].add(modelXbrl if getattr(val.modelXbrl, "isStreamingMode", False)
                                                     else cntx)
-        if cntx.hasSegment:
+        if cntx in contextsWithSegments:
```
aaroncameron-wk: Since `contextsWithSegments` is a list, this is `O(n)`. We should either have these utility functions return a set, or just leave this use as it was previously since we are already iterating over all contexts anyway.

arelle/utils/validate/Contexts.py
```diff
@@ -0,0 +1,62 @@
...
+
+def getContextsWithPeriodTime(modelXbrl: ModelXbrl) -> list[ModelContext | None]:
+    """Return contexts whose period elements (startDate/endDate/instant) include time content."""
+    result: list[ModelContext | None] = []
+    datetimePattern = lexicalPatterns["XBRLI_DATEUNION"]
+    for context in modelXbrl.contexts.values():
+        for uncastElt in context.iterdescendants(
+            "{http://www.xbrl.org/2003/instance}startDate",
+            "{http://www.xbrl.org/2003/instance}endDate",
+            "{http://www.xbrl.org/2003/instance}instant",
```
aaroncameron-wk: Pre-existing, but these seem like good candidates for `XbrlConst`.

arelle/plugin/validate/ESEF/ESEF_2021/ValidateXbrlFinally.py
```diff
@@ -499,45 +505,23 @@ def checkFootnote(elt: ModelInlineFootnote | ModelResource, text: str) -> None:
...
             return # no more checks apply
 
-        contextsWithDisallowedOCEs = []
-        contextsWithDisallowedOCEcontent = []
-        contextsWithPeriodTime: list[ModelContext] = []
-        contextsWithPeriodTimeZone: list[ModelContext] = []
+        contextsWithPeriodTime = getContextsWithPeriodTime(modelXbrl)
+        contextsWithPeriodTimeZone = getContextsWithPeriodTimeZone(modelXbrl)
+        contextsWithDisallowedOCEs = getContextsWithSegments(modelXbrl)
+        contextsWithDisallowedOCEcontent = getContextsWithScenarioContent(modelXbrl)
```
aaroncameron-wk: This is effectively adding 4 iterations over all contexts. While it does reduce code duplication, I'm not sure this is an improvement for this implementation.
Perhaps this validation and the new utility functions could make use of shared boolean functions (`contextHasPeriodTime`, `contextHasSegment`, etc.)

arelle/utils/validate/ContextData.py
```diff
@@ -0,0 +1,66 @@
...
+    violate common XBRL/ESEF context rules (period format, segment/scenario usage).
+    """
+    contextsWithImproperContent: set[ModelContext]
+    contextsWithPeriodTime: set[ModelContext]
+    contextsWithPeriodTimeZone: set[ModelContext]
+    contextsWithSegments: set[ModelContext]
+    contextsWithWrongInstantDate: set[ModelContext]
+
+
+def checkContexts(modelXbrl: ModelXbrl,esefYear: int | None = None,) -> ContextData:
```
austinmatherne-wk: A function name of `getContextIssues` and a class name of `ContextIssues` might be better indicators of what these are/do.

PR #2241, opened 2026-03-16
Add type hints for `PluginManager`

aaroncameron-wk:
#### Steps to Test
CI

**review**:
@Arelle/arelle

### DIFF
--- a/arelle/PluginManager.py
+++ b/arelle/PluginManager.py
@@ -42,13 +42,7 @@ PLUGIN_TRACE_LEVEL = logging.WARNING
 
 # plugin control is static to correspond to statically loaded modules
 pluginJsonFile = None
-# The code in this module mostly assumes these values are not `NoneType` and
-# relies on try/except catching AttributeError, TypeError, etc. if they ever are.
-# Rather than add type casting or assertions throughout the code, we will ignore
-# the assignment type hint error here and rely on the initialization process to set
-# these values before use.
-pluginConfig: dict[str, Any] = None # type: ignore[assignment]
-
+pluginConfig: dict[str, Any] | None = None
 pluginConfigChanged = False
 pluginTraceFileLogger = None
 modulePluginInfos: dict[str, Any] = {}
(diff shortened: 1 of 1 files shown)


arelle/PluginManager.py
```diff
@@ -36,22 +36,32 @@
...
 # plugin control is static to correspond to statically loaded modules
 pluginJsonFile = None
-pluginConfig: dict | None = None
+# The code in this module mostly assumes these values are not `NoneType` and
+# relies on try/except catching AttributeError, TypeError, etc. if they ever are.
+# Rather than add type casting or assertions throughout the code, we will ignore
+# the assignment type hint error here and rely on the initialization process to set
+# these values before use.
+pluginConfig: dict[str, Any] = None # type: ignore[assignment]
+_cntlr: Cntlr = None # type: ignore[assignment]
```
aaroncameron-wk: I don't feel great about this, but the alternative is adding a lot of casts or non-None assertions that can potentially change behavior considering how much this code relies on try/except catching errors.

arelle/PluginManager.py
```diff
@@ -36,22 +36,32 @@
...
+modulePluginInfos: dict[str, Any] = {}
+pluginMethodsForClasses: dict[str, Any] = {}
 _pluginBase = None
-EMPTYLIST = []
+EMPTYLIST: list[Any] = []
 _ERROR_MESSAGE_IMPORT_TEMPLATE = "Unable to load module {}"
 
+TypeModuleInfoKey = Union[
+    str, bytes, int, float, complex, types.EllipsisType, None
+]
```
austinmatherne-wk: Out of curiosity, in practice this is always a string, right?

austinmatherne-wk (approved):

arelle/PluginManager.py
austinmatherne-wk: how about:
```python
 _cntlr: Cntlr | None = None

  def _getCntlr() -> Cntlr:
      assert _cntlr is not None, "PluginManager.init() must be called before use"
      return _cntlr
```

The only place that looks like `None` is possibly permitted is in `parsePluginInfo` with `openFileStream`, so that could still use `_cntlr` directly

arelle/PluginManager.py
aaroncameron-wk: Yes. And it does seem preferable to be able to assume `str` throughout so I've added an assertion.

PR #2243, opened 2026-03-17
Table Constraints Metadata Parser

austinmatherne-wk:
#### Reason for change
Implements an initial xBRL-CSV Table Constraints metadata parser.

#### Description of change
Adds a metadata parser for the Table Constraints extension to xBRL-CSV. This PR covers metadata parsing and most of the structural validation layer. Additional validations and report level constraint enforcement will follow separately. The Table Constraints sub-package is prefixed with an underscore while it is still actively being built so users know not to write code against it and we're free to refactor as needed.

#### Steps to Test
* CI (note: I uploaded an updated draft conformance suite to S3. This PR is configured to run against the updated test suite IDs, but master is still configured against the old suite and will fail until this is merged).

**review**:
@Arelle/arelle

### DIFF
--- a/arelle/oim/_tc/metadata/parser.py
+++ b/arelle/oim/_tc/metadata/parser.py
@@ -26,6 +26,9 @@ from arelle.oim._tc.metadata.model import (
     TCUniqueKey,
     TCValueConstraint,
 )
+from arelle.typing import TypeGetText
+
+_: TypeGetText
 
 
 class TCMetadataParseError(TCError):
@@ -48,7 +51,11 @@ class TCMetadataParseError(TCError):
 class TCMetadataParseTypeError(TCMetadataParseError):
     def __init__(self, expected_type: type, actual_value: Any, *path: str) -> None:
         super().__init__(
-            f"Expected {expected_type.__name__}, got {type(actual_value).__name__}: {actual_value!r}",
+            _("Expected {expected}, got {actual}: {value}").format(
+                expected=expected_type.__name__,
+                actual=type(actual_value).__name__,
+                value=repr(actual_value),
+            ),
             *path,
         )
 
@@ -56,7 +63,7 @@ class TCMetadataParseTypeError(TCMetadataParseError):
 class TCMetadataUnknownPropertiesError(TCMetadataParseError):
     def __init__(self, property_names: list[str]) -> None:
         names = ", ".join(repr(n) for n in sorted(property_names))
-        super().__init__(f"Unknown properties: {names}")
+        super().__init__(_("Unknown properties: {names}").format(names=names))
 
 
 def _prepend_paths(errors: list[TCMetadataParseError], *segments: str) -> list[TCMetadataParseError]:
@@ -476,7 +483,9 @@ def _parse_bounded_int(
 ) -> int | None:
     val = _parse_primitive_field(obj, key, int, errors, default=default)
     if val is not None and val < min_value:
-        errors.append(TCMetadataParseError(f"Value {val} is less than minimum {min_value}", key))
+        errors.append(
+            TCMetadataParseError(_("Value {value} is less than minimum {min}").format(value=val, min=min_value), key)
+        )
         return default
     return val
 


arelle/oim/_tc/const.py
```diff
@@ -0,0 +1,75 @@
...
+    "https://xbrl.org/WGWD/YYYY-MM-DD/tc/metadataerror",
+    "https://xbrl.org/CR/2025-10-07/tc/metadataerror",
+})
+
+TC_REPORT_ERROR_NAMESPACES = frozenset({
+    "https://xbrl.org/WGWD/YYYY-MM-DD/tc/reporterror",
+    "https://xbrl.org/CR/2025-10-07/tc/reporterror",
+})
+
+TC_PREFIX = "tc"
```
brettkail-wk, line 25: Out of curiosity, is this prefix required or conventional?  I was under the impression that OIM supported arbitrary prefixes so long as they were mapped, but the code seems to look for property names with specific prefixes.

arelle/oim/Load.py
```diff
@@ -1229,6 +1230,15 @@ def checkMemberTypes(obj, path, pathParts):
             documentInfoProperties.add("final")
             reportProperties = {"documentInfo", "tableTemplates", "tables", "parameters", "parameterURL", "dimensions", "decimals", "links"}
             columnProperties = {"comment", "decimals", "dimensions", "propertyGroups", "parameterURL", "propertiesFrom"}
+            try:
+                modelXbrl.tcMetadata = parse_tc_metadata(oimObject, namespaces)
+            except TCMetadataParseError as err:
```
brettkail-wk: Do you think it would be useful to be able to keep going and report all errors at once?  That will be more work for the parser, but it seems like it would be a better user experience rather than getting one error at a time, particularly since it can lead to ambiguity: did my change remove an error and let me get to the next one, or did my change introduce a new error that I need to fix first?

arelle/oim/_tc/const.py
austinmatherne-wk: In this case it's required:
> `The namespace URI https://xbrl.org/.../tc MUST NOT be bound to a prefix other than tc`

brettkail-wk: Interesting, thanks.  Do you know happen to know the rationale for that?

austinmatherne-wk: It's aligned with the use of reserved prefixes in other OIM specs, but as far as the rational of [reserved prefixes/URI alias in OIM](https://www.xbrl.org/Specification/oim-common/REC-2021-10-13/oim-common-REC-2021-10-13.html#term-reserved-prefix) in general, I don't know. I wasn't involved in the working group back then.

arelle/oim/Load.py
austinmatherne-wk: Good call. It's more complicated, but I reworked the PR to gather and report all errors.

arelle/oim/_tc/metadata/parser.py
```diff
@@ -0,0 +1,605 @@
...
+    local_errors: list[TCMetadataParseError] = []
+
+    col = _parse_column_constraints(template_obj.get("columns", {}), local_errors)
+    params = _parse_param_constraints(template_obj, local_errors)
+    keys = _parse_template_keys(template_obj, local_errors)
+    column_order = _parse_ordered_set(template_obj, TC_COLUMN_ORDER_PROPERTY_NAME, local_errors)
+    table_constraints = _parse_template_table_constraints(template_obj, local_errors)
+
+    for error in local_errors:
+        error.prepend_path("tableTemplates", template_id)
```
brettkail-wk: This is completely fine.  I guess this could be pushed into the `if local_errors`?  I wonder if a `_prepend_paths(errors, *segments)` would compress the `.extend(_prepend_paths(...))`.

brettkail-wk: This is fine, but it's somewhat surprising to me that `.get(key)` and `prepend_path(key,...)` are in different functions.  I would have expected those to be colocated (in `parse_tc_metadata` in this case).

austinmatherne-wk: Updated.

austinmatherne-wk: Updated.

arelle/oim/_tc/metadata/parser.py
```diff
@@ -0,0 +1,602 @@
...
+@overload
+def _parse_bounded_int(
+    obj: dict[str, Any], key: str, errors: list[TCMetadataParseError], *, min_value: int, default: None = ...
+) -> int | None: ...
+def _parse_bounded_int(
+    obj: dict[str, Any], key: str, errors: list[TCMetadataParseError], *, min_value: int, default: int | None = None
+) -> int | None:
+    val = _parse_primitive_field(obj, key, int, errors, default=default)
+    if val is not None and val < min_value:
+        errors.append(TCMetadataParseError(f"Value {val} is less than minimum {min_value}", key))
```
aaroncameron-wk: I'm fine with doing this later, but should we lay the groundwork for translations of these messages with gettext (`_(...)`)?

austinmatherne-wk: Good catch. Updated.

PR #2244, opened 2026-03-17
Fix false-positive DORA formula filtering across handled dimensions

ignatremizov:
#### Reason for change

Arelle was producing false-positive DORA validation failures on xBRL-CSV reports for `B_05.01` assertions such as `v8851_m_0` through `v8855_m_0`, even when the row data satisfied the rules.

The reproduced pattern was that facts expected for a single `B_05.01` row were present in the instance, but formula evaluation was allowing facts from a different dimensional signature to survive filtering and interfere with the assertion bindings.

#### Description of change

This PR makes one formula-evaluator change and adds focused unit coverage around related fallback behavior.

In `trialFilterFacts()`:
- require all non-complement handled dimensions to be present for the optimized handled-dimension path
- require all complement handled dimensions to be absent for the optimized complement path

This changes the optimized path from effectively matching any handled dimension to matching the full handled-dimension signature intended by the group filters.

The PR also adds focused unit tests covering:
- fallback after a non-sequence fact variable already yields a fact evaluation
- fallback use when sequence binding has candidate facts but yields no matching evaluation
- fallback preservation when sequence binding yields a matching sequence evaluation
- handled-dimension optimization requiring all non-complement dimensions
- handled-dimension complement optimization requiring all complemented dimensions to be absent

These fallback-related tests are included as regression guards for existing Formula 1.0 bind-empty behavior. After the CI-guided removals of the earlier `implicitFilter()` and `evaluationResults()` experiments, the `trialFilterFacts()` handled-dimension change is the only net-new formula-evaluator behavior change.

#### Steps to Test

1. Run the focused unit tests:
   ```bash
   .venv/bin/python -m pytest tests/unit_tests/arelle/formula/test_formula_evaluator.py -q
   ```
2. Validate a DORA xBRL-CSV report that previously showed false-positive `B_05.01` assertion failures:
   ```bash
   .venv/bin/python arelleCmdLine.py \
     -f report.json \
     -v \
     --internetConnectivity offline \
     --logFile /tmp/arelle-validation.log \
     --logFileMode w \
     --logFormat '[%(messageCode)s] %(message)s'
   ```
3. Confirm that the report no longer produces false-positive `v8851_m_0` through `v8855_m_0` messages.
4. As a regression control, run the Formula 1.0 conformance shard covering `22180 Bind To Empty Sequence` and confirm those cases still pass.

**review**:
@Arelle/arelle

### DIFF
--- a/arelle/formula/FormulaEvaluator.py
+++ b/arelle/formula/FormulaEvaluator.py
@@ -748,11 +748,11 @@ def trialFilterFacts(xpCtx, vb, facts, filterRelationships, filterType, var=None
             # filter now with filter info
             outFacts = set()
             for fact in facts:
-                if fact.isItem:
-                    for varFilterRel, dimQname in noComplHandledFilterRels:
-                        dim = fact.context.qnameDims.get(dimQname)
-                        if dim is not None:
-                            outFacts.add(fact)
+                if (
+                    fact.isItem
+                    and all(fact.context.qnameDims.get(dimQname) is not None for _, dimQname in noComplHandledFilterRels)
+                ):
+                    outFacts.add(fact)
             facts = outFacts
             if len(facts) == 0:
                 return facts
@@ -799,11 +799,11 @@ def trialFilterFacts(xpCtx, vb, facts, filterRelationships, filterType, var=None
             # filter now with filter info
             outFacts = set()
             for fact in facts:
-                if fact.isItem:
-                    for varFilterRel, dimQname in complHandledFilterRels:
-                        dim = fact.context.qnameDims.get(dimQname)
-                        if dim is None:
-                            outFacts.add(fact)
+                if (
+                    fact.isItem
+                    and all(fact.context.qnameDims.get(dimQname) is None for _, dimQname in complHandledFilterRels)
+                ):
+                    outFacts.add(fact)
             facts = outFacts
     return facts
 
--- a/tests/unit_tests/arelle/formula/test_formula_evaluator.py
+++ b/tests/unit_tests/arelle/formula/test_formula_evaluator.py
@@ -1,0 +1,145 @@
+from dataclasses import dataclass
+from types import SimpleNamespace
+
+import arelle.ModelFormulaObject
+from arelle.ModelValue import qname
+from arelle.formula import FormulaEvaluator
+from arelle.formula.FormulaEvaluator import VariableBinding, trialFilterFacts
+
+
+@dataclass
+class DummyBoundFact:
+    context: str
+
+
+@dataclass
+class DummyContext:
+    qnameDims: dict
+
+
+class DummyItemFact:
+    def __init__(self, name: str, dims: dict) -> None:
+        self.name = name
+        self.isItem = True
+        self.context = DummyContext(dims)
+
+
+class DummyFilter:
+    def aspectsCovered(self, var_binding) -> set:
+        return set()
+
+
+def _handled_filter_rel(dim) -> tuple:
+    return (
+        SimpleNamespace(
+            toModelObject=DummyFilter(),
+            isCovered=False,
+        ),
+        dim,
+    )
+
+
+def _group_filter_var_set(no_compl_dims: tuple = (), compl_dims: tuple = ()) -> SimpleNamespace:
+    return SimpleNamespace(
+        filterInfo=True,
+        noComplHandledFilterRels=[_handled_filter_rel(dim) for dim in no_compl_dims],
+        complHandledFilterRels=[_handled_filter_rel(dim) for dim in compl_dims],
+        unHandledFilterRels=[],
+    )
+
+
+def test_fact_variable_evaluation_results_yield_fallback_after_facts_exist() -> None:
+    vb = VariableBinding.__new__(VariableBinding)
+    vb.isFactVar = True
+    vb.isBindAsSequence = False
+    vb.facts = [DummyBoundFact("c1")]
+    vb.values = ["fallback"]
+    vb.yieldedFact = None
+    vb.yieldedFactContext = None
+    vb.yieldedEvaluation = None
+    vb.isFallback = False
+
+    results = list(vb.evaluationResults)
+
+    assert results == [vb.facts[0], vb.values]
+
+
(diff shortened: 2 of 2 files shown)


ignatremizov:
@austinmatherne-wk  Thanks for the detailed review.

On the reproduction side, I filed an issue here: #2257.

The synthetic package in that issue is a shared reproducer, but it was not the original discovery case. The problem was first observed while validating a full DORA xBRL-CSV filing for submission in Arelle. That filing did pass EBA's validation flow afterwards, but current Arelle still emits unexpected `B_05.01` assertion failures. The same behavior also shows up when validating the extracted `report.json` directly in the GUI, so the issue does not look like a simple ZIP-versus-extracted-package issue.

Before treating this as an Arelle bug, I spent time trying to disprove that and rule out a filing-content problem on the real report itself. The sequence there was:

1. I compared the reported errors against the actual `B_05.01` data and found that the visible row content did not match the failures Arelle was reporting.
2. I then tested whether sparse non-direct ICT provider rows were the trigger, first by filling `0090` and `0100` while leaving `0030` and `0040` empty.
3. When that did not change the behavior, I tried a more aggressive variation and filled the remaining `B_05.01` cells as well.
4. The failures still persisted, and in the "fully filled" variation the failing messages actually shifted from `v8854_m_0` / `v8855_m_0` to a broader `v8850_m_0` through `v8854_m_0` pattern.

That last point was important, because it showed that the apparent "missing" field was changing with the rule being evaluated, not with the actual CSV content. In other words, the issue stopped looking like "some columns are blank" and started looking like "the evaluator is binding the wrong facts for the row."

Only after that did I move into source-level investigation. I cloned Arelle, reproduced the same behavior locally, and then used the real filing plus the taxonomy files to trace where the row bindings were going wrong. I also considered whether this could be explained by package loading, cache state, or an outdated plug-in, but the failures were taxonomy formula assertion messages, not plug-in-specific filing-rule messages, so the main fault line looked to be in formula evaluation.

The code investigation itself also went through a few iterations.

My first experimental change was in `implicitFilter()`, because that was where the bad bindings were becoming visible during evaluation, but that did not resolve the issue, but it also did help narrow where to look next.

I then changed `VariableBinding.evaluationResults()` while investigating whether fallback evaluation was being emitted in places where a real fact evaluation had already been bound. That did address part of what I was seeing locally, but it was still not the main cause of the DORA false positives.

The core issue, as I understand it now, is in the optimized handled-dimension filtering path in `trialFilterFacts()`. In this part of Arelle, a "fact" is one concrete reported value from the filing, and a "formula" assertion is evaluated by binding a set of fact variables to the facts that are supposed to belong to the same logical row or context. A fact is not identified only by its concept name. It also carries contextual qualifiers, including dimensions. In this DORA case, one of the important dimensions is `RPR`, and different `B_05.01` columns live on different `RPR` members, for example `x0`, `x211`, and `x53`.

That means the engine cannot just ask "do I have a fact for this concept?" It has to ask "do I have the fact for this concept in the right dimensional context for this row?" The variable filters in the taxonomy are expressing exactly that. They are trying to say, in effect, "for this assertion, bind the `B_05.01` facts that belong to the same row signature, including the required handled dimensions."

What I believe is happening in the current code is that this early filtering step is too permissive when more than one handled dimension is involved. Instead of requiring the whole handled-dimension signature to match, it effectively lets a fact through if it matches only part of that signature. It is acting like "this fact matches enough of the row" instead of "this fact matches the full row definition."

In the DORA case, that matters because `B_05.01` is not a flat table where every relevant value is in the same aspect shape. For example, one rule may be trying to evaluate a row where:

- `0020` comes from a fact on one dimensional combination
- `0080` comes from a fact on a different `RPR` member
- `0110` and `0120` come from yet another `RPR` member

If the early filter allows a fact from the wrong table/context to survive just because it shares the concept name or only part of the handled dimensions, Arelle then starts the later formula evaluation from the wrong candidate set. Once that happens, the rest of the row no longer lines up correctly. The assertion still tries to bind the other row-mate variables, but now it is binding them relative to the wrong surviving fact. That is why the error messages end up saying things like "if `0020` is present then `0080` must be present" and then show `0080` as `None`, even though the filing really does contain `0080` for that row.

An important clue was that the supposedly "missing" field changed depending on the rule. In one run Arelle would say `0020` was present but `0080` or `0110` was missing. In another variation, after more of the row was filled in, the failures shifted and Arelle instead reported a broader `v8850_m_0` to `v8854_m_0` pattern. That is much easier to explain as a fact-binding problem than as a CSV-content problem: the data was there, but the formula engine was assembling the row from the wrong candidate facts.

So the `trialFilterFacts()` change makes that first-stage filtering align with the taxonomy's row definition: a fact should survive only if it matches the full handled-dimension signature for that variable, not just part of it. Once that early filtering is corrected, the later formula evaluation sees the right row-mate facts together, and the false `None` values behind `v8850_m_0` to `v8855_m_0` disappear.

Concretely, the fix requires:

- non-complement handled dimensions to all be present
- complement handled dimensions to all be absent

The CI regressions were useful in separating the real fix from the earlier experiments:

- the first draft regressed KVK validation suites, which showed that the `implicitFilter()` change was a false lead and should be removed
- after that, the next failures were Formula 1.0 "Bind To Empty Sequence" conformance tests, which showed that my initial `evaluationResults()` fallback suppression was too broad

So the final patch is narrower than the earlier drafts:

- the `implicitFilter()` change was removed entirely
- the `evaluationResults()` change was tightened so that fallback behavior is preserved for sequence bindings and suppressed only after a real non-sequence fact evaluation has already been yielded
- the `trialFilterFacts()` handled-dimension change remains the core DORA fix

So the TLDR is:

1. observe the bug on a real filing that had passed EBA validation
2. try multiple data variations on the real filing to rule out a content problem
3. conclude from those experiments that the failures were not tracking the actual CSV values
4. clone Arelle and reproduce the problem locally from source
5. try a couple of narrower code-level hypotheses first (`implicitFilter()`, then `evaluationResults()`)
6. identify `trialFilterFacts()` handled-dimension filtering as the real fault line
7. use CI regressions to remove the false lead and tighten the fallback behavior change
8. create the minimized synthetic reproducer in #2257 afterward so you and other reviewers have a shareable testcase independent of the original filing

On the CLA: I have already submitted it to `support@arelle.com`.

PR #2246, opened 2026-03-18
Add type hints to Aspect.py and update mypy configuration

volodymyrtur-wk:
#### Steps to Test
CI
**review**:
@Arelle/arelle

### DIFF
--- a/arelle/Aspect.py
+++ b/arelle/Aspect.py
@@ -33,7 +33,7 @@ class Aspect:
         OMIT_DIMENSIONS: "omit dimensions",
         PRECISION: "precision",
         DECIMALS: "decimals",
-        }
+    }
 
 
 def aspectStr(aspect: int | QName) -> str:
@@ -91,6 +91,7 @@ aspectFromToken: dict[str, int] = {
 
 aspectToToken: dict[int, str] = {
     Aspect.LOCATION: "location", Aspect.CONCEPT: "concept",
+    Aspect.ENTITY_IDENTIFIER: "entity-identifier",
     Aspect.PERIOD: "period", Aspect.UNIT:"unit",
     Aspect.NON_XDT_SEGMENT: "non-XDT-segment",
     Aspect.NON_XDT_SCENARIO: "non-XDT-scenario",


arelle/Aspect.py
```diff
@@ -89,18 +89,17 @@ def isDimensionalAspect(aspect):
     "scenario": Aspect.COMPLETE_SCENARIO, "complete-scenario": Aspect.COMPLETE_SCENARIO,
 }
 
-aspectToToken = {
+aspectToToken: dict[int, str] = {
     Aspect.LOCATION: "location", Aspect.CONCEPT: "concept",
-    Aspect.ENTITY_IDENTIFIER: "entityIdentifier", Aspect.ENTITY_IDENTIFIER: "entity-identifier",
```
austinmatherne-wk, line 94: Looks like both `ENTITY_IDENTIFIER` entries were removed instead of just the overridden one.

arelle/Aspect.py
```diff
@@ -33,32 +33,32 @@ class Aspect:
         OMIT_DIMENSIONS: "omit dimensions",
         PRECISION: "precision",
         DECIMALS: "decimals",
-    }
+        }
```
austinmatherne-wk: nit: we typically indent closing brackets to align with the first character of the line that starts the construct like the original code did here. This is commonly [how formatters](https://black.readthedocs.io/en/stable/the_black_code_style/current_style.html#how-black-wraps-lines) (black/ruff) in the python ecosystem handle this.

volodymyrtur-wk: Thanks, accidentally moved it

arelle/Aspect.py
```diff
@@ -89,18 +89,18 @@ def isDimensionalAspect(aspect):
     "scenario": Aspect.COMPLETE_SCENARIO, "complete-scenario": Aspect.COMPLETE_SCENARIO,
 }
 
-aspectToToken = {
+aspectToToken: dict[int, str] = {
     Aspect.LOCATION: "location", Aspect.CONCEPT: "concept",
-    Aspect.ENTITY_IDENTIFIER: "entityIdentifier", Aspect.ENTITY_IDENTIFIER: "entity-identifier",
+    Aspect.ENTITY_IDENTIFIER: "entityIdentifier",
```
austinmatherne-wk: Looks like the wrong value for the duplicate `ENTITY_IDENTIFIER` entry was deleted.

Issue #2248, opened 2026-03-18
Breaking Change: Plugin Manager Class


aaroncameron-wk:
## Overview
As part of a strategic initiative to reduce the use of global state within Arelle, the `PluginManager` module is being refactored into an instantiable class. This architecture enables `PluginManager` instances to be explicitly owned and injected by `Cntlr` instances, facilitating improved isolation and concurrency. Legacy touchpoints, such as `pluginClassMethods`, must now be invoked via the controller-provided instance (e.g., `modelXbrl.modelManager.cntlr.pluginManager.pluginClassMethods(...)`).

### Users Affected
- **Python API:** Consumers interacting with module-level functions and static variables within the `PluginManager` namespace.
  - [ ] Provide migration instructions.
- **Plugin Hooks:** Maintainers defining or invoking custom plugin hooks must transition from direct `PluginManager` module imports to referencing the active controller-level `pluginManager` instance.
  - [ ] Provide migration instructions.

## Rollout Plan
In accordance with the [breaking change rollout lifecycle](https://arelle.readthedocs.io/en/latest/versioning.html), the transition is scheduled as follows:
- [x] Announcement (March 25, 2026)
- [x] Phase 1: Opt-In
  - [x] **Python API and Plugin Hooks:** The controller-resident `PluginProvider` instance becomes available. Module-level functions remain functional as a compatibility layer, with both patterns temporarily sharing a singleton state to ensure stability.
- [ ] Phase 2: Opt-Out / Deprecation
  - [ ] **Python API and Plugin Hooks:** Usage of module-level functions and variables will trigger formal deprecation warnings (`DeprecationWarning`).
- [ ] Phase 3: Removal
  - [ ] **Python API and Plugin Hooks:** Module-level functions and variables are decommissioned. Each `Cntlr` instance will utilize a dedicated, isolated `PluginManager` instance, finalized for multi-tenant environments.

PR #2250, opened 2026-03-18
`PluginManager` Class

aaroncameron-wk:
> [!IMPORTANT]
> Review by commit *with whitespace hidden*. Changes were structured in a way to minimize code movement, but this is only realized without whitespace changes.

#### Reason for change
This is the non-breaking first phase of #2248.
Current plan is to eventually be followed by the addition of deprecation warnings with the eventual removal of the singleton instance and module-level variables and functions altogether.

#### Description of change
Introduces a `PluginManager` class. Temporarily defines a singleton instance of that class to maintain current functionality. Internal references to the module-level `PluginManager` variables and functions are replaced with references through the controller, (e.g., `modelXbrl.modelManager.cntlr.pluginManager.pluginClassMethods`)

#### Steps to Test
CI

**review**:
@Arelle/arelle

### DIFF
--- a/arelle/logging/handlers/LogToXmlHandler.py
+++ b/arelle/logging/handlers/LogToXmlHandler.py
@@ -191,10 +191,12 @@ class LogToXmlHandler(LogHandlerWithXml):
             self.htmlTitle)
         ]
         if self.logRecordBuffer:
+            filters = None
             if self.cntlr is not None:
-                for logRec in self.logRecordBuffer:
-                    if all(p(logRec) for p in self.cntlr.pluginManager.pluginClassMethods("Cntlr.Log.RecFilter.Html")):
-                        html.append(self.recordToHtml(logRec))
+                filters = self.cntlr.pluginManager.pluginClassMethods("Cntlr.Log.RecFilter.Html")
+            for logRec in self.logRecordBuffer:
+                if not filters or all(_filter(logRec) for _filter in filters):
+                    html.append(self.recordToHtml(logRec))
             if clearLogBuffer:
                 self.clearLogBuffer()
             html.append("</tbody>\n</table>\n</body>\n</html>\n")


arelle/logging/handlers/LogToXmlHandler.py
```diff
@@ -180,9 +187,10 @@ def getHtml(self, clearLogBuffer: bool = True) -> str:
             self.htmlTitle)
         ]
         if self.logRecordBuffer:
-            for logRec in self.logRecordBuffer:
-                if all(p(logRec) for p in PluginManager.pluginClassMethods("Cntlr.Log.RecFilter.Html")):
-                    html.append(self.recordToHtml(logRec))
+            if self.cntlr is not None:
```
volodymyrtur-wk, line 195: You can union 2 conditions in one, like: 
```
if self.logRecordBuffer and self.cntlr is not None:
    for logRec in self.logRecordBuffer:
        if all(p(logRec) for p in self.cntlr.pluginManager.pluginClassMethods("Cntlr.Log.RecFilter.Html")):
            html.append(self.recordToHtml(logRec))

File: arelle/PluginManager.py [hunk 2]
```diff
@@ -39,755 +39,736 @@
...
 
-    _pluginConfig = _getPluginConfig()
-    orderedModules = {
-        moduleName: sortModuleInfo(_pluginConfig['modules'][moduleName])
-        for moduleName in sorted(_pluginConfig['modules'].keys())
+    def save(self, cntlr: Cntlr) -> None:
+        if self.pluginConfigChanged and cntlr.hasFileSystem and not cntlr.disablePersistentConfig:
+            pluginJsonFile = cntlr.userAppDir + os.sep + "plugins.json"
+            with open(pluginJsonFile, 'w', encoding='utf-8') as f:
+                jsonStr = str(json.dumps(self.orderedPluginConfig(), ensure_ascii=False, indent=2)) # might not be unicode in 2.7
```
volodymyrtur-wk: Since Arelle no longer supports `Python 2.7` and `json.dumps` returns a string, we can remove the forced type cast to string.

arelle/PluginManager.py
```diff
@@ -39,755 +39,736 @@
...
+        When `entryPoint` is provided, it's location and other details will be used to generate the module info
+        dictionary.
+
+        :param moduleURL: A URL that loosely maps to the file location of a plugin (may be transformed)
+        :param entryPoint: An `EntryPoint` instance
+        :param reload:
+        :param parentImportsSubtree:
+        :return:s
+        """
+        if (moduleURL is None) == (entryPoint is None):
```
volodymyrtur-wk, line 463: operator `is` more accurate to check boolean values

arelle/PluginManager.py
```diff
@@ -39,755 +39,736 @@
...
-            pass # on GAE no userAppDir, will always come here
-    if not pluginConfig:
-        pluginConfig = {  # savable/reloadable plug in configuration
-            "modules": {}, # dict of moduleInfos by module name
-            "classes": {}  # dict by class name of list of class modules in execution order
+class PluginManager:
+
+    def __init__(self) -> None:
+        self.pluginConfig: dict[str, Any] = {}
+
```
volodymyrtur-wk, line 50: Python's traditional method for creating a class as a singleton pattern is to override the `__new__` constructor.
```
class PluginManager:
    _instance = None

    def __new__(cls, *args, **kwargs):
        if cls._instance is None:
            cls._instance = super().__new__(cls)
        return cls._instance
```
in this case, no matter how many times the PluginManager is created, it will always be the same

arelle/logging/handlers/LogToXmlHandler.py
aaroncameron-wk: This would unnecessarily change behavior: `clearLogBuffer` would not be called despite not relying on a `cntlr` value.

arelle/PluginManager.py
aaroncameron-wk: This is new information to me, thanks! 

With that approach, if I'm understanding correctly, the `__init__` method is still called on subsequent `PluginManager()` calls, which would reset `pluginConfig` which is not intended.

The nature of this being a singleton is specific to preserving existing behavior and is temporary, so I lean towards leaving it as-is.

arelle/PluginManager.py
aaroncameron-wk: I made an effort to not make any changes that weren't strictly necessary for this PR since it's large and more difficult to review as-is. A future step in this larger refactor effort will provide good opportunity to make these kind of changes, so I'll update this at that point.

arelle/PluginManager.py
volodymyrtur-wk: Yes, it will. To prevent such behavior, you can override `__init__` something like:
```
def __init__(self, value=None):
    if not hasattr(self, "_initialized"):
        self.pluginConfig: dict[str, Any] = {}
        self._initialized = True
```
This is not a request to change, just FYI

arelle/logging/handlers/LogToXmlHandler.py
volodymyrtur-wk: I don't fully understand the answer. I'll explain my point of view. You have 2 required conditions before starting the loop. If the first condition `if self.logRecordBuffer:`  is True, we go to the next condition `if self.cntlr is not None:` and if this condition is True as well, we go to the next step and run the loop. My suggestion with `if self.logRecordBuffer and self.cntlr is not None:` does the same. Second part of the condition `self.cntlr is not None` will evaluate only if the first part `self.logRecordBuffer` is True.

aaroncameron-wk: This code is also beneath that first `if` condition:
```
if clearLogBuffer:
	self.clearLogBuffer()
html.append("</tbody>\n</table>\n</body>\n</html>\n")
```
I still want that to execute when there's not `cntlr`.

arelle/logging/handlers/LogToXmlHandler.py
```diff
@@ -180,9 +187,10 @@ def getHtml(self, clearLogBuffer: bool = True) -> str:
...
+                for logRec in self.logRecordBuffer:
+                    if all(p(logRec) for p in self.cntlr.pluginManager.pluginClassMethods("Cntlr.Log.RecFilter.Html")):
+                        html.append(self.recordToHtml(logRec))
```
austinmatherne-wk: Maybe I'm not reading this right, but if we don't have a controller for some reason and can't check for plugin filters, wouldn't we want to append `logRec` (no controller -> no plugins -> no filters -> do the thing)?

arelle/PluginManager.py
```diff
@@ -39,755 +39,736 @@
...
+    def init(self, cntlr: Cntlr, loadPluginConfig: bool = True) -> None:
+        self.pluginJsonFile: str | None = None
+        self._cntlr: Cntlr = cntlr
+        self.pluginConfigChanged = False
+        self.pluginTraceFileLogger: logging.Logger | None = None
+        self.modulePluginInfos: dict[str, Any] = {}
+        self.pluginMethodsForClasses: dict[str, Any] = {}
```
austinmatherne-wk, line 57: Any particular reason not to assign these to defaults in `__init__`? I'd expect that to help with typing.

arelle/logging/handlers/LogToXmlHandler.py
```diff
@@ -18,26 +21,30 @@ class LogToXmlHandler(LogHandlerWithXml):
...
         self.filemode = mode
 
     def flush(self) -> None:
         # Note to developers: breakpoints in this method don't work, please debug with print statements
         securityIsActive = securityHasWritten = False
-        for pluginMethod in PluginManager.pluginClassMethods("Security.Crypt.IsActive"):
-            securityIsActive = pluginMethod(self) # must be active for the cntlr object to effect log writing
+        if self.cntlr is not None:
+            for pluginMethod in self.cntlr.pluginManager.pluginClassMethods("Security.Crypt.IsActive"):
+                securityIsActive = pluginMethod(self) # must be active for the cntlr object to effect log writing
```
austinmatherne-wk: Skipping these security checks makes me nervous (I don't really understand how the SEC uses them). Better to raise an exception then silently skip and potentially do something very bad?

arelle/PluginManager.py
```diff
@@ -944,20 +936,155 @@ def _normalizePluginSearchTerm(search: str) -> str:
...
+        _singleton.save(cntlr)
+
+
+def close() -> None:
+    if _singleton is not None:
+        _singleton.close()
+
+
+def logPluginTrace(message: str, level: int) -> None:
+    if _singleton is not None:
```
austinmatherne-wk: Here and the surrounding functions: should we raise an exception instead of silently skipping? Or in this case, fallback to standard logging?

arelle/Cntlr.py
```diff
@@ -289,6 +290,7 @@ def __init__(
 
         # start plug in server (requres web cache initialized, but not logger)
         PluginManager.init(self, loadPluginConfig=hasGui)
+        self.pluginManager = PluginManager._singleton  # type: ignore[assignment]
```
austinmatherne-wk: Instead of accessing a private field from another module could we add a `PluginManager.getInstance()` method?

arelle/PluginManager.py
aaroncameron-wk: `__init__` is only called once when the singleton is instantiated. `init` is called multiple times (every `Cntlr` instantiation calls it, and every command line run [instantiates](https://github.com/Arelle/Arelle/blob/d00752f8af783073d4a94b105c9a73843d58e7f5/arelle/CntlrCmdLine.py#L172) a temporary `CntlrCmdLine` for pre-parsing arguments), so it is effectively a reset method.

arelle/PluginManager.py
aaroncameron-wk: I left `reset` and `close` as I think it's appropriate for those to do nothing in this scenario (it might even be required based on our usage). Let me know if you feel strongly that we should have those be more strict.

austinmatherne-wk: Those make sense to me, too.

Issue #2242, opened 2026-03-17
Very poor formula performance on ESEF filing with duplicate text facts


paulwarren-wk:
### What happened?

Attempting to validate [this ESEF filing](https://filings.xbrl.org/213800EC6PW5VU4J9U64/2025-12-31/ESEF/FI/1/213800EC6PW5VU4J9U64-2025-12-31.zip) with formula validation enabled appears to crash - process is stuck on 100% CPU and is still running days later.

The offending assertion appears to be this one:

```xml
<va:valueAssertion xlink:type="resource" xlink:label="valueAssertion_10" id="fac_TextFactMustBeAssignedWithOrInheritedLangAttribute" test="if (not(empty($langAttribute))) then (count($textFact) ge 0) else (count($textFact) eq 0)" aspectModel="dimensional" implicitFiltering="true"/>
    <variable:factVariable xlink:type="resource" xlink:label="factVariable_3" id="fv_013" bindAsSequence="true"/>
    <variable:variableArc xlink:type="arc" xlink:arcrole="http://xbrl.org/arcrole/2008/variable-set" xlink:from="valueAssertion_10" xlink:to="factVariable_3" name="textFact"/>
    <cf:conceptDataType xlink:type="resource" xlink:label="conceptDataType_3" id="dt_003" strict="false">
      <cf:type>
        <cf:qnameExpression>QName ('http://www.w3.org/2001/XMLSchema' , 'string')</cf:qnameExpression>
      </cf:type>
    </cf:conceptDataType>
    <variable:variableFilterArc xlink:type="arc" xlink:arcrole="http://xbrl.org/arcrole/2008/variable-filter" xlink:from="factVariable_3" xlink:to="conceptDataType_3" cover="true" complement="false"/>
    <gf:general xlink:type="resource" xlink:label="general_2" id="ge_002" test="not(@xml:lang)"/>
    <variable:variableFilterArc xlink:type="arc" xlink:arcrole="http://xbrl.org/arcrole/2008/variable-filter" xlink:from="factVariable_3" xlink:to="general_2" order="2.0" cover="true" complement="false"/>
    <variable:generalVariable xlink:type="resource" xlink:label="generalVariable_9" id="gv_016" select="data(//xbrli:xbrl/@xml:lang)" bindAsSequence="false"/>
[...]

PR #2252, opened 2026-03-19
Fix performance bug in evaluationIsUnnecessary

austinmatherne-wk:
#### Reason for change
See #2242 

When no hash matches were found for indexed variables, the fallback `matchingEvals = otherEvals` caused the final duplicate check to scan all prior evaluations. This made the ESEF formula validation rule in #2242 take days instead of minutes.

#### Description of change
Add an early exit check after the fallback dependency exclusion. If any variable that will actually be compared has a hash not seen in prior evaluations, no prior evaluation can match, so return immediately. Reduces per evaluation cost for novel evaluations (time per evaluation stays at a constant ~0.001 seconds in #2242 instead of increasing over time).

#### Steps to Test
* Use the doc in #2242 and confirm it validates in minutes instead of days.

**review**:
@Arelle/arelle

### DIFF
--- a/arelle/formula/FormulaEvaluator.py
+++ b/arelle/formula/FormulaEvaluator.py
@@ -1092,63 +1092,64 @@ def factsPartitions(xpCtx, facts, aspects):
 
 def evaluationIsUnnecessary(thisEval, xpCtx):
     otherEvals = xpCtx.evaluations
-    if otherEvals:
-        otherEvalHashDicts = xpCtx.evaluationHashDicts
-        if all(e is None for e in thisEval.values()):
-            # evaluation not necessary, all fallen back
-            return True
-        nonNoneEvals = {
-            vQn: vBoundFact
-            for vQn, vBoundFact in thisEval.items()
-            if vBoundFact is not None
-        }
-        # hash check if any hashes merit further look for equality
-        otherEvalSets = [
-            otherEvalHashDicts[vQn][hash(vBoundFact)]
-            for vQn, vBoundFact in nonNoneEvals.items()
-            if vQn in otherEvalHashDicts
-            if hash(vBoundFact) in otherEvalHashDicts[vQn]
-        ]
-        if otherEvalSets:
-            matchingEvals = [otherEvals[i] for i in set.intersection(*otherEvalSets)]
-        else:
-            matchingEvals = otherEvals
-        # find set of vQn whose dependencies are fallen back in this eval but not others that match
-        varBindings = xpCtx.varBindings
-        vQnDependent = set(
-            vQn
-            for vQn in nonNoneEvals
-            if vQn in varBindings
+    if not otherEvals:
+        return False
+
+    otherEvalHashDicts = xpCtx.evaluationHashDicts
+    if all(e is None for e in thisEval.values()):
+        # evaluation not necessary, all fallen back
+        return True
+    nonNoneEvals = {
+        vQn: vBoundFact
+        for vQn, vBoundFact in thisEval.items()
+        if vBoundFact is not None
+    }
+    # hash check if any hashes merit further look for equality
+    otherEvalSets = [
+        otherEvalHashDicts[vQn][hash(vBoundFact)]
+        for vQn, vBoundFact in nonNoneEvals.items()
+        if vQn in otherEvalHashDicts
+        if hash(vBoundFact) in otherEvalHashDicts[vQn]
+    ]
+    if otherEvalSets:
+        matchingEvals = [otherEvals[i] for i in set.intersection(*otherEvalSets)]
+    else:
+        matchingEvals = otherEvals
+    # find set of vQn whose dependencies are fallen back in this eval but not others that match
+    varBindings = xpCtx.varBindings
+    vQnDependent = set(
+        vQn
+        for vQn in nonNoneEvals
+        if vQn in varBindings
+        and any(
+            varBindings[varRefQn].isFallback
             and any(
-                varBindings[varRefQn].isFallback
-                and any(
-                    m[varRefQn] is not None
-                    for m in matchingEvals
-                )
-                for varRefQn in varBindings[vQn].var.variableRefs()
-                if varRefQn in varBindings
+                m[varRefQn] is not None
+                for m in matchingEvals
             )
+            for varRefQn in varBindings[vQn].var.variableRefs()
+            if varRefQn in varBindings
         )
-        evalsNotDependent = [
-            (vQn, vBoundFact)
-            for vQn, vBoundFact in nonNoneEvals.items()
-            if vQn not in vQnDependent
-        ]
-        # If any compared variable has an indexed hash not seen in prior
-        # evaluations, no prior evaluation can match on all compared
-        # variables simultaneously, so this evaluation is unique.
-        for vQn, vBoundFact in evalsNotDependent:
-            if vQn in otherEvalHashDicts and hash(vBoundFact) not in otherEvalHashDicts[vQn]:
-                return False
-        # detects evaluations which are not different (duplicate) and extra fallback evaluations
-        # vBoundFact may be single fact or tuple of facts
-        return any(
-            all(
-                vBoundFact == matchingEval[vQn]
-                for vQn, vBoundFact in evalsNotDependent
-            ) for matchingEval in matchingEvals
-        )
-    return False
+    )
+    evalsNotDependent = [
+        (vQn, vBoundFact)
+        for vQn, vBoundFact in nonNoneEvals.items()
+        if vQn not in vQnDependent
+    ]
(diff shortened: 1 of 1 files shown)


arelle/formula/FormulaEvaluator.py
```diff
@@ -1128,6 +1128,12 @@ def evaluationIsUnnecessary(thisEval, xpCtx):
             for vQn, vBoundFact in nonNoneEvals.items()
             if vQn not in vQnDependentOnOtherVarFallenBackButBoundInOtherEval
         ]
+        # If any compared variable has an indexed hash not seen in prior
+        # evaluations, no prior evaluation can match on all compared
+        # variables simultaneously, so this evaluation is unique.
+        for vQn, vBoundFact in evalsNotDependentOnVarFallenBackButBoundInOtherEval:
+            if vQn in otherEvalHashDicts and hash(vBoundFact) not in otherEvalHashDicts[vQn]:
+                return False
```
brettkail-wk: I don't know this code at all.  I've been staring at it for awhile, and I'm still not at all able to decipher commit f0989c9b44bc2cf58a728d5aceb70adcecb450ec (or [Variables 1.0 section 4 bullet 3](https://www.xbrl.org/Specification/variables/per-2011-11-30/variables-per-2011-11-30.html#sec-variable-evaluation)), but fully iterating over all other evals somehow feels like a problem.  I wonder if that commit was wrong to introduce `matchingEvals = otherEvals`, and instead it should have changed `otherEvalSets` to be built based on `evalsNotDependent` rather than `nonNoneEvals`.  The `any`/`all` below requires `vBoundFact` to match anyway, so why do we need to linearly scan all evals rather than looking up just the ones that match `vBoundFact`?  (For the rest of this comment, let's assume I just misunderstand the code, and we're proceeding with the PR as-is.)

Given both `evalsNotDependent` and `otherEvalSets` are both built from `nonNoneEvals`, it seems like we could exit a lot earlier and avoid duplicating the `hash(vBoundFact) in otherEvalHashDicts[vQn]`?
```
otherEvalSets = []
for vQn, vBoundFact in nonNoneEvals.items()
  if vQn in otherEvalHashDicts:
    if hash(vBoundFact) not in otherEvalHashDicts[vQn]:
      return False
    otherEvalSets.append(otherEvalHashDicts[vQn][hash(vBoundFact)])
```

Pre-existing, it might be nice to rename `vQnDependent` to `evalsDependent` to match all the other names.  Also pre-existing, it's kind of odd to build the full `vQnDependent...` when it only exists just to build the inverse.  It seem the two vars could be collapsed?
```
        evalsNotDependentOnVarFallenBackButBoundInOtherEval = [
            (vQn, vBoundFact)
            for vQn, vBoundFact in nonNoneEvals.items()
            if not (...big vQnDependentOnOtherVarFallenBackButBoundInOtherEval expression...)
        ]
```

austinmatherne-wk: This code melts my brain, but the conformance suite covers these scenarios well so I'm not flying blind.

I tried moving the hash check into the `otherEvalSets` construction. It breaks four conformance tests with evaluations that should have been suppressed.

The issue is that `nonNoneEvals` includes variables that will later be excluded. My understanding of the linked spec bullet is that if variable `$A` depends on a currently fallback variable `$B`, then `$A` gets excluded from the duplicate check. If the early exit sees a novel hash for `$A` and returns `False`, it short circuits before discovering that `$A` would have been excluded and the evaluation is actually a duplicate. That's why the placement after the dependency exclusion matters.

On `matchingEvals = otherEvals`, that's due to a circular dependency that I'm not sure how you'd avoid otherwise. You can't build `otherEvalSets` from `evalsNotDependent` because `evalsNotDependent` is built from `matchingEvals`, and `matchingEvals` is derived from `otherEvalSets`. So when hash narrowing comes up empty, it falls back to all prior evaluations as the candidate set for the dependency analysis.

Collapsing `vQnDependent...` would work, but the collapsed comprehension ends up with three levels of nested `any()` inside a negation. Given how hard this function already is to reason about, I think the intermediate set earns its keep as a waypoint for the reader.

On the rename, at least to me, the `vQn` prefix is useful since it tells me the set contains QNames, whereas the `evals` prefix on `evalsNotDependent...` is referring to (QName, boundFact) tuples.

brettkail-wk: > That's why the placement after the dependency exclusion matters.

Thanks, I (kind of) understand that now.

> You can't build `otherEvalSets` from `evalsNotDependent` because `evalsNotDependent` is built from `matchingEvals`

Thanks, I see the circularity.

It seems like [this expression](https://github.com/Arelle/Arelle/pull/2252/changes#diff-b7b994550682d637618fbdd2a2edc7ecc97c1da64c47f67ec5c6e34c80f97936R1118-R1121) is checking "does there exist any other evaluation with the qname", so it seems like that could be simplified to `varRefQn in otherEvalHashDicts`:
```
        vQnDependentOnOtherVarFallenBackButBoundInOtherEval = set(
            vQn
            for vQn, vBoundFact in nonNoneEvals.items()
            if vQn in varBindings
            and any(
                varBindings[varRefQn].isFallback
                and varRefQn in otherEvalHashDicts
                for varRefQn in varBindings[vQn].var.variableRefs()
                if varRefQn in varBindings
            )
        )
```
Does that change alone work?  If no, I'm still missing some detail (would not be surprising :-)).  If yes, that seems like it would be an improvement since it removes one walk over "all previous evals".  It would also seem to break the cycle so that `otherEvalSets` can be pushed down and be based on `evalsNotDependent`?  Somewhat regardless, even if I'm mis-analyzed that, it seems like [this expression](https://github.com/Arelle/Arelle/blob/d00752f8af783073d4a94b105c9a73843d58e7f5/arelle/formula/FormulaEvaluator.py#L1139-L1144) could use an expression similar to `otherEvalSets` but limited to non-dependent qnames rather than all qnames?

> On the rename, at least to me, the `vQn` prefix is useful since it tells me the set contains QNames, whereas the `evals` prefix on `evalsNotDependent...` is referring to (QName, boundFact) tuples.

Thanks, I understand now, the vars make sense.  Maybe it would help if the vars were just shorter (`vQnDependent` and `evalsNotDependent`?)...

austinmatherne-wk: > it seems like that could be simplified to `varRefQn in otherEvalHashDicts`

`otherEvalHashDicts` includes evaluations that differ from the current one on other variable bindings. I tried the suggested change to confirm my understanding and it does indeed fail the conformance suite.

> it seems like [this expression](https://github.com/Arelle/Arelle/blob/d00752f8af783073d4a94b105c9a73843d58e7f5/arelle/formula/FormulaEvaluator.py#L1139-L1144) could use an expression similar to otherEvalSets but limited to non-dependent qnames rather than all qnames?

I'm not sure I understand the suggestion here. Can you elaborate?

> Maybe it would help if the vars were just shorter (vQnDependent and evalsNotDependent?)

Sure. Updated.

brettkail-wk: Discussed offline.  This PR doesn't seem wrong to me (so the team can/should review so that we can unblock the linked issue).

However, I think it only partially solves the root problem, which is that `else: matchingEvals = otherEvals` means we're going to walk over every prior evaluation, and it does so by adding another layer of complexity.

> > it seems like that could be simplified to `varRefQn in otherEvalHashDicts`
>
>  does indeed fail the conformance suite

Focusing exclusively on `else: matchingEvals = otherEvals` case, it _seems_ like this logic is saying "did some prior evaluation contain this `varRefQn`":
```
            and any(
                m[varRefQn] is not None
                for m in matchingEvals
            )
```
...which seems like it should be equivalent to `varRefQn in otherEvalHashDicts`.  Austin tried that, but it didn't work, so we have some incomplete understanding.  (As a minor point, I'm not sure how we can be certain that `m[varRefQn]` doesn't ever produce `KeyError`; perhaps it's that all evaluations are guaranteed to have the same qnames?)

While writing this, it occurs to me that perhaps the change fails because it breaks the `if: matchingEvals = ...` case because my change is checking "did any prior evaluation contain this `varRefQn`" but it needs to check specifically "did any prior _matching_ evaluation contain this `varRefQn`".  (I don't understand why that's important, but I have yet to understand that Variables 1.0 bullet :-).)  Perhaps we can recover the optimization by defining `vQnDependent` differently depending on the `matchingEvals` cases.

> I'm not sure I understand the suggestion here. Can you elaborate?

Similarly, the final `return any(...all(...))` in the `else: matchingEvals = otherEvals` case doesn't need to look at all possible evals, it can use `otherEvalHashDicts` similar to `otherEvalSets` but limited `evalsNotDependent`.

derekgengenbacher-wf:

PR #2256, opened 2026-03-19
Update Dockerfiles to a multi-stage build

ryangeiser1:
#### Reason for change
Update Dockerfiles to a multi-stage build

#### Description of change
Update Dockerfiles to a multi-stage build, add a non-root user, and add entrypoint, start, and healthcheck scripts.

#### Steps to Test
```shell
# Build and run containers
docker compose -f docker/arelle/docker-compose.yml up -d --build
docker compose -f docker/arelle-EDGAR/docker-compose.yml up -d --build
```

**review**:
@Arelle/arelle

### DIFF
--- a/docker/README.md
+++ b/docker/README.md
@@ -1,5 +1,5 @@
 # Docker containers
-Build and run an Arelle webserver with Docker. The arelle/EDGAR image contains the necessary dependencies
+Build and run an Arelle webserver with Docker. The arelle-EDGAR image contains the necessary dependencies
 for the EDGAR plugins (render, validate, and transform) and EFM disclosure systems:
 - Arelle
 - EDGAR plugins
--- a/docs/source/install.md
+++ b/docs/source/install.md
@@ -98,7 +98,7 @@ arelleGUI
 ```
 
 ## Docker containers
-Build and run an Arelle webserver with Docker. The arelle/EDGAR image contains the necessary dependencies
+Build and run an Arelle webserver with Docker. The arelle-EDGAR image contains the necessary dependencies
 for the EDGAR plugins (render, validate, and transform) and EFM disclosure systems:
 - Arelle
 - EDGAR plugins
@@ -113,7 +113,7 @@ cd Arelle
 docker compose -f docker/arelle/docker-compose.yml up -d --build
 
 # or build and run the Arelle container with the EDGAR plugin
-docker compose -f docker/arelle/EDGAR/docker-compose.yml up -d --build
+docker compose -f docker/arelle-EDGAR/docker-compose.yml up -d --build
 ```
 or in one line
 ```shell


docker/EDGAR/Dockerfile
```diff
@@ -1,16 +1,36 @@
...
 RUN pip install --no-cache-dir -r requirements.txt -r requirements-plugins.txt
 ## Optionally install gunicorn to run via wsgi server
-# RUN pip install --no-cache-dir gunicorn
+# RUN pip install --no-cache-dir --upgrade gunicorn
+
+## --------------------- Production Stage --------------------- 
+FROM python:3.14.3 AS production
 
+WORKDIR /usr/src/app
+RUN apt-get update
```
austinmatherne-wk: Both build stages use the full `python:3.14.3` image as the base image, so the production image ends up the same size as a single stage build. It also means `apt-get` update runs twice during the build. What's the benefit you're going for here?

docker/EDGAR/Dockerfile
```diff
@@ -42,12 +62,23 @@ RUN if [ ! -d arelle/plugin/xule ]; then \
...
+COPY docker/start /start
+RUN sed -i 's/\r$//g' /start
+RUN chmod +x /start
 
-## Optionally run with gunicorn with 2 workers, recycling workers after each request
-## WARNING: Uncomment only one CMD at a time.
-# CMD ["gunicorn", "arelle.CntlrCmdLine:wsgiApplication()", "--workers", "2", "--timeout", "180", "--max-requests", "1", "--bind", "0.0.0.0:8080"]
+COPY docker/healthcheck /healthcheck
+RUN sed -i 's/\r$//g' /healthcheck
+RUN chmod +x /healthcheck
```
austinmatherne-wk: Does `COPY --chmod=755 docker/healthcheck /healthcheck` avoid the need to run chmod? Might also be able to avoid `sed` by creating `.gitattributes` with (`docker/* text eol=lf`)

docker/EDGAR/Dockerfile
```diff
@@ -42,12 +62,23 @@ RUN if [ ! -d arelle/plugin/xule ]; then \
         && rm -rf xule; \
     fi
```
austinmatherne-wk, line 46: If we do keep the multistage build, I would expect plugin git clone operations to happen in a build stage. A slim production image likely wouldn't include `git`.

docker/EDGAR/Dockerfile
```diff
@@ -42,12 +62,23 @@ RUN if [ ! -d arelle/plugin/xule ]; then \
...
+RUN chmod +x /start
 
-## Optionally run with gunicorn with 2 workers, recycling workers after each request
-## WARNING: Uncomment only one CMD at a time.
-# CMD ["gunicorn", "arelle.CntlrCmdLine:wsgiApplication()", "--workers", "2", "--timeout", "180", "--max-requests", "1", "--bind", "0.0.0.0:8080"]
+COPY docker/healthcheck /healthcheck
+RUN sed -i 's/\r$//g' /healthcheck
+RUN chmod +x /healthcheck
 
-EXPOSE 8080
```
austinmatherne-wk, line 52: It's probably still a good idea to use `EXPOSE` to document the default port used by `/start`

docker/scripts/healthcheck.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
```
austinmatherne-wk, line 1: May want to consider using `.sh` file extensions for clarity.

docker/EDGAR/Dockerfile
ryangeiser1: Yes, these solutions work. Thank you.

docker/EDGAR/Dockerfile
ryangeiser1: In lieu of using docker compose and the /start script, one could build with `docker build -t edgar-arelle -f docker/EDGAR/Dockerfile .` and then invoke with `docker run edgar-arelle python arelleCmdLine.py --help` to execute a cli command in a new container. One would not want or need to expose a port in this scenario. 

One could also deploy behind a reverse proxy. If the Arelle container and proxy are on the same docker network, the Arelle container does not need to expose a port.

Similarly, one could run the webserver with `docker run -d -p 8080:8080 edgar-arelle python arelleCmdLine.py --webserver=0.0.0.0:8080` if the webserver and ports need to be exposed directly.

If we want to keep `EXPOSE` for documentation purposes, I suggest we add an `ARG` to toggle exposing the port, and make the port value an environment variable in case 8080 is already in use. What are your thoughts?

docker/EDGAR/Dockerfile
ryangeiser1: I agree, I will move the git clone operations to the build stage. 

The .dockerignore file omits .git directories. `docker exec -it edgar-arelle-1 /bin/bash` and then `find . -type d | grep git` inside the container returns no results. The containers shouldn't include the .git directories from Arelle nor the plugins. Are you seeing something different on your end?

docker/EDGAR/Dockerfile
austinmatherne-wk: That makes sense. Thanks for explaining :)

docker/EDGAR/Dockerfile
austinmatherne-wk: To clarify, I was referring to the `git` executable being absent from the production stage image, not the `.git` directories.

When building multistage images, our general practice is to minimize both production image size and security attack surface. It makes deployment faster (when scaling on demand with Kubernetes), and every additional executable or library is a potential vulnerability, so we prefer slim base images that exclude build time tools like `git`.

I'll be honest, I'm not certain all plugins/features will work from a slim image. Arelle has a lot of dependencies, though the painful ones are mostly GUI related which doesn't apply here. Using `python:3.14.3` for the build stage and `python:3.14.3-slim` for production seems like a reasonable starting point, and we can always revert if something breaks.

PR #2258, opened 2026-03-20
Configurable Formula Timeouts

### LINKED ISSUE #2245: Support a timeout parameter for running formula assertions and validation

It would be handy if there were additional RuntimeOptions such as 

- `formulaAssertionTimeout: int | None` ; resets for each assertion
- `formulaReportTimeout: int | None` ; resets for each `Session.run()`

introduced, with perhaps a default of None or 300 seconds.

If either timeout triggers, a suitable validation error can be included explaining that the timeout occurred.

I was imagining the implementation as just sampling elapsed time in various parts of the formula calling stack and throwing a FormulaTimeoutException if the timeout has been exceeded (rather than something more complicated involving threads).

Inspired by the recent formula #2209 and #2242 related issues.

austinmatherne-wk:
#### Reason for change
Resolves #2245

#### Description of change
Leverages the existing formula profiler timer to add two new options:
`--formulaReportTimeout` (aborts all formula processing after timeout)
`--formulaVarSetTimeout` (aborts per variable set evaluation after timeout)

#### Steps to Test
Use the doc from #2242 and verify that after the configured timeout Arelle raises an error and gracefully continues processing.
```sh
python arelleCmdLine.py \
    --internetConnectivity offline \
    --package esef_taxonomy_2024.zip \
    --file 213800EC6PW5VU4J9U64-2025-12-31.zip \
    --validate \
    --formulaReportTimeout 5

python arelleCmdLine.py \
    --internetConnectivity offline \
    --package esef_taxonomy_2024.zip \
    --file 213800EC6PW5VU4J9U64-2025-12-31.zip \
    --validate \
    --formulaVarSetTimeout 5
```

**review**:
@Arelle/arelle

### DIFF
--- a/arelle/formula/ValidateFormula.py
+++ b/arelle/formula/ValidateFormula.py
@@ -852,6 +852,7 @@ def validate(val, xpathContext=None, parametersOnly=False, statusMsg='', compile
                                 varSetTimeoutTimer.start()
                             else:
                                 varSetTimeoutTimer = None
+                            savedInScopeVars = xpathContext.inScopeVars.copy()
                             try:
                                 evaluate(xpathContext, modelVariableSet)
                             except XPathContext.VariableSetRunTimeExceededException:
@@ -862,6 +863,7 @@ def validate(val, xpathContext=None, parametersOnly=False, statusMsg='', compile
                                     variableSet=varSetId,
                                     secs=varSetTimeout,
                                 )
+                                xpathContext.inScopeVars = savedInScopeVars
                             finally:
                                 if varSetTimeoutTimer:
                                     varSetTimeoutTimer.cancel()


docs/source/command_line.md
```diff
@@ -132,6 +132,8 @@ The following file options determine the type of file saved by the extension of
...
 | `--formulaCompileOnly` _or_<br/>`--formulacompileonly`                                             | Specify formula are to be compiled but not executed.                                                                                                                                                                                                                                                 |
 | `--formulaCacheSize` _or_<br/>`--formulacachesize`                                                 | Specify the number of fact aspect combinations to cache during formula evaluations. Negative numbers have no limit.                                                                                                                                                                                  |
+| `--formulaVarSetTimeout` _or_<br/>`--formulavarsettimeout`                                   | Maximum number of seconds per variable set evaluation before logging a timeout error.                                                                                                                                                                                                                 |
```
volodymyrtur-wk: looks like broken alignment of the table between a Flag and a Description

arelle/formula/ValidateFormula.py
```diff
@@ -843,7 +846,26 @@ def validate(val, xpathContext=None, parametersOnly=False, statusMsg='', compile
...
-                            evaluate(xpathContext, modelVariableSet)
+                            varSetTimeout = formulaOptions.formulaVarSetTimeout
+                            if varSetTimeout is not None:
+                                varSetTimeoutTimer = Timer(varSetTimeout, xpathContext.variableSetRunTimeExceededCallback)
+                                varSetTimeoutTimer.start()
+                            else:
+                                varSetTimeoutTimer = None
+                            try:
+                                evaluate(xpathContext, modelVariableSet)
+                            except XPathContext.VariableSetRunTimeExceededException:
```
stevenbronson-wk, line 858: What happens if something else uses this variable set?  I don't know this code, but I wonder if something might crash on partially initialized data, or see unexpected results if some items cached on XPathContext aren't cleared.  Compared to not terminating that seems fine, but people might not understand the timeout argument that way.

austinmatherne-wk: inScopeVars wasn't being reset, but the other state on XPathContext is reset by evaluate() at the top of each call. I put the timeout here because variable sets are the top level unit of evaluation (besides just aborting all formula processing), so snapshotting and restoring at this level largely avoids dealing with internal processing dependencies.

allenbross-wf:
QA +10
All remote and local testing passed.

PR #828, opened 2023-08-16
Reorganize argument parsing and CLI controller initialization

allenbross-wf:
#### Reason for change

Splitting apart the ParseAndRun functionality in CntlrCmdLine

#### Description of change

- Added some more info to SystemInfo.py
- Created the ArelleBaseOptions class
- Split out the parse from the run function.

#### Steps to Test

Smoketest with a combination of options and filings.

**review**:
@Arelle/arelle

PR #2268, opened 2026-03-23
Type hints/XT-5689

volodymyrtur-wk:
#### Steps to Test
CI
**review**:
@Arelle/arelle

### DIFF
--- a/arelle/Cntlr.py
+++ b/arelle/Cntlr.py
@@ -630,11 +630,11 @@ class Cntlr:
                     win32clipboard.OpenClipboard()
                     if text is None:
                         if win32clipboard.IsClipboardFormatAvailable(win32clipboard.CF_TEXT):
-                            return win32clipboard.GetClipboardData().decode("utf8")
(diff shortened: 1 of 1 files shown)


arelle/RuntimeOptions.py
```diff
@@ -49,8 +49,8 @@ class RuntimeOptions:
     diagnostics: Optional[bool] = None
     diffFile: Optional[str] = None
     dimFile: Optional[str] = None
-    disablePersistentConfig: Optional[bool] = None
-    disableRtl: Optional[bool] = None
+    disablePersistentConfig: bool = False
+    disableRtl: bool = False
```
aaroncameron-wk, line 53: @allenbross-wf Do you happen to remember why these were made optional initially? I couldn't find any related discussion in the [original PR](https://github.com/Arelle/Arelle/pull/828). I'm guessing it was just for consistency with the old argument parser yielding `None` values for these when not explicitly set. If that's the case then these changes are fine.

allenbross-wf: I do not remember.  The consistency guess makes a lot of sense, though.

Issue #2269, opened 2026-03-23
Breaking Change: Package Manager Class


aaroncameron-wk:
## Overview
As part of a strategic initiative to reduce the use of global state within Arelle, the `PackageManager` module is being refactored into an instantiable class. This architecture enables `PackageManager` instances to be explicitly owned and injected by `Cntlr` instances, facilitating improved isolation and concurrency. Legacy touchpoints, such as `packageInfo`, must now be invoked via the controller-provided instance (e.g., `modelXbrl.modelManager.cntlr.packageManager.packageInfo(...)`).

### Users Affected
- **Python API & Plugins:** Consumers interacting with module-level functions and static variables within the `PackageManager` namespace.
  - [ ] Provide migration instructions.

## Rollout Plan
In accordance with the [breaking change rollout lifecycle](https://arelle.readthedocs.io/en/latest/versioning.html), the transition is scheduled as follows:
- [x] Announcement (March 25, 2026)
- [ ] Phase 1: Opt-In
  - [ ] **Python API & Plugins:** The controller-resident `PackageProvider` instance becomes available. Module-level functions remain functional as a compatibility layer, with both patterns temporarily sharing a singleton state to ensure stability.
- [ ] Phase 2: Opt-Out / Deprecation
  - [ ] **Python API & Plugins:** Usage of module-level functions and variables will trigger formal deprecation warnings (`DeprecationWarning`).
- [ ] Phase 3: Removal
  - [ ] **Python API & Plugins:** Module-level functions and variables are decommissioned. Each `Cntlr` instance will utilize a dedicated, isolated `PackageManager` instance, finalized for multi-tenant environments.

PR #2271, opened 2026-03-24
`PackageManager` type hints

aaroncameron-wk:
#### Reason for change
Part of the larger type hint effort as well as a precursor to #2269 

#### Steps to Test
CI

**review**:
@Arelle/arelle

### DIFF
--- a/arelle/CntlrCmdLine.py
+++ b/arelle/CntlrCmdLine.py
@@ -2325,7 +2325,7 @@ class CntlrCmdLine(Cntlr.Cntlr):
         packageInfo = PackageManager.addPackage(self, package, packageManifestName)
         if packageInfo:
             self.addToLog(_("Activation of package {0} successful.").format(packageInfo.get("name")),
-                          messageCode="info", file=str(packageInfo.get("URL")))
+                          messageCode="info", file=packageInfo.get("URL", ""))
         else:
             self.addToLog(_("Unable to load package \"%(name)s\". "),
                           messageCode="arelle:packageLoadingError",
@@ -2353,7 +2353,7 @@ class CntlrCmdLine(Cntlr.Cntlr):
                 packageInfo = PackageManager.addPackage(self, cmd[1:], packageManifestName)
                 if packageInfo:
                     self.addToLog(_("Addition of package {0} successful.").format(packageInfo.get("name")),
-                                  messageCode="info", file=str(packageInfo.get("URL")))
+                                  messageCode="info", file=packageInfo.get("URL", ""))
                 else:
                     self.addToLog(_("Unable to load package."), messageCode="info", file=cmd[1:])
             elif cmd.startswith("~"):
--- a/arelle/PackageManager.py
+++ b/arelle/PackageManager.py
@@ -10,7 +10,7 @@ import time
 import zipfile
 from collections import defaultdict
 from fnmatch import fnmatch
-from typing import Any, IO, TYPE_CHECKING
+from typing import Any, IO, TYPE_CHECKING, cast
 from urllib.parse import urljoin
 
 from lxml import etree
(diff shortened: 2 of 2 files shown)


arelle/PackageManager.py
```diff
@@ -484,7 +498,7 @@ def validateTaxonomyPackage(
                 fileSource=filesource,
                 **validation.args
             )
-            errors.append(validation.codes)
+            errors.append(str(validation.codes))
```
austinmatherne-wk: Is this right? Seems like this should either be `.extends(...)` or `errors: list[str]` is wrong?

arelle/CntlrCmdLine.py
```diff
@@ -2322,10 +2322,10 @@ def showStatusOnPipe(self, message: str | None, clearAfter: int | None = None) -
 
     def loadPackage(self, package: str, packageManifestName: str) -> None:
         from arelle import PackageManager
-        packageInfo = PackageManager.addPackage(self, package, packageManifestName)  # type: ignore[no-untyped-call]
+        packageInfo = PackageManager.addPackage(self, package, packageManifestName)
         if packageInfo:
             self.addToLog(_("Activation of package {0} successful.").format(packageInfo.get("name")),
-                          messageCode="info", file=packageInfo.get("URL"))
+                          messageCode="info", file=str(packageInfo.get("URL")))
```
austinmatherne-wk: Use a blank default instead of wrapping?
```suggestion
                          messageCode="info", file=packageInfo.get("URL", ""))
```

volodymyrtur-wk: or add `# type: ignore[arg-type]`

arelle/PackageManager.py
```diff
@@ -48,27 +49,33 @@
...
     while baseElt is not None:
         if baseAttr := baseElt.get("{http://www.w3.org/XML/1998/namespace}base"):
             base = baseAttr if baseAttr.startswith("/") else baseAttr + base
         baseElt = baseElt.getparent()
     return base
 
-def xmlLang(element):
-    return (element.xpath('@xml:lang') + element.xpath('ancestor::*/@xml:lang') + [''])[0]
+def xmlLang(element: etree._Element) -> str:
+    return str((element.xpath('@xml:lang') + element.xpath('ancestor::*/@xml:lang') + [''])[0])
```
volodymyrtur-wk: You can use `f-string` instead of concatenating and casting

arelle/PackageManager.py
```diff
@@ -144,7 +151,7 @@ def _parsePackageMetadata(
         return pkg
 
     root = tree.getroot()
-    ns = root.tag.partition("}")[0][1:]
+    ns = str(root.tag).partition("}")[0][1:]
```
volodymyrtur-wk: `assert isinstance(root.tag, str)` will work too instead of casting or `cast(str, root.tag).partition("}")[0][1:]` might work as well

arelle/PackageManager.py
volodymyrtur-wk: In case `validation.codes` will be a tuple of strings, casting to str is not a good idea. I suggest updating the code to cover both options:
```
if isinstance(validation.codes, str):
    errors.append(validation.codes)
else:
    errors.extend(validation.codes)
```

arelle/PackageManager.py
aaroncameron-wk: It isn't concatenating strings, it's building a list of lang attribute values and then selecting the first value. I'll update to perform a type hint cast.

arelle/PackageManager.py
```diff
@@ -495,7 +499,10 @@ def validateTaxonomyPackage(
                 fileSource=filesource,
                 **validation.args
             )
-            errors.append(str(validation.codes))
+            if isinstance(validation.codes, str):
+                errors.append(validation.codes)
+            else:
+                errors.extend(validation.codes)
```
austinmatherne-wk, line 505: Doesn't need to be in this PR, but this is awkward enough that we may want to add a separate computed property on validations that wraps single string codes so there's a consistent `tuple[str, ...]` access pattern

aaroncameron-wk: Agreed, that's much cleaner than changing the type.

PR #2275, opened 2026-03-25
Fix No Entry Points in Archive Error Message

austinmatherne-wk:
#### Reason for change

Archives without XBRL entry points show a confusing `[IOerror]` instead of `No XBRL entry points could be loaded`. This was a regression introduced in #2023 when `filesourceEntrypointFiles` was changed to preserve `entrypointFiles` when no new entry points were discovered.

Master:
```sh
> python arelleCmdLine.py --file empty.zip
[IOerror] empty.zip: file error: [Errno 2] Archive /Users/austinmatherne/git/Arelle/empty.zip: '' - empty.zip
```
PR:
```sh
> python arelleCmdLine.py --file empty.zip
[error] No XBRL entry points could be loaded from provided file: empty.zip -
```

#### Description of change

The loadFromExcel plugin now uses the `FileSource.EntrypointFiles` hook to declare its entry points. This lets the archive block safely clear `entrypointFiles` unconditionally again, restoring the error message for archives that both can't be loaded by a plugin and contain no detectable entry points.

#### Steps to Test

* CI (`no_entrypoints_zip.py` and `load_from_excel.py` test scripts)

**review**:
@Arelle/arelle

### DIFF
--- a/tests/integration_tests/scripts/tests/no_entrypoints_zip.py
+++ b/tests/integration_tests/scripts/tests/no_entrypoints_zip.py
@@ -1,0 +1,55 @@
+from __future__ import annotations
+
+import os
+import zipfile
+from pathlib import Path
+
+import regex
+
+from tests.integration_tests.scripts.script_util import (
+    assert_result,
+    parse_args,
+    prepare_logfile,
+    run_arelle,
+    validate_log_file,
+)
+
+errors = []
+this_file = Path(__file__)
+args = parse_args(
+    this_file.stem,
+    "Verify error message when loading a zip file with no XBRL entry points.",
+)
+arelle_command = args.arelle
+arelle_offline = args.offline
+test_directory = Path(args.test_directory)
+arelle_log_file = prepare_logfile(test_directory, this_file)
+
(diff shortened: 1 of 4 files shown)


tests/integration_tests/scripts/tests/no_entrypoints_zip.py
```diff
@@ -0,0 +1,52 @@
...
+)
+arelle_command = args.arelle
+arelle_offline = args.offline
+test_directory = Path(args.test_directory)
+arelle_log_file = prepare_logfile(test_directory, this_file)
+
+no_entrypoints_zip_path = test_directory / 'no_entrypoints.zip'
+print(f"Creating zip file with no XBRL entry points: {no_entrypoints_zip_path}")
+with zipfile.ZipFile(no_entrypoints_zip_path, 'w') as zf:
+    zf.writestr('readme.md', 'This archive has no XBRL entry points.')
```
volodymyrtur-wk: To keep consistency, use one type of quote. I suggest sticking to double-quotes

Issue #2247, opened 2026-03-18
Breaking Change: Arelle Test Engine


aaroncameron-wk:
## Overview
We've published a new conformance suite test engine that we're evaluating for integration into Arelle proper. Among other benefits, the new engine allows parallel execution of test case variations. We'd love for folks to start experimenting with it and share feedback before we make the switch.

https://github.com/Arelle/arelle-test-engine/

When this releases, it will replace the current testcase processing flow. Effort will be made to minimize changes to user workflows, but adjustments may be necessary. These adjustments will be documented and migration instructions will be provided.

### Users Affected
- **Python API:** Users with code that interacts deeply with internal validation logic may be affected. A migration path may not be feasible.
- **CLI:** Users that validate testcases via the GUI will instead invoke the new test engine package, e.g. `python -m arelle_test_engine ...`.
  - [ ] Provide migration instructions.
- **GUI:** Users that validate testcases via the GUI may need to adjust workflows:
  - [ ] Provide migration instructions.
- **Web Server:** Users that validate testcases via the web server may need to adjust workflow. We are considering dropping support for conformance suite execution via web server:
  - [ ] Provide migration instructions.
- **Plugin Hooks:** Users of the following plugin hooks (which will be removed or repurposed) should plan on customizing their interaction with Arelle's testing system through other means. Plugin integration with the test engine itself will not be supported. 
  - `ModelTestcaseVariation.ArchiveIxds`
  - `ModelTestcaseVariation.ExpectedCount`
  - `ModelTestcaseVariation.ExpectedResult`
  - `ModelTestcaseVariation.ExpectedSeverity`
  - `ModelTestcaseVariation.ReadMeFirstUris`
  - `ModelTestcaseVariation.ReportPackageIxdsOptions`
  - `ModelTestcaseVariation.ReportPackageIxds`
  - `ModelTestcaseVariation.ResultXbrlInstanceUri`
  - `TestcaseVariation.Validated`
  - `TestcaseVariation.Xbrl.Loaded`
  - `TestcaseVariation.Xbrl.Validated`
  - `Testcases.Start`

## Rollout Plan
In accordance with our [breaking change rollout lifecycle](https://arelle.readthedocs.io/en/latest/versioning.html), the rollout is planned as follows:
- [x] Announcement (Feb. 25, 2026)
- [ ] Phase 1: Opt-In
[...]

PR #2281, opened 2026-03-27
Clean up XbrlConst.py to use module level frozenset as much as possible

strr:
#### Reason for change

`XbrlConst.py` has a number of collections that are either:

- not frozen sets (some tuples)
- sets created within a method every time the method is called rather than defined at the module level

#### Description of change

- Make sure all collections in XbrlConst.py are frozensets except a couple which seem to need to be tuples for now.
- Update various `XbrlConst.isBlah(uri: str)` methods to take an argument that is `str | None` to reduce the type-checking noise / caller-burden when using a namespaceUri / arcrole that is `str | None`

#### Steps to Test

Should be a straight re-factor ... no functional change.

**review**:
@Arelle/arelle

### DIFF
--- a/arelle/XbrlConst.py
+++ b/arelle/XbrlConst.py
@@ -367,7 +367,7 @@ enum2s = frozenset({
 })
 enum_1x = frozenset({
     "http://xbrl.org/2014/extensible-enumerations",
-    "http://www.xbrl.org/PWD/2016-10-12/extensible-enumerations-1.1",
+    "http://xbrl.org/PWD/2016-10-12/extensible-enumerations-1.1",
     "http://xbrl.org/WGWD/YYYY-MM-DD/extensible-enumerations-1.1",
 })
 enums = enum_1x | enum2s
@@ -739,7 +739,7 @@ standardNamespaceSchemaLocations: dict[str, str] = {
 }
 
 
-numericXsdType = frozenset({
+numericXsdTypes = frozenset({
         "integer",
         "positiveInteger",
         "negativeInteger",
@@ -758,7 +758,7 @@ numericXsdType = frozenset({
         "double",
     }
 )
-decimalXsdType = frozenset({
+decimalXsdTypes = frozenset({
         "integer",
         "positiveInteger",
         "negativeInteger",
@@ -793,11 +793,11 @@ integerXsdTypes = frozenset({
 )
 
 def isNumericXsdType(xsdType: str | None) -> bool:
-    return xsdType in numericXsdType
+    return xsdType in numericXsdTypes
 
 
 def isDecimalXsdType(xsdType: str | None) -> bool:
-    return xsdType in decimalXsdType
+    return xsdType in decimalXsdTypes
 
 
 def isIntegerXsdType(xsdType: str | None) -> bool:


arelle/XbrlConst.py
```diff
@@ -338,24 +337,24 @@ def qnIsoCurrency(token: str | None) -> QName | None:
 vercu = "http://xbrl.org/2013/versioning-concept-use"
 vercd = "http://xbrl.org/2013/versioning-concept-details"
 verdim = "http://xbrl.org/2013/versioning-dimensions"
-verPrefixNS = {
+verPrefixNS = frozenset({
```
austinmatherne-wk: This one was also a dict, not a mutable set

arelle/XbrlConst.py
```diff
@@ -895,30 +910,28 @@ def isNumericRole(role: str) -> bool:
 })
 
 
-standardDefinitionArcroles = frozenset(standardDimensionArcroles | {
+standardDefinitionArcroles = frozenset({
```
austinmatherne-wk: Removing dimension arcroles from this collection is likely to impact both [internal](https://github.com/Arelle/Arelle/blob/095e4dd05b0e280f984050021f6d0e80bc8bf8c4/arelle/LinkbaseType.py#L73) and external code that expects the current set of items.

arelle/ModelDocument.py
```diff
@@ -213,9 +213,10 @@ def load(modelXbrl, uri, base=None, referringElement=None, isEntry=False, isDisc
         if file:
             file.close()
         # retry in case of well known schema locations
-        if not isIncluded and namespace and namespace in XbrlConst.standardNamespaceSchemaLocations and uri != XbrlConst.standardNamespaceSchemaLocations[namespace]:
-            return load(modelXbrl, XbrlConst.standardNamespaceSchemaLocations[namespace],
-                        base, referringElement, isEntry, isDiscovered, isIncluded, namespace, reloadCache)
+        shouldLoad = namespace and (canonicalUri := XbrlConst.standardNamespaceSchemaLocations.get(namespace)) is not None and uri != canonicalUri
+        if not isIncluded and shouldLoad:
+            return load(modelXbrl, canonicalUri,
+                        base, referringElement, isEntry, isDiscovered, isIncluded, bool(namespace), reloadCache)
```
austinmatherne-wk: There's a preexisting bug here: There's an optional `isSupplemental` arg between `isIncluded` and `namespace` so the last two parameters are mapped to the wrong position in the function signature. Can you update to kwargs to fix this?

```python
            return load(
                modelXbrl=modelXbrl,
                uri=canonicalUri,
                base=base,
                referringElement=referringElement,
                isEntry=isEntry,
                isDiscovered=isDiscovered,
                isIncluded=isIncluded,
                namespace=namespace,
                reloadCache=reloadCache,
            )
```

arelle/XbrlConst.py
```diff
@@ -4,20 +4,17 @@
 from __future__ import annotations
 
 import os
-from typing import TYPE_CHECKING, Tuple, cast
+from typing import TYPE_CHECKING, cast
+from gettext import gettext as _
```
austinmatherne-wk: This introduces a subtle bug that [I've tripped over myself](https://github.com/Arelle/Arelle/commit/8dbcf00469db69fa0f6c6782b0c4251c7d84300b). Arelle uses `gettext.translation("arelle", ...).install()` which installs a domain specific translator into builtins. `gettext.gettext` looks up the global domain, not the `arelle` domain that was installed. So it will silently skip Arelle's translations.

arelle/XbrlConst.py
```diff
@@ -449,7 +444,7 @@ def qnIsoCurrency(token: str | None) -> QName | None:
 qnEqualityTestA = qname("{http://xbrl.org/2008/variable/aspectTest}aspectTest:a")
 qnEqualityTestB = qname("{http://xbrl.org/2008/variable/aspectTest}aspectTest:b")
 formula = "http://xbrl.org/2008/formula"
-tuple = "http://xbrl.org/2010/formula/tuple"
+formulaTuple = "http://xbrl.org/2010/formula/tuple"
```
austinmatherne-wk, line 464: `XbrlConst` is used by plugins outside of this repo and by codebases that integrate with Arelle as a library. While converting collections to frozensets and consolidating related checks into helper functions are improvements, renaming or removing public symbols (tuple > formulaTuple) are breaking interface changes that would require [a major version bump](https://arelle.readthedocs.io/en/latest/versioning.html). Not something I think we want to do just for this.

We're planning other breaking changes in the upcoming months (#2269, #2248, #2247). We could keep the original names as aliases alongside the new names with a deprecation comment, then later this year when we cut an Arelle 3.0 release with the other breaking changes we can also remove the deprecated aliases in `XbrlConst`.

The type changes (`set`/`tuple` > `frozenset`) are technically breaking changes, too. But I think those are minor enough that it's fine. If anyone is mutating these collections, I'd consider that a bug we're patching.

arelle/ModelDocument.py
strr: Ouch! Thanks, changed in [0f0c1e6](https://github.com/Arelle/Arelle/pull/2281/commits/0f0c1e6dda19f6c8bce6dfe289237b5a4cc0f902)

arelle/XbrlConst.py
strr: Thank you, I couldn't work out what was going on and the IDE type-checker doesn't like the current solution at all.

I've reverted in 1476565.

arelle/XbrlConst.py
strr: Reinstated this in 
1476565

but via __getattr__ so that I could add a DeprecationWarning

See what you think.

arelle/XbrlConst.py
strr: 1476565 reinstates standardDimensionArcroles and standardDefinitionArcroles but defined in terms of other things that are easier to reason about:

```python
standardDimensionArcroles = dimensionsSpecArcroles
standardDefinitionArcroles = baseSpecDefinitionArcroles | dimensionsSpecArcroles
```

arelle/XbrlConst.py
strr: With both, there was a certain snow blindness going on ... I've added the dict[str, str] type annotations so that they are easier to spot.

arelle/XbrlConst.py
```diff
@@ -337,24 +346,27 @@ def isDtrTypeNamespace(namespace: str | None) -> bool:
...
 # extended enumeration spec
 enum2s = frozenset({
     "http://xbrl.org/2020/extensible-enumerations-2.0",
     "http://xbrl.org/WGWD/YYYY-MM-DD/extensible-enumerations-2.0",
 })
-enums = frozenset({
+enum_1x = frozenset({
     "http://xbrl.org/2014/extensible-enumerations",
-    "http://xbrl.org/PWD/2016-10-12/extensible-enumerations-1.1",
+    "http://www.xbrl.org/PWD/2016-10-12/extensible-enumerations-1.1",
```
austinmatherne-wk: Was the `www` addition to the PWD namespace intentional?

arelle/XbrlConst.py
```diff
@@ -738,10 +757,8 @@ def isNumericXsdType(xsdType: str) -> bool:
         "float",
         "double",
     }
-
-
-def isDecimalXsdType(xsdType: str) -> bool:
-    return xsdType in {
+)
+decimalXsdType = frozenset({
```
austinmatherne-wk: nit pick: `decimalXsdType` and `numericXsdType` are singular, other vars (`integerXsdTypes`) in this module use plural names.

arelle/XbrlConst.py
strr: No! Good spot. I honestly have no idea how that snuck in.

strr: Fixed in 4d8277ed9748bb2721967dd40e975cc25a53946b

arelle/XbrlConst.py
strr: Fixed in 4d8277ed9748bb2721967dd40e975cc25a53946b

PR #2283, opened 2026-03-30
`PackageManager` class

aaroncameron-wk:
> [!IMPORTANT]
> Code movement will be easier to track if you review by commit *with whitespace hidden*.

#### Reason for change
This is part of the non-breaking first phase of #2269.
This will be followed by:
- Updating internal use of `PackageManager` to use a controller-managed instance rather than the module-level functions and variables.
- Refactoring of `PackageManager` class with clear separation of public and internal API.
- Addition of deprecation warnings with the eventual removal of the singleton instance and module-level variables and functions altogether.

#### Description of change
Introduces a `PackageManager` class. Temporarily defines a singleton instance of that class to maintain current functionality. Retains module-level functions and variables for temporary backward-compatability.

#### Steps to Test
CI

**review**:
@Arelle/arelle

### DIFF
--- a/arelle/PackageManager.py
+++ b/arelle/PackageManager.py
@@ -388,19 +388,18 @@ class PackageManager:
             except Exception:
                 pass # on GAE no userAppDir, will always come here
         if not self.packagesConfig:
-            self.packagesConfig = {  # savable/reloadable plug in configuration
+            self.packagesConfig = {  # savable/reloadable package configuration
                 "packages": [], # list taxonomy packages loaded and their remappings
                 "remappings": {}  # dict by prefix of remappings in effect
             }
-            self.packagesConfigChanged = False # don't save until something is added to pluginConfig
-        pluginMethodsForClasses: dict[str, Any] = {} # dict by class of list of ordered callable function objects
+            self.packagesConfigChanged = False # don't save until something is added to packagesConfig
         self._cntlr = cntlr
 
-    def reset(self) -> None:  # force reloading modules and plugin infos
+    def reset(self) -> None:
         if self.packagesConfig:
-            self.packagesConfig.clear()  # dict of loaded module pluginInfo objects by module names
+            self.packagesConfig.clear()
         if self.packagesMappings:
-            self.packagesMappings.clear() # dict by class of list of ordered callable function objects
+            self.packagesMappings.clear()
 
     def orderedPackagesConfig(self) -> dict[str, Any]:
         _cfg = self._getPackagesConfig()
@@ -445,7 +444,7 @@ class PackageManager:
                 f.write(jsonStr)
             self.packagesConfigChanged = False
 
-    def close(self) -> None:  # close all loaded methods
+    def close(self) -> None:
         self._getPackagesConfig().clear()
         self.packagesMappings.clear()
 


arelle/PackageManager.py
```diff
@@ -306,232 +874,50 @@ def _parseCatalog(
...
-    return packagesConfig
+def init(cntlr: Cntlr, loadPackagesConfig: bool = True) -> None:
+    return getInstance().init(cntlr, loadPackagesConfig)
 
 
-def _getCntlr() -> Cntlr:
-    assert _cntlr is not None, "PackageManager.init() must be called before use"
-    return _cntlr
+def reset() -> None:
+    return getInstance().reset()
```
aaroncameron-wk, line 891: I chose not to include any assertions or checks for initialization. This most accurately retains existing behavior (initialization checks did not exist).

volodymyrtur-wk (approved):

derekgengenbacher-wf (dismissed):

arelle/PackageManager.py
```diff
@@ -49,28 +49,771 @@
...
+_SINGLETON_ATTRS = frozenset({
+    "packagesJsonFile", "packagesConfig", "packagesConfigChanged",
+    "packagesMappings", "_cntlr",
+})
+
+
+def __getattr__(name: str) -> Any:
+    if name in _SINGLETON_ATTRS:
+        return getattr(_singleton, name)
+    raise AttributeError(f"module {__name__!r} has no attribute {name!r}")
```
austinmatherne-wk, line 803: This handles reads. Do we need `__setattr__` for backwards compatible reassignments? How are the [changes from the DialogPackageManager](https://github.com/Arelle/Arelle/blob/7e6b4b6faf38b42c2cb03ab65df2cd8cb4d142e6/arelle/DialogPackageManager.py#L265-L266) handled? Does that silently bypass the singleton?

arelle/CntlrCmdLine.py
```diff
@@ -2378,7 +2378,7 @@ def loadPackages(self, packages: list[str], packageManifestName: str) -> None:
         if savePackagesChanges:
             PackageManager.save(self)
         else:
-            PackageManager.packagesConfigChanged = False
+            PackageManager.getInstance().packagesConfigChanged = False
```
austinmatherne-wk: Should we update all the `PackageManager` operations (the others are only reads) in this method to use `getInstance()`?

arelle/PackageManager.py
```diff
@@ -49,28 +49,771 @@
...
+        if not self.packagesConfig:
+            self.packagesConfig = {  # savable/reloadable plug in configuration
+                "packages": [], # list taxonomy packages loaded and their remappings
+                "remappings": {}  # dict by prefix of remappings in effect
+            }
+            self.packagesConfigChanged = False # don't save until something is added to pluginConfig
+        pluginMethodsForClasses: dict[str, Any] = {} # dict by class of list of ordered callable function objects
+        self._cntlr = cntlr
+
+    def reset(self) -> None:  # force reloading modules and plugin infos
```
austinmatherne-wk: Preexisting and not an issue for this PR: it's unfortunate how many comments in this file were copied from `PluginManager` and don't make sense

arelle/PackageManager.py
aaroncameron-wk: I had changes to `DialogPackageManager` queued up on a future PR but you're correct to point out that as-is those reassignments would be broken. I'll update.

arelle/CntlrCmdLine.py
aaroncameron-wk: The next PR in this effort will swap all the uses to `Cntlr._packageManager`, but we might as well be consistent here for the interim.

PR #2284, opened 2026-03-31
Type hints/xt 5691

volodymyrtur-wk:
#### Steps to Test
CI
**review**:
@Arelle/arelle

### DIFF
--- a/.git-blame-ignore-revs
+++ b/.git-blame-ignore-revs
@@ -2,3 +2,13 @@
 956cfcca1aaa18dd2521172a851da3375abd2978
 # Fixed whitespace in arelle/plugin/validate/FERC/config.xml
 32529af727261eb1e9bf51a78280e558ef80f580
+# formatting changes in arelle/FunctionIxt.py
+f7a9df3d4d8971519ef0d9db3063dbddae459603
+# formatting changes in arelle/FunctionUtil.py
+d93240e5ae326f055971aad9b8fa08a36db47d9c
+# formatting changes in arelle/FunctionFn.py
+e5c51b9c3ccf4080589dd8713ea98b90a6ddfdbe
+# formatting changes in arelle/FunctionXs.py
+e4052d44491282927dd253928eb8a405b073822c
+# formatting changes in arelle/FunctionXfi.py
+43786571258444da879f76c9c165e76685689e4e


arelle/FunctionXfi.py
```diff
@@ -1039,51 +1594,80 @@ def concept_relationships(xc, p, args, nestResults=False, targetRole=False): # n
...
-        qnLink = qnameArg(xc, p, args, 5, 'QName', emptyFallback=None)
+        qnLink = qnameArg(xc, p, args, 5, "QName", emptyFallback=None)
     else:
         qnLink = None
-    if lenArgs > 5:
-        qnArc = qnameArg(xc, p, args, 6, 'QName', emptyFallback=None)
+    if lenArgs > 6:
+        qnArc = qnameArg(xc, p, args, 6, "QName", emptyFallback=None)
     else:
         qnArc = None
```
aaroncameron-wk, line 1660: This change does seem "correct" since it was probably the original intent, but this code is 15 years old, so it would appear to not be causing issues. I'm hesitant to make any functional changes in this area of the codebase since our team does not have a lot of familiarity with it. There's at least a little risk without much clear benefit other than the code looking more correct.

arelle/FunctionXfi.py
```diff
@@ -1092,116 +1676,193 @@ def concept_relationships(xc, p, args, nestResults=False, targetRole=False): # n
...
                     rels = relationshipSet.fromModelObject(srcConcept)
                     # get descendants
-                    if 0 <= generations <= 1:
+                    if generations == 0:
                         g = 0
                         visited = set()
                     else:
-                        g += 2
+                        # g is undefined, not sure if it is supposed to be generations variable
+                        g = generations
```
aaroncameron-wk, line 1729: Is there a reason this block gets `g = generations` while the similar block above gets `g = generations + 1`?

arelle/FunctionXfi.py
```diff
@@ -1092,116 +1676,193 @@ def concept_relationships(xc, p, args, nestResults=False, targetRole=False): # n
...
                     indexOfSelfInRoots = rootQNs.index(qnSource)
-                    if indexOfSelfInRels >= 0:
-                        concept_relationships_step(xc, inst, relationshipSet, [relToSelf], 'descendant', g, result, visited, nestResults, targetRole)
-                        if len(result) >= 2 and isinstance(result[1], list): # contains [self, [descendants...]
-                            if axis == 'sibling-or-descendant':
-                                return rootQNs[:indexOfSelfInRoots] + result[1] + rootQNs[indexOfSelfInRoots+1:]
-                            if axis == 'sibling-or-descendant-or-self':
-                                return rootQNs[:indexOfSelfInRoots] + result + rootQNs[indexOfSelfInRoots+1:]
+                    if indexOfSelfInRoots >= 0:
```
aaroncameron-wk, line 1731: This is a very complex and fairly obscure part of the codebase for me (and I think for the rest of the team as well). Any changes to behavior could have unintended side effects that I can't confidently predict without a deeper dive into this code.

Since the pre-existing reference to `indexOfSelfInRels` would have raised a `NameError` whenever it was executed, I would conclude that this code either:
- does not execute in practice.
- executes only in obscure scenarios that are not introduced by our integration tests or conformance suites, and is not disruptive enough for us to have heard about it.
- executes in common scenarios but is silenced or otherwise handled higher in the call chain.

So there isn't a clear necessity to fix this code. But there is a non-negligible risk that we fix this code in the wrong way so that when it does run (if it even can) it does so with more nuanced errors. While it feels ridiculous to say "if it ain't broke, don't fix it" about obviously broken code, I lean towards leaving it (and handling typing issues with ignore comments, if possible) or replacing it by raising an error explicitly since this scenario is clearly not supported by Arelle.

AustinMatherne: Some missing context here: Vova was understandably confused about this code and asked me about it. I initially told him to slap a type ignore on the errors and move on, but was curious and looked into it and provided him with a git patch with proper fixes.

This `nestResults` branch is only reachable via table linkbase rendering with a sibling-or-descendant axis targeting a root concept. Possible, but I'm not exactly surprised that a bug hasn't been reported for it.

arelle/FunctionXfi.py
AustinMatherne: This also came from me. The function signature is defined here: https://specifications.xbrl.org/registries/functions-registry-1.0/90507%20xfi.concept-relationships/90507%20xfi.concept-relationships%20function.html

arelle/FunctionXfi.py
AustinMatherne: The first block calls step with `[relToSelf]` (parent > self relationship), which consumes one generation before reaching actual descendants. The second block passes children directly to step.

arelle/FunctionIxt.py
```diff
@@ -305,44 +316,48 @@ def search(self, target):
...
-        elif len(arg) == 2:
-            return '20' + arg
-    return arg
+def z2(arg: str | None) -> str | None:   # zero pad to 2 digits
+    return arg.zfill(2) if isinstance(arg, str) else arg
+
 
-def yrin(arg, _mo, _day):   # zero pad to 4 digits
+def yr4(arg: str | None) -> str | None:   # zero pad to 4 digits
+    return "2" + arg.zfill(3) if isinstance(arg, str) and len(arg) <= 2 else arg
```
austinmatherne-wk: Both `z2` and `yr4` diverge from `master`'s behavior when given empty strings as input. Do you know if this functional change impacts anything?

arelle/FunctionIxt.py
```diff
@@ -1,6 +1,6 @@
-'''
+"""
 See COPYRIGHT.md for copyright information.
-'''
+"""
```
austinmatherne-wk, line 3: Can you add the git commit sha's for the formatting only commits to [.git-blame-ignore-revs](https://github.com/Arelle/Arelle/blob/master/.git-blame-ignore-revs)?

arelle/FunctionXfi.py
```diff
@@ -1142,12 +1142,12 @@ def concept_relationships(xc: XPathContext.XPathContext, p: OperationDef, args:
...
                     return [rel for rel in rels if rel.toModelObject != srcConcept]
                 # get descendants
-                if 0 <= generations <= 1:
+                if generations == 0:
                     g = 0
                     visited = set()
                 else:
                     # g is undefined, not sure if it is supposed to be generations variable
-                    g += 2  # type: ignore[used-before-def]
+                    g = generations + 1
```
austinmatherne-wk: This comment is now stale and should be removed.

arelle/FunctionXfi.py
```diff
@@ -1165,21 +1165,20 @@ def concept_relationships(xc: XPathContext.XPathContext, p: OperationDef, args:
                         return [c for c in rootQNs if c != qnSource]
                     rels = relationshipSet.fromModelObject(srcConcept)
                     # get descendants
-                    if 0 <= generations <= 1:
+                    if generations == 0:
                         g = 0
                         visited = set()
                     else:
                         # g is undefined, not sure if it is supposed to be generations variable
```
austinmatherne-wk: This comment is also stale.

arelle/FunctionXfi.py
```diff
@@ -1092,116 +1675,193 @@ def concept_relationships(xc, p, args, nestResults=False, targetRole=False): # n
...
                         return [c for c in rootQNs if c != qnSource]
                     rels = relationshipSet.fromModelObject(srcConcept)
                     # get descendants
-                    if 0 <= generations <= 1:
+                    if generations == 0:
                         g = 0
                         visited = set()
                     else:
-                        g += 2
+                        # g is undefined, not sure if it is supposed to be generations variable
```
austinmatherne-wk: The comment about g is now stale and should be removed.

arelle/FunctionIxt.py
volodymyrtur-wk: If we pass an empty string or something that can't be converted to an int as an input, it'll throw an error. It does not increase performance, only readability. I'll roll back changes

arelle/FunctionXfi.py
volodymyrtur-wk: yes, sure, I missed it when I cleaned the code

PR #2285, opened 2026-03-31
Implement public API for plugin interaction (`PluginProvider`)

aaroncameron-wk:
#### Reason for change
Part of the non-breaking phase of #2248.
Most of the functions and variables provided by `PluginManager` that are implicitly part of the public API are not necessary for plugins or Python API users.

#### Description of change
Introduce a new component of the plugin system: `PluginProvider`. This component serves as a public API wrapper for interacting with the internal plugin system.

#### Steps to Test
CI

**review**:
@Arelle/arelle

### DIFF
--- a/arelle/plugin_system/_plugin_manager.py
+++ b/arelle/plugin_system/_plugin_manager.py
@@ -46,6 +46,8 @@ class PluginManager:
 
     def __init__(self) -> None:
         self.pluginConfig: dict[str, Any] = {}
+        self.modulePluginInfos: dict[str, Any] = {}
+        self.pluginMethodsForClasses: dict[str, Any] = {}
         self._isInitialized = False
 
     @property
@@ -58,8 +60,8 @@ class PluginManager:
         self._cntlr: Cntlr = cntlr
         self.pluginConfigChanged = False
         self.pluginTraceFileLogger: logging.Logger | None = None
-        self.modulePluginInfos: dict[str, Any] = {}
-        self.pluginMethodsForClasses: dict[str, Any] = {}
+        self.modulePluginInfos = {}
+        self.pluginMethodsForClasses = {}
         self._pluginBase: str = cntlr.pluginDir + os.sep
         self._entryPointRefCache: list[EntryPointRef] | None = None
         self._entryPointRefAliasCache: dict[str, list[EntryPointRef]] | None = None


arelle/Cntlr.py
```diff
@@ -291,8 +292,9 @@ def __init__(
 
         # start plug in server (requres web cache initialized, but not logger)
         from arelle.PluginManager import getInstance as _getPluginManagerInstance
-        self.pluginManager = _getPluginManagerInstance()
-        self.pluginManager.init(self, loadPluginConfig=hasGui)
+        self._pluginManager = _getPluginManagerInstance()
```
austinmatherne-wk, line 295: The public `pluginManager` property type is [declared on the class above](https://github.com/Arelle/Arelle/blob/7e6b4b6faf38b42c2cb03ab65df2cd8cb4d142e6/arelle/Cntlr.py#L151)

arelle/logging/handlers/LogToXmlHandler.py
```diff
@@ -193,7 +193,7 @@ def getHtml(self, clearLogBuffer: bool = True) -> str:
         if self.logRecordBuffer:
             filters = None
             if self.cntlr is not None:
-                filters = self.cntlr.pluginManager.pluginClassMethods("Cntlr.Log.RecFilter.Html")
+                filters = self.cntlr.plugins.hooks("Cntlr.Log.RecFilter.Html")
             for logRec in self.logRecordBuffer:
                 if not filters or all(_filter(logRec) for _filter in filters):
```
austinmatherne-wk, line 198: Missed this in the previous PR, but `filters` is a generator of functions that is being reused within the `logRec` loop. It will be exhausted and then return `True` for subsequent `logRec`s because it's exhausted/empty

arelle/plugin_system/plugin_provider.py
```diff
@@ -0,0 +1,46 @@
...
+    def __init__(self, plugin_manager: PluginManager) -> None:
+        self._plugin_manager = plugin_manager
+
+    @staticmethod
+    def _parse_module_info(module_info: dict[str, Any]) -> PluginHandle:
+        return PluginHandle(
+            aliases=frozenset(module_info.get("aliases", set())),
+            author=str(module_info.get("author", "")),
+            description=str(module_info.get("description", "")),
+            entry_point=dict(module_info.get("entryPoint", {})).copy(),
```
austinmatherne-wk: This is creating a copy of a copy: `dict(...).copy()`

arelle/api/Session.py
```diff
@@ -130,7 +130,7 @@ def run(
             self._check_thread()
             PackageManager.reset()
             if self._cntlr is not None:
-                self._cntlr.pluginManager.reset()
+                PluginManager.getInstance().reset()
```
austinmatherne-wk: The `self._cntlr is not None` guard no longer makes sense here and in `close`?

PR #2286, opened 2026-04-01
Table Constraints Metadata Validation

### LINKED ISSUE #2197: TC Metadata: Value Constraint Type & Applicability Validation

Validate metadata time constraints on value constraint objects.

Error codes: `tcme:unknownType`, `tcme:illegalConstraint`, `tcme:unknownPeriodType`, `tcme:unknownDurationType`.

austinmatherne-wk:
#### Reason for change
More Table Constraints metadata validation.

#### Description of change
- Add `OimReport` class to consolidate OIM properties previously scattered on `ModelXbrl`
- Move TC metadata parsing from load time into the validation flow
- Implement remaining `tcme:invalidJSONStructure` error
- Add `TCMetadataValidator` with validation for:
  - `tcme:misplacedOrUnknownProperty`
  - `tcme:columnParameterConflict`
  - `tcme:invalidNamespacePrefix`
  - `tcme:unknownType`, `tcme:illegalConstraint`
  - `tcme:unknownPeriodType`, `tcme:unknownDurationType`
- Fix false positive `oimce:invalidSQName`
- Fix exception on single column xBRL-CSV tables during dialect sniffing

#### Steps to Test
* CI (conformance suites and unit tests)

**review**:
@Arelle/arelle


aaroncameron-wk (approved):

arelle/oim/_tc/metadata/common.py
```diff
@@ -0,0 +1,26 @@
...
+        self.path_segments: list[str] = list(path)
+        self._message = message
+        super().__init__(code)
+
+    def prepend_path(self, *segments: str) -> None:
+        self.path_segments = [*segments, *self.path_segments]
+
+    @property
+    def json_pointer(self) -> str:
+        return "/" + "/".join(self.path_segments)
```
brettkail-wk, line 21: Pre-existing, do we need `~0`/`~1` escaping?

arelle/oim/_model.py
```diff
@@ -0,0 +1,29 @@
...
+        document_info = self.oim_object.get("documentInfo", {})
+        return document_info if isinstance(document_info, dict) else {}
+
+    @cached_property
+    def namespaces(self) -> dict[str, str]:
+        namespaces = self.document_info.get("namespaces", {})
+        return {
+            prefix: namespace
+            for prefix, namespace in namespaces.items()
+            if isinstance(namespace, str) and isinstance(prefix, str)
```
brettkail-wk, line 28: It seems odd to create an instance of an object named `OimReport` with the "raw"/unparsed object.  It seems like the `document_info`/`oim_object` are more of a helper for parsing OIM reports rather than "the" OIM report.  Do we somewhere need to give an error if `documentInfo` or `namespaces` aren't actually objects, or if the value isn't actually a string?

It also seems odd to have a single structure with the unparsed object (input) and `loading_errors` (output).

It also seems odd to store the unparsed object on `ModelXbrl`.  That seems like the same kind of problem we have with other structures, where we allow access to "raw" lxml (input) objects rather than just the model objects.  Does it also prevent all the input object memory from being GC'ed after converting to model objects?

arelle/oim/_tc/metadata/value_constraint_validation.py
```diff
@@ -0,0 +1,338 @@
...
+        return False
+    prefix, sep, local_name = type_value.partition(":")
+    if sep:
+        return prefix == "xs" and local_name in _ALLOWED_XS_TYPES
+    return type_value in _CORE_DIMENSION_AND_DECIMALS_TYPES
+
+
+def resolve_effective_type(constraint_type: str, namespaces: dict[str, str]) -> QName | None:
+    if constraint_type in _CORE_DIMENSION_EFFECTIVE_TYPES:
+        return _CORE_DIMENSION_EFFECTIVE_TYPES[constraint_type]
```
volodymyrtur-wk, line 317: another way to write this condition:
```
if core_dimension_effective_type := _CORE_DIMENSION_EFFECTIVE_TYPES.get(constraint_type):
    return core_dimension_effective_type

volodymyrtur-wk (approved):

PR #2288, opened 2026-04-02
Publish Docker images to Docker Hub

ryangeiser1:
#### Reason for change
Publish Docker images to Docker Hub upon release.
Add images containing all plugins and a slim image.

#### Description of change
This change creates a new Github workflow to publish updated and tagged images upon new releases. My test repository can be found [here](https://hub.docker.com/r/ryangeiser1/arelle/tags)
- Tags on new base images will include latest, 2, 2.39, 2.39.6
- Tags on new slim images will include slim, 2-slim, 2.39-slim, 2.39.6-slim

#### Steps to Test
1. Generate Docker Hub token with read/write access 
   - Docker Hub > Profile (upper right corner) > Account settings > Personal access tokens (left menu bar) > Generate new token
   - Description: GitHub Actions
   - Expiration: ...
   - Access permissions: Read & Write
2. Add 2 environment secrets to "release" environment
   - DOCKER_USERNAME=arelleproject
   - DOCKER_TOKEN=arelleproject_docker_hub_token_12345 (created in step 1)
3. Actions > Publish Docker images > Run workflow > Tag:2.39.6 > Run workflow
   - Will need a new release because these changes aren't in 2.39.6 
4. `docker run --name arelle-webserver -p 8080:8080 ryangeiser1/arelle:2.39.6 /opt/start.sh`
5. `docker run --name arelle-slim-webserver -p 8080:8080 ryangeiser1/arelle:2.39.6-slim /opt/start.sh`
6. `docker run --name arelle-webserver -p 8080:8080 ghcr.io/ryangeiser1/arelle:2.39.6 /opt/start.sh`
7. `docker run --name arelle-slim-webserver -p 8080:8080 ghcr.io/ryangeiser1/arelle:2.39.6-slim /opt/start.sh`
8. https://github.com/ryangeiser1/Arelle/pkgs/container/arelle

#### TODO
- Add Docker Hub token to secrets
- Can we combine the 2x build-push and attestation steps into one step each?
- Return artifacts to release report

**review**:
@Arelle/arelle


.github/workflows/publish-docker.yml
```diff
@@ -0,0 +1,137 @@
...
+    permissions:
+      packages: write
+      contents: read
+      attestations: write
+      id-token: write
+    steps:
+      - name: Checkout Arelle
+        uses: actions/checkout@v6.0.2
+        with: 
+          ref: ${{ inputs.arelle_ref }}
```
austinmatherne-wk: I think there's a subtle race condition here. When `release.yml` calls `publish-docker.yml`, it passes no inputs. `arelle_ref` defaults to `master`, so the checkout step explicitly checks out master HEAD not the release tag. If any commit lands on master between tagging and the Docker build (if say a PR is merged after a release is tagged, but while tests are still running), the image will contain unreleased code while being labeled with the release version.

The other release workflows (`build-linux.yml`, `tests.yml`, etc.) don't have this problem because they use `actions/checkout` without specifying a ref, which defaults to `GITHUB_REF` (the release tag). The interface on GitHub for triggering workflow dispatch runs allows selecting the branch/tag independently of inputs, so I don't think there's a need to define/include `arelle_ref`.

docker/Dockerfile
```diff
@@ -63,16 +60,10 @@ COPY --from=builder /opt/venv /opt/venv
...
+COPY --chmod=755 docker/scripts/start.sh /opt/start.sh
+COPY --chmod=755 docker/scripts/healthcheck.sh /opt/healthcheck.sh
 
 WORKDIR /usr/src/app
 
-## Add user with home directory (Arelle caches to the user's home directory by default)
-## Assign app ownership and switch to new user
-RUN useradd -m arelle \
-  && chown -R arelle:arelle /usr/src/app
-USER arelle
```
austinmatherne-wk, line 76: Was this removed because it was causing issues? It's best to avoid root users if possible. I suspect the `chown` probably created a large layer, but I also can't think of a reason why the user would need to own the app directory.

docs/source/install.md
```diff
@@ -98,47 +98,20 @@ arelleGUI
...
-### Build
-```shell
-docker build -t arelle-webserver -f docker/Dockerfile .
-
-# With the EDGAR plugins
-docker build --build-arg INCLUDE_EDGAR="true" --build-arg EXTRA_PIP="-r requirements-plugins.txt" -t arelle-edgar-webserver -f docker/Dockerfile .
-```
+Run Arelle with Docker. New releases are tagged and published to [Docker Hub][docker-hub]. There are a couple different configurations/tags:
+- latest - Arelle with all plugins ([EDGAR plugins][edgar-plugin], [XULE plugin][xule-plugin], [iXBRL Viewer plugin][ixbrl-viewer-plugin], and [EFM][efm-plugin] disclosure systems)
+- slim - Arelle with no plugins
```
austinmatherne-wk: The slim image still includes all the plugins directly in this repo (under [arelle/plugin](https://github.com/Arelle/Arelle/tree/master/arelle/plugin)). Suggest:

```suggestion
- slim - Arelle without EDGAR, XULE, or iXBRL Viewer plugins
```

austinmatherne-wk (commented):
Thank you! I'll get this tested as soon as the secrets are added to the repo.

.github/workflows/publish-docker.yml
ryangeiser1: Good catch!

docker/Dockerfile
ryangeiser1: I think anyone using these containers in a production environment will install additional dependencies (WSGI server, custom plugins, custom start.sh scripts, etc...). In a custom Dockerfile, we have to switch back to the root user, install the packages, and then switch back to the arelle user. I think anyone deploying to a production environment will know to use a non-root user. 

Looking at it again, it only adds one line to a custom Dockerfile so may be best to leave the arelle user in the published image. 

```shell
# custom.Dockerfile
FROM arelleproject/arelle:2.39.5

WORKDIR /usr/src/app

## Creating arelle user in published image requires this line
# USER root

ENV PYTHONUNBUFFERED=1
ENV PATH="/opt/venv/bin:$PATH"
RUN pip install gunicorn

COPY --chmod=755 start.sh /opt/

## Add user with home directory (Arelle caches to the user's home directory by default)
## Assign app ownership and switch to new user
RUN useradd -m arelle \
  && chown -R arelle:arelle /usr/src/app
USER arelle

ENTRYPOINT ["/opt/entrypoint.sh"]

```
Non-root users can also be specified in docker-compose.yml, though I'm not 100% sure where docker compose creates the user. If docker compose creates/switches the user before `FROM python:3.14.2 AS builder`, this would not work. If it switches after the last line of the Dockerfile, this shouldn't affect the build. This should not affect a custom docker-compose file. The below should work, though I've always handled non-root users in the custom.Dockerfile. 

```shell
# docker-compose.custom.yml
services:
  arelle-webserver
    image: arelleproject/arelle:2.39.5
    user: arelle:arelle
    command: /opt/start.sh
    volumes:
      - /localdata:/containerdata
```

Regarding the non-root user owning the app, this should only be necessary if Arelle writes to any of the files inside the app (updating ruleset maps, etc...). If Arelle does not need to write inside the app, we can probably remove the chown.

austinmatherne-wk (approved):

austinmatherne-wk:
@ryangeiser1 thanks for doing this. I was able to test and confirm the workflows are working.

ryangeiser1:
Awesome, thank you @austinmatherne-wk! Which assets, if any, do you want to return to the release report? The GHCR image is linked under "Packages" on the repository index but not in the report.

austinmatherne-wk:
> Which assets, if any, do you want to return to the release report?

I don't think there's anything from this workflow that makes sense to include there.

PR #2295, opened 2026-04-07
CntlrWinMain: Add type hints and clean up type ignore comments

volodymyrtur-wk:
#### Description of change
Type Hints: arelle.CntlrWinMain

#### Steps to Test
CI
**review**:
@Arelle/arelle

PR #2296, opened 2026-04-07
Move GUI specific code out of XbrlConst.py and into ViewUtil.py

strr:
#### Reason for change
Clean-up

#### Description of change
Should just be a refactor, no functional change.

#### Steps to Test

If you have vermin (pip install vermin) installed:
```
vermin --target=3.10 --eval-annotations --backport dataclasses --backport enum --backport typing --backport typing_extensions arelle/
```
it should say:
```
Minimum required versions: 3.10
Incompatible versions:     2.x
```

**review**:
@Arelle/arelle


arelle/CntlrWinMain.py
```diff
@@ -1600,7 +1600,10 @@ def helpAbout(self, event=None):
     # worker threads addToLog
     def addToLog(self, message, messageCode="", messageArgs=None, file="", refs=[], level=logging.INFO):
         if isinstance(level, str):
-            level = logging.getLevelNamesMapping().get(level, logging.INFO)
+            try:
+                level = logging.getLevelNamesMapping().get(level, logging.INFO) # novermin
+            except AttributeError:
+                level = logging._nameToLevel.get(level, logging.INFO)  # private but 3.10-safe
```
austinmatherne-wk: This module [recently added type checking](https://github.com/Arelle/Arelle/pull/2295). If merged with the latest changes from master, we get a type checking error here. mypy doesn't track that `level` can't have been reassigned to `int` when `getLevelNamesMapping()` raises `AttributeError` (the assignment never completed, so `level` is still `str`).

```sh
> mypy arelle tests --namespace-packages
arelle/CntlrWinMain.py:1624: error: Argument 1 to "get" of "dict" has incompatible type "int | str"; expected "str"  [arg-type]
Found 1 error in 1 file (checked 538 source files)
```

Suggest capturing the narrowed `str` value before the try block so both branches use it directly:
```suggestion
            levelStr = level
            try:
                level = logging.getLevelNamesMapping().get(levelStr, logging.INFO) # novermin
            except AttributeError:
                level = logging._nameToLevel.get(levelStr, logging.INFO)  # private but 3.10-safe
```

austinmatherne-wk (approved):

PR #2298, opened 2026-04-07
Consolidation isExtension checks

allenbross-wf:
#### Description of change

Consolidation isExtension checks

#### Steps to Test

CI

**review**:
@Arelle/arelle

### DIFF
--- a/arelle/plugin/validate/ESEF/Dimensions.py
+++ b/arelle/plugin/validate/ESEF/Dimensions.py
@@ -14,7 +14,8 @@ from arelle.ModelObject import ModelObject
 from arelle.PrototypeDtsObject import PrototypeObject
 from arelle.ValidateXbrl import ValidateXbrl
 from arelle.typing import TypeGetText
-from .Util import isExtension, isInEsefTaxonomy
+from arelle.utils.validate.Common import isExtensionUri
+from .Util import isInEsefTaxonomy
 
 _: TypeGetText  # Handle gettext
 
@@ -25,6 +26,7 @@ def checkFilingDimensions(
     lineItemsNotQualifiedLinkroles: tuple[str, ...],
 ) -> None:
 
+    standardTaxonomyUriPrefixes = cast(frozenset[str], val.authParam["standardTaxonomyURIs"])
     val.primaryItems = set() # concepts which are line items (should not also be dimension members
     val.domainMembers = set()  # concepts which are dimension domain members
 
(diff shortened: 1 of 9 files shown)


arelle/plugin/validate/ESEF/Dimensions.py
```diff
@@ -52,18 +54,18 @@ def addDomMbrs(sourceDomMbr: ModelConcept, ELR: str, membersSet: set[ModelConcep
                 val.primaryItems.update(hcPrimaryItems)
                 hc = hasHcRel.toModelObject
                 if hasHypercubeArcrole == XbrlConst.all:
-                    if not hasHcRel.isClosed and isExtension(val, hasHcRel):
+                    if not hasHcRel.isClosed and isExtensionUri(hasHcRel.modelDocument.uri, val.modelXbrl, standardTaxonomyUriPrefixes):
```
aaroncameron-wk, line 57: Having `isExtensionUri` be a common utility that makes our extension URI classification consistent across the codebase is the main value-add to me. That value is still realized if a plugin implements a wrapper like `isExtension` around that.

I'm open to discussion, but I'm not sure needing to duplicate `standardTaxonomyUriPrefixes = cast(frozenset[str], val.authParam["standardTaxonomyURIs"])` any time this plugin needs to check for an extension URI, as well as this lengthier syntax, is an improvement over the current state. I think refactoring to have `Util.isExtensionUri`, `Util.isExtensionDoc`, and `Util.isExtensionObject` all ultimately making use of `Common.isExtensionUri` would be a better end result.

Alternatively, we could look into standardizing how "standard taxonomy URI prefixes" are configured so they could be made available to `Common.isExtensionUri` without the caller needing to compile them.

arelle/plugin/validate/ESEF/Util.py
```diff
@@ -38,6 +39,33 @@
...
+        setattr(val, ESEF_STANDARD_TAXONOMY_URI_PREFIXES_ATTR, taxonomyPrefixes)
+    return taxonomyPrefixes
+
+
+def isEsefExtensionUri(val: ValidateXbrl, uri: str) -> bool:
+    return isExtensionUri(uri, val.modelXbrl, standardTaxonomyUriPrefixes(val))
+
+
+def isExtensionDoc(val: ValidateXbrl, modelDocument: ModelDocument) -> bool:
+    return isExtensionUri(modelDocument.uri, val.modelXbrl, standardTaxonomyUriPrefixes(val))
```
aaroncameron-wk: Nit:
```suggestion
    return isEsefExtensionUri(val, modelDocument.uri)
```

Similar in `isExtensionObject` below.

aaroncameron-wk (dismissed):

arelle/plugin/validate/ESEF/Util.py
```diff
@@ -38,6 +39,33 @@
     "UKFRC", "UKFRC-2022", "UKFRC-2023",
 })
 
+ESEF_STANDARD_TAXONOMY_URI_PREFIXES_ATTR = "_esefStandardTaxonomyUriPrefixes"
+
+
+def standardTaxonomyUriPrefixes(val: ValidateXbrl) -> frozenset[str]:
+    taxonomyPrefixes: frozenset[str] | None = getattr(val, ESEF_STANDARD_TAXONOMY_URI_PREFIXES_ATTR, None)
+    if taxonomyPrefixes is None:
+        taxonomyPrefixes = cast(frozenset[str], val.authParam["standardTaxonomyURIs"])
```
austinmatherne-wk: This is fine, but the value is a `set[str]`, not `frozenset[str]`. Would be better to just use a real `frozenset`:
```suggestion
        taxonomyPrefixes = frozenset(val.authParam["standardTaxonomyURIs"])
```

austinmatherne-wk (dismissed):

stevenbronson-wk (approved):

austinmatherne-wk (approved):

PR #2306, opened 2026-04-14
Fix a few bugs in the Windows installer build script and enhance it to set installation and version metadata

strr:
#### Reason for change

- If you have `winget`, Arelle is found but upgrade is not offered without specifying --include-unknown because the current installed version is unknown (not set by the installer in the Windows Registry as DisplayVersion)
  - The winget manifests that have been contributed ad-hoc are [here](https://github.com/microsoft/winget-pkgs/commits/master/manifests/a/Arelle/Arelle)

```
PS C:\> winget upgrade --include-unknown
Name   Id            Version Available Source
---------------------------------------------
Arelle Arelle.Arelle Unknown 2.39.5    winget
1 upgrades available.
```
- The installer exe icon is the old nullsoft one not the Arelle icon
- The Arelle icon on modern Windows is heavily pixelated
- The license text displayed during install has mojibake `Â`
- The installer is compressed using gzip/deflate

#### Description of change
- Set consistent version metadata on the installer exe metadata and in the Windows registry written during installation
- Use the newer incantations to set the installer icon to `arelle.ico`
- Regenerate `arelle.ico` using `arelle.icns` (the MAC OS X icons) as the source
- Build time convert LICENCE.md to LICENCE.rtf using `pandoc` 
- Switch the installer to LZMA compression (looks to be about a 75 MiB download instead of 208 MiB)
- Drop the unused and somewhat broken `installWin86.nsi`

#### Steps to Test
- Following an install with the new installer, Settings -> Apps -> Installed Apps should display more metadata for Arelle including the version, a current installation date, a non-pixellated icon and the estimated disk usage
- Arelle icon in the start menu and in the corner of the window should be the newer winged-a (rather than the wing with arelle written underneath)
- You can play with the registry to set DisplayVersion and see winget upgrade's behaviour change:
```pwsh
Set-ItemProperty -Path "HKLM:\SOFTWARE\WOW6432Node\Microsoft\Windows\CurrentVersion\Uninstall\Arelle" -Name "DisplayVersion" -Value "1.23.4"
```
and
```pwsh
Remove-ItemProperty -Path "HKLM:\SOFTWARE\WOW6432Node\Microsoft\Windows\CurrentVersion\Uninstall\Arelle" -Name "DisplayVersion"
```

**review**:
@Arelle/arelle


austinmatherne-wk (approved):

austinmatherne-wk:
@strr thank you for the improvements!

PR #2307, opened 2026-04-15
Type Hints: arelle.ModelDocument

volodymyrtur-wk:
#### Description of change
- arelle.ModelDocument
- arelle.ModelDocumentType

These modules are excluded from mypy type checking. We should remove them from the exclusion list and add types so it passes mypy checks.

#### Steps to Test
CI
**review**:
@Arelle/arelle

### DIFF
--- a/arelle/ModelDocument.py
+++ b/arelle/ModelDocument.py
@@ -8,7 +8,9 @@ from zipfile import ZipFile
 import regex as re
 from collections import defaultdict
 from pathlib import Path
-from typing import Any, cast, Self, TYPE_CHECKING
+from typing import Any, cast, TYPE_CHECKING
+# for python 3.10
+from typing_extensions import Self
 from lxml import etree
 from xml.sax import SAXParseException
 
@@ -751,7 +753,7 @@ class ModelDocument(ModelDocumentBase):
 
     def setTitleInBackground(self) -> None:
         try:
-            cntlr = cast(CntlrWinMain, self.modelXbrl.modelManager.cntlr)
+            cntlr = cast("CntlrWinMain", self.modelXbrl.modelManager.cntlr)
             uiThreadQueue = cntlr.uiThreadQueue
             uiThreadQueue.put((self.setTitle, [cntlr]))
         except AttributeError:
@@ -762,7 +764,7 @@ class ModelDocument(ModelDocumentBase):
         updateFileHistory = getattr(myCntlr, 'updateFileHistory', None)
         if updateFileHistory:
             try:
-                cntlr = cast(CntlrWinMain, self.modelXbrl.modelManager.cntlr)
+                cntlr = cast("CntlrWinMain", self.modelXbrl.modelManager.cntlr)
                 uiThreadQueue = cntlr.uiThreadQueue
                 uiThreadQueue.put((updateFileHistory, [self.filepath, False]))
             except AttributeError:
@@ -1763,8 +1765,8 @@ def inlineIxdsDiscover(modelXbrl: ModelXbrl, modelIxdsDocument: ModelDocument, s
                 modelXbrl.error(ixMsgCode("tupleMemberOrderMissing", modelFact, sect="validation"),
                                 _("Inline XBRL tuple member %(qname)s must have a numeric order attribute"),
                                 modelObject=modelFact, qname=modelFact.qname)
-            if modelFact.get("target") == tuple.get("target") and hasattr(modelFact, "_ixFactParent"):
-                modelFact._ixFactParent = tuple # support ModelInlineFact parentElement()
+            if modelFact.get("target") == tuple.get("target"):
+                modelFact._ixFactParent = tuple  # type: ignore[attr-defined] # support ModelInlineFact parentElement()
             else:
                 modelXbrl.error(ixMsgCode("tupleMemberDifferentTarget", modelFact, sect="validation"),
                                 _("Inline XBRL tuple member %(qname)s must have a tuple parent %(tuple)s with same target"),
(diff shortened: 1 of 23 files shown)


derekgengenbacher-wf (dismissed):

arelle/plugin/validate/NL/rules/fg_nl.py
```diff
@@ -54,15 +54,19 @@ def rule_fg_nl_03(
     for modelDocument in val.modelXbrl.urlDocs.values():
         if modelDocument.type == ModelDocumentType.INSTANCE:
             continue
+
+        assert isinstance(modelDocument.targetXbrlElementTree, ElementTree)
```
austinmatherne-wk, line 58: Does this throw an exception on lxml versions older than 6 (`Element` and `ElementTree` were previously factory functions, not classes)? We [currently support](https://github.com/Arelle/Arelle/blob/f2c91b51a40bc04d09b43b569f2128786efc9121/pyproject.toml#L37) versions going back to 4

arelle/plugin/validate/NL/rules/fr_kvk.py
```diff
@@ -125,7 +125,7 @@ def rule_fr_kvk_2_03(
             for refDoc, docRef in doc.referencesDocument.items():
                 if "href" not in docRef.referenceTypes:
                     continue
-                if docRef.referringModelObject.localName != "schemaRef":
+                if docRef.referringModelObject is not None and docRef.referringModelObject.localName != "schemaRef":
```
austinmatherne-wk, line 128: Fixing a preexiting issue, but the new code looks like it should be `is None or`

arelle/ModelDocument.py
```diff
@@ -949,7 +967,7 @@ def schemaDiscover(self, rootElement, isIncluded, isSupplemental, namespace):
         else:
             if isIncluded == True and namespace:
```
austinmatherne-wk: preexisting but related to explicit bool type change:
```suggestion
            if isIncluded and namespace:
```

arelle/plugin/validate/NL/rules/fg_nl.py
volodymyrtur-wk: they are type aliases for `_Element` and `_ElementTree` in version 6. I can't even install something less than version 6 for Python 3.10, even in a clean environment. What do we use for package requirements `pyproject.toml`or `requirements.txt`?

arelle/plugin/validate/NL/rules/fr_kvk.py
volodymyrtur-wk: This extra condition for mypy to show that referringModelObject is not none and referringModelObject has an attribute localName

austinmatherne-wk: Understood. I was suggesting that the error below shouldn't be raised for the None case:

```suggestion
                if docRef.referringModelObject is None or docRef.referringModelObject.localName != "schemaRef":
```

arelle/plugin/validate/NL/rules/fg_nl.py
austinmatherne-wk: We use both for different purposes. Arelle is an application and a Python library. `pyproject.toml` is used by the python package (it supports dependency ranges to provide users who embed Arelle within their own Python applications with more flexibility). `requirements*.txt` is used for the application builds (and CI/local dev) where exact versions provide a more consistent and reproducible experience.

volodymyrtur-wk: I managed to install the 4th version of lxml and did some tests. It looks like `Element` and `_Element` are the same. When creating an instance of `Element`, it has the same attributes and methods as `_Element`, and if I make a check via the `isinstance` function, it returns `True`. Before initialization, the `Element` has type `<class 'cython_function_or_method'>`, after `<class 'lxml.etree._Element'>`. So I think we can use `Element` instead of `_Element` in the code, and it should work fine. Same for `ElementTree`

austinmatherne-wk: The values returned by `etree.ElementTree`/`etree.Element` are classes, but `etree.ElementTree` and `etree.Element` are functions not class definitions, so they can't be used as the second argument to `isintance`:

For example, with lxml 5.4.0:

```sh
lxml version: (5, 4, 0, 0)
type(etree.Element)     = cython_function_or_method
type(etree.ElementTree) = cython_function_or_method

isinstance against etree.Element / etree.ElementTree:
  isinstance(elem, etree.Element)     -> TypeError: isinstance() arg 2 must be a type, a tuple of types, or a union
  isinstance(tree, etree.ElementTree) -> TypeError: isinstance() arg 2 must be a type, a tuple of types, or a union
```

```python
from lxml import etree

print(f"lxml version: {etree.LXML_VERSION}")

print(f"type(etree.Element)     = {type(etree.Element).__name__}")
print(f"type(etree.ElementTree) = {type(etree.ElementTree).__name__}")

elem = etree.Element("root")
tree = etree.ElementTree(elem)

print("\nisinstance against etree.Element / etree.ElementTree:")
try:
    result = isinstance(elem, etree.Element)
    print(f"  isinstance(elem, etree.Element)     = {result}")
except TypeError as e:
    print(f"  isinstance(elem, etree.Element)     -> TypeError: {e}")

try:
    result = isinstance(tree, etree.ElementTree)
    print(f"  isinstance(tree, etree.ElementTree) = {result}")
except TypeError as e:
    print(f"  isinstance(tree, etree.ElementTree) -> TypeError: {e}")

```

volodymyrtur-wk: I've got your point, yes, you are right, I made changes

aaroncameron-wk (approved):

austinmatherne-wk (approved):

PR #2312, opened 2026-04-20
Type Hints: arelle.ModelManager

volodymyrtur-wk:
#### Description of change

#### Steps to Test
CI
**review**:
@Arelle/arelle

### DIFF
--- a/arelle/ModelManager.py
+++ b/arelle/ModelManager.py
@@ -4,16 +4,20 @@ See COPYRIGHT.md for copyright information.
 from __future__ import annotations
 
 from collections.abc import Callable
-from typing import TYPE_CHECKING
-import gc, sys, traceback, logging
-from arelle import ModelXbrl, Validate, DisclosureSystem, PackageManager, ValidateXbrlCalcs, ValidateDuplicateFactsConst
+from typing import TYPE_CHECKING, Any
+import gc, traceback, logging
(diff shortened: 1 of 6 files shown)


derekgengenbacher-wf (dismissed):

arelle/ModelManager.py
```diff
@@ -149,43 +154,42 @@ def load(self, filesource, nextaction=None, taxonomyPackages=None, **kwargs):
...
-               schemaRefs=None,
-               createModelDocument=True,
-               isEntry=False,
-               errorCaptureLevel=None,
-               initialXml=None,
-               base=None) -> ModelXbrl.ModelXbrl:
-        self.modelXbrl = ModelXbrl.create(self,
+               newDocumentType: int | None = None,
+               url: str | None = None,
+               schemaRefs: str | None = None,
```
austinmatherne-wk: I think the `schemaRefs` type might be wrong (here and the preexisting type in `ModelXbrl.create`). Downstream code treats it as a list: https://github.com/Arelle/Arelle/blob/f2c91b51a40bc04d09b43b569f2128786efc9121/arelle/ModelDocument.py#L497-L499

volodymyrtur-wk: Yes, you are right. I picked type annotations from `ModelXbrl.create` and didn't check further where it uses.

austinmatherne-wk (approved):

derekgengenbacher-wf (approved):

derekgengenbacher-wf:

PR #630, opened 2023-03-29
Changeover to lxml ixbrl schema validation

hermfischer:
#### Reason for change
Arelle had home-made ixbrl validation based on DTD and custom ix-element/attribute validation because lxml previously was not capable of inline schema-based validation.

With it now possible, monster UBS 20-F filing ix validation goes down from 3.843 secs to 0.377 secs on a MacBook Pro (M1).

#### Description of change
Ditch DTD and custom ix validation.

Build custom resolver for lxml to deal with absence of baseURI provided from parent to child being resolved.

Custom code around lxml failure to validate @class="" needing minLenth 1.

Transfer XhtmlValidate ix checks which are not schema based to ModelDocument parsing pass over ix html elements.

#### Steps to Test
Run XII inline conformance suite.

**review**:
@Arelle/arelle

PR #2316, opened 2026-04-22
Type Hints: arelle.XhtmlValidate

volodymyrtur-wk:
#### Description of change

#### Steps to Test
CI
**review**:
@Arelle/arelle

### DIFF
--- a/arelle/XhtmlValidate.py
+++ b/arelle/XhtmlValidate.py
@@ -3,12 +3,18 @@ See COPYRIGHT.md for copyright information.
 
 (originally part of XmlValidate, moved to separate module)
 '''
+from typing import cast
+
 from arelle import XbrlConst, XmlUtil, XmlValidate, ValidateFilingText, UrlUtil
 from arelle.ModelXbrl import ModelXbrl
 from arelle.ModelObject import ModelObject
 from lxml import etree
 import os
 from urllib.parse import urljoin
+
+from arelle.typing import TypeGetText
+
+_: TypeGetText
 
 htmlEltUriAttrs = { # attributes with URI content (for relative correction and %20 canonicalization
     "a": {"href"},
@@ -59,7 +65,13 @@ INLINE_1_0_SCHEMA = "http://www.xbrl.org/2008/inlineXBRL/xhtml-inlinexbrl-1_0.xs
 INLINE_1_1_SCHEMA = "http://www.xbrl.org/2013/inlineXBRL/xhtml-inlinexbrl-1_1.xsd"
 
 
-def ixMsgCode(codeName, elt=None, sect="constraint", ns=None, name=None) -> str:
+def ixMsgCode(
+        codeName: str,
+        elt: ModelObject | None = None,
+        sect: str = "constraint",
+        ns: str | None = None,
+        name: str | None = None
+    ) -> str:
     if elt is None:
         if ns is None: ns = XbrlConst.ixbrl11
         if name is None: name = "other"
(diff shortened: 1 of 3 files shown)


arelle/XhtmlValidate.py
```diff
@@ -91,24 +103,19 @@ def xhtmlValidate(modelXbrl: ModelXbrl, elt: ModelObject) -> None:
...
 
-    try:
-        if validateEntryText:
-            ValidateFilingText.validateHtmlContent(modelXbrl, elt, elt, "InlineXBRL", valHtmlContentMsgPrefix, isInline=True)
-    except XMLSyntaxError as err:
-        modelXbrl.error("html:syntaxError",
-            _("%(element)s error %(error)s"),
-            modelObject=elt, element=elt.localName.title(), error=', '.join(dtdErrs()))
+    if validateEntryText:
+        ValidateFilingText.validateHtmlContent(modelXbrl, elt, elt, "InlineXBRL", valHtmlContentMsgPrefix, isInline=True)  # type: ignore[no-untyped-call]
```
aaroncameron-wk, line 108: What's the reasoning behind stripping this try/except block?

austinmatherne-wk: Vova asked me about this after standup this morning. It would've been good to commit this separately with an explanatory comment. The except case used an undeclared function (`dtdErrs`) that was removed in #630. It didn't cause issues at runtime because `XMLSyntaxError` is never thrown by the try block. It was missed cleanup in the previous PR.

arelle/XhtmlValidate.py
```diff
@@ -72,13 +84,13 @@ def ixMsgCode(codeName, elt=None, sect="constraint", ns=None, name=None) -> str:
...
                 name = "resources"
-    return "{}:{}".format(ixSect[ns].get(name,"other")[sect], codeName)
+    return "{}:{}".format(ixSect[ns].get(name,"other")[sect], codeName)  # type: ignore[index]
 
 def xhtmlValidate(modelXbrl: ModelXbrl, elt: ModelObject) -> None:
     from lxml.etree import XMLSyntaxError
     validateEntryText = modelXbrl.modelManager.disclosureSystem.validateEntryText
     if validateEntryText:
-        valHtmlContentMsgPrefix = modelXbrl.modelManager.disclosureSystem.validationType + ".5.02.05."
+        valHtmlContentMsgPrefix = cast(str, modelXbrl.modelManager.disclosureSystem.validationType) + ".5.02.05."
```
austinmatherne-wk: This is fine, but could be an assertion instead of a cast.

austinmatherne-wk (dismissed):

arelle/XhtmlValidate.py
volodymyrtur-wk: I think the `TypeError`, in case `validationType` is `None`, will be more informative than `AssertionError`

aaroncameron-wk (dismissed):

arelle/XhtmlValidate.py
austinmatherne-wk: Out of curiosity, why? I don't feel strongly, but the requirement here is that if a disclosure system has `validateEntryText` set than `validationType` must also be set. At least in my head that maps cleanly to assertion error which directly explains what is wrong and needs to be fixed. The `TypeError` here doesn't explain the requirement or provide an informative error message?

derekgengenbacher-wf (dismissed):

derekgengenbacher-wf (approved):

austinmatherne-wk (approved):

PR #2317, opened 2026-04-22
Implement `PackageSource`

aaroncameron-wk:
#### Reason for change
Part of the non-breaking phase of #2269.
Establish the official public API for package mapping and interaction.

#### Description of change
Introduce a new component of the package mapping system: `PackageSource`. This component serves as a public API wrapper for interacting with the internal package mapping system.

Migrate most internal package mapping use to new API. Remaining `PackageManager` uses will be addressed in subsequent PRs.

#### Steps to Test
CI

**review**:
@Arelle/arelle

### DIFF
--- a/arelle/CntlrWinMain.py
+++ b/arelle/CntlrWinMain.py
@@ -593,15 +593,15 @@ class CntlrWinMain(Cntlr.Cntlr):
         self.fileMenu.add_cascade(label=_("Recent imports"), menu=self.recentAttachMenu, underline=0)
         self.packagesMenu = Menu(self.menubar, tearoff=0)
         hasPackages = False
-        for i, packageInfo in enumerate(sorted(self._packageManager.packagesConfig.get("packages", []),
-                                               key=lambda packageInfo: (packageInfo.get("name",""),packageInfo.get("version",""))),
+        for i, packageMeta in enumerate(sorted(self.packages.get_packages(),
+                                               key=lambda packageMeta: (packageMeta.name,packageMeta.version)),
                                         start=1):
-            name = packageInfo.get("name", "package{}".format(i))
-            version = packageInfo.get("version")
+            name = packageMeta.name if packageMeta.name is not None else "package{}".format(i)
+            version = packageMeta.version
             if version:
                 name = "{} ({})".format(name, version)
-            URL = packageInfo.get("URL")
-            if name and URL and packageInfo.get("status") == "enabled":
+            URL = packageMeta.url
+            if name and URL and packageMeta.status == "enabled":
                 self.packagesMenu.add_command(
                      label=name,
                      command=lambda url=URL: self.fileOpenFile(url))  # type: ignore[misc]
--- a/arelle/CntlrWinMain.py
+++ b/arelle/CntlrWinMain.py
@@ -593,9 +593,7 @@ class CntlrWinMain(Cntlr.Cntlr):
         self.fileMenu.add_cascade(label=_("Recent imports"), menu=self.recentAttachMenu, underline=0)
         self.packagesMenu = Menu(self.menubar, tearoff=0)
         hasPackages = False
-        for i, packageMeta in enumerate(sorted(self.packages.get_packages(),
-                                               key=lambda packageMeta: (packageMeta.name,packageMeta.version)),
-                                        start=1):
+        for i, packageMeta in enumerate(self.packages.get_packages(), start=1):
             name = packageMeta.name if packageMeta.name is not None else "package{}".format(i)
             version = packageMeta.version
             if version:
--- a/arelle/CntlrWinMain.py
+++ b/arelle/CntlrWinMain.py
@@ -78,7 +78,6 @@ from arelle import (
     DialogPluginManager,
     DialogURL,
     ModelDocument,
-    PackageManager,
     TableStructure,
     Updater,
     ViewFileConcepts,
@@ -594,7 +593,7 @@ class CntlrWinMain(Cntlr.Cntlr):
         self.fileMenu.add_cascade(label=_("Recent imports"), menu=self.recentAttachMenu, underline=0)
         self.packagesMenu = Menu(self.menubar, tearoff=0)
         hasPackages = False
-        for i, packageInfo in enumerate(sorted(PackageManager.packagesConfig.get("packages", []),
+        for i, packageInfo in enumerate(sorted(self._packageManager.packagesConfig.get("packages", []),
                                                key=lambda packageInfo: (packageInfo.get("name",""),packageInfo.get("version",""))),
                                         start=1):
             name = packageInfo.get("name", "package{}".format(i))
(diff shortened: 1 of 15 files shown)


arelle/packages/package_meta.py
```diff
@@ -0,0 +1,60 @@
...
+from __future__ import annotations
+
+from dataclasses import dataclass
+from typing import Any
+
+
+@dataclass(frozen=True)
+class PackageMeta:
+    description: str
+    entry_points: dict[str, frozenset[tuple[str, str, str]]] # Path, URL, Description
```
austinmatherne-wk: Thoughts on typing read-only maps as `Mapping[..., ...]`? On the table constraints branch, I'm using `Mapping` for the type and `MappingProxyType` for the implementation on `dataclass` fields. But I also realize I'm being inconsistent with how I handle read-only sets, where I use `frozenset` for both the type and the implementation rather than `collections.abc.Set` as the type.

arelle/plugin/formulaXPathChecker.py
```diff
@@ -290,16 +290,16 @@ def validateUtilityRun(cntlr, options, *args, **kwargs):
 def setupPackageEntrypoints(cntlr, options, filesource, entrypointFiles, *args, **kwargs):
     # check package entries formula code
     if getattr(options, "checkPackageEntries", False) and not entrypointFiles:
-        for packageInfo in sorted(PackageManager.packagesConfig.get("packages", []),
-                                  key=lambda packageInfo: (packageInfo.get("name",""),packageInfo.get("version",""))):
+        for packageMeta in sorted(cntlr.packages.get_packages(),
+                                  key=lambda packageInfo: (packageMeta.name ,packageMeta.version)):
```
austinmatherne-wk: Typo: local lambda var is still `packageInfo` not `packageMeta`

arelle/packages/package_meta.py
```diff
@@ -0,0 +1,60 @@
...
+                    (path, url, description)
+                    for path, url, description in entry_points
+                )
+                for name, entry_points in config.get("entryPoints", {}).items()
+            },
+            file_date=str(config.get("fileDate", "")),
+            identifier=str(config.get("identifier", "")),
+            license=str(config.get("license", "")),
+            manifest_name=str(config.get("manifestName", "")),
+            name=str(config.get("name", "")),
```
austinmatherne-wk: I think "unset" values [wind up](https://github.com/Arelle/Arelle/blob/fddb9bd9278e369bccdf5d34a1431ccccfed6ef7/arelle/packages/_package_manager.py#L650) as `None` values in config? So those values result in `str(None)` -> `"None"` strings here instead of the default empty string values that this is trying use?

arelle/CntlrWinMain.py
```diff
@@ -594,15 +593,15 @@ def loadFileMenuHistory(self) -> None:
...
         self.packagesMenu = Menu(self.menubar, tearoff=0)
         hasPackages = False
-        for i, packageInfo in enumerate(sorted(PackageManager.packagesConfig.get("packages", []),
-                                               key=lambda packageInfo: (packageInfo.get("name",""),packageInfo.get("version",""))),
+        for i, packageMeta in enumerate(sorted(self.packages.get_packages(),
+                                               key=lambda packageMeta: (packageMeta.name,packageMeta.version)),
                                         start=1):
-            name = packageInfo.get("name", "package{}".format(i))
-            version = packageInfo.get("version")
+            name = packageMeta.name if packageMeta.name is not None else "package{}".format(i)
```
austinmatherne-wk, line 597: `packageMeta.name` is typed as `str` not `str | None` so the fallback never happens? If the previous fallback behavior is needed it should be based on an empty name string.

arelle/packages/package_source.py
```diff
@@ -0,0 +1,52 @@
+"""
+See COPYRIGHT.md for copyright information.
+"""
+from __future__ import annotations
+
+from arelle.packages._package_manager import PackageManager
+from arelle.packages.package_meta import PackageMeta
+
+
+class PackageSource:
```
austinmatherne-wk: This is fine, but the `PackageSource` class name is a little confusing to me. Perhaps `PackageProvider` (consistent with `PluginProvider`, but maybe makes it easy to mix up the two), `PackageCatalog`, or `PackageRegistry`?

arelle/packages/package_meta.py
volodymyrtur-wk: I agree with Austin. Instead of casting, you could use `name=config.get("name") or ""`. This solution allows us to avoid  cases like `str(None)` -> `"None"` and set a blank line as "default" value

arelle/packages/package_meta.py
aaroncameron-wk: I think the ideal is `Mapping`/`Set` for hints. I'll update here, and we can make an effort to be consistent with that going forward.

arelle/packages/package_source.py
aaroncameron-wk: `PackageProvider` is what I started with, but I felt that a "provider" having a public "add" method was counter-intuitive. I think `PackageRegistry` is a good alternative.

arelle/packages/package_meta.py
austinmatherne-wk: Sounds good. I'll also update my branch to use `Set` for hints instead of `frozenset`

arelle/CntlrWinMain.py
aaroncameron-wk: I opted to pass through `None` values to these fields rather than force a `str` value.

arelle/packages/package_meta.py
aaroncameron-wk: I opted to pass through `None` values to these fields rather than force a `str` value.

arelle/CntlrWinMain.py
```diff
@@ -594,15 +593,15 @@ def loadFileMenuHistory(self) -> None:
         self.fileMenu.add_cascade(label=_("Recent imports"), menu=self.recentAttachMenu, underline=0)
         self.packagesMenu = Menu(self.menubar, tearoff=0)
         hasPackages = False
-        for i, packageInfo in enumerate(sorted(PackageManager.packagesConfig.get("packages", []),
-                                               key=lambda packageInfo: (packageInfo.get("name",""),packageInfo.get("version",""))),
+        for i, packageMeta in enumerate(sorted(self.packages.get_packages(),
+                                               key=lambda packageMeta: (packageMeta.name,packageMeta.version)),
```
austinmatherne-wk: With the package meta type changes to `str | None` fields, both here and the sort in [formulaXPathChecker.py‎](https://github.com/Arelle/Arelle/pull/2317/changes#diff-a97e3f17adbda648e0c9e99bedea2cf123662ac5079b0ef0476c88d1e41e9e8d) need to guard against type errors being raised from comparing `None` and `str` values across different packages.

aaroncameron-wk: Good catch. I think this sorting actually became redundant since `get_packages` internally sorts the packages by essentially the same criteria, so I've removed it.

austinmatherne-wk (approved):

derekgengenbacher-wf (approved):

derekgengenbacher-wf:

PR #2319, opened 2026-04-23
Table Constraints Metadata Validation

### LINKED ISSUE #2195: TC Metadata: Property Validation and non-Parsing Metadata Validation Code Path

Validate that xBRL-CSV metadata files do not include any `tc:` prefixed properties other than those defined in the spec, and that defined properties do not appear within JSON objects other than those specified.

Error code: `tcme:misplacedOrUnknownProperty`.

This should also set up the non-parsing metadata validation code path for other metadata validations to reuse.

### LINKED ISSUE #2196: TC Metadata: Column Parameter Conflict Detection

Validate that a constrained column does not have the same name as a defined parameter in the same table template. This is the metadata-time check that can be performed statically from the template definitions.

Error code: `tcme:columnParameterConflict`.

### LINKED ISSUE #2226: TC Metadata: Set/Array and Required Property Validation

Validate set, array, and required property structural constraints on TC related JSON metadata.

Error code: `tcme:invalidJSONStructure`.

### LINKED ISSUE #2227: TC Metadata: Unknown Properties in TC Objects

Validate unknown properties in TC objects.

Error code: `tcme:invalidJSONStructure`.

austinmatherne-wk:
#### Reason for change
Building on table constraints metadata support. In an effort to keep reviews manageable I've removed value and key metadata constraint validation from this branch and will PR them in two subsequent tickets.

Changes are intended to be reviewed commit by commit.

#### Description of change
There's an inherent awkwardness to Arelle's validation flow that's not new for table constraints, but is surfaced by it. For the CLI we parse and then validate immediately after (depending on whether `--validate` is used). In the GUI we parse and keep the model open, and then some time later the user can validate the report one or multiple times. Some validation messages are logged during the parsing stage while most occur during the validation flow. The result is that if the user clears the GUI messages (or performs other operations that clutter the log) and then validates, they don't see any validation errors that occurred during parsing. Multiple users have messaged support because they were confused by this behavior. Table constraints is in an odd place because it doesn't serve any purpose unless validating, so a reasonable case can be made that we shouldn't parse it at all until validation is requested. I may end up moving TC parsing behind validation, but for the moment I've opted to align with the existing parse/validation separation. To support that split, this PR introduces a typed, resolved xBRL-CSV metadata model and a parser that converts the merged OIM dicts into those dataclasses. The CSV and TC metadata are wrapped in a single `XbrlCsvLoadingContext` attached to `ModelXbrl`, replacing the previous `tcMetadata` attribute.

Conformance suite bumped from CR to PR. Note, another PR suite is expected with many more tests. 

`TCMetadataValidationError` now supports multiple JSON pointer paths per error and properly escapes paths.

A new `TCMetadataValidator` is invoked during OIM validation and currently produces the following errors:
- `tcme:invalidNamespacePrefix`
- `tcme:columnParameterConflict`
- `tcme:inconsistentColumnOrderDefinition`

Various issues surfaced while wiring this up:
- Spurious `oimce:invalidSQName` for columns named "dimensions"
- CSV dialect sniffer logged errors on single column xBRL-CSV files.
- `DUPJSONVALUE` sentinel in merged OIM object wasn't cleared after loading/validating duplicate keys/values.
- Spurious `oime:invalidXBRL` caused by `numFactCreationXbrlErrors` going negative when unrelated prior errors were present.
- `isModified` was set on exception and None return values.

#### Steps to Test
* CI

**review**:
@Arelle/arelle


arelle/oim/_tc/metadata/parser.py
```diff
@@ -229,6 +272,84 @@ def _parse_template_table_constraints(
...
+
+
+def _validate_misplaced_in_columns(
+    template_id: str,
+    columns: Any,
+    errors: list[TCMetadataParseError],
+) -> None:
+    if not isinstance(columns, dict):
+        return
+    for col_id, col_obj in columns.items():
```
volodymyrtur-wk, line 313: you can simplify the function by making opposite condition `if isinstance(columns, dict)` and move the loop under it

arelle/oim/_tc/metadata/parser.py
```diff
@@ -229,6 +272,84 @@ def _parse_template_table_constraints(
...
+            _walk_validate_misplaced_value(item, errors, *path, str(i))
+
+
+def _validate_xs_type_prefixes(
+    template_constraints: dict[str, TCTemplateConstraints],
+    namespaces: dict[str, str],
+) -> TCMetadataParseError | None:
+    if namespaces.get("xs") == xsd:
+        return None
+    if any(
```
volodymyrtur-wk, line 341: If you change the condition to `if namespaces.get("xs") != xsd:` it allows you to move the rest of the code under it and reduce the number of `return` statements

arelle/oim/csv/metadata/parser.py
```diff
@@ -0,0 +1,277 @@
...
+
+
+def _parse_taxonomy_defined_dimensions(dims: dict[str, Any]) -> Mapping[str, str | None]:
+    return MappingProxyType(
+        {
+            key: val
+            for key, val in dims.items()
+            if isinstance(key, str)
+            if key not in _BUILT_IN_DIMENSION_KEYS
+            if isinstance(val, str) or val is None
```
volodymyrtur-wk, line 188: Why don't you use `or` instead of a bunch of `if`?
```
{
    key: val
    for key, val in dims.items()
    if isinstance(key, str) 
       or key not in _BUILT_IN_DIMENSION_KEYS 
       or (isinstance(val, str) or val is None)
}

File: arelle/oim/csv/metadata/parser.py [hunk 4]
```diff
@@ -0,0 +1,277 @@
...
+def _as_bool(item: Any) -> bool:
+    return item if isinstance(item, bool) else False
+
+
+def _as_decimals_or_none(item: Any) -> int | str | None:
+    if isinstance(item, bool):
+        return None
+    if isinstance(item, (int, str)):
+        return item
+    return None
```
volodymyrtur-wk, line 277: the condition `if isinstance(item, bool)` doesn't make sense because we return `None` in any case except `if isinstance(item, (int, str))`. The function `_as_str_or_none` is a good example of a solution

austinmatherne-wk: It's required because in Python bool is a subclass of int:

```python
if isinstance(True, int):
    print("True is an int")
else:
    print("True is not an int")

> True is an int
```

arelle/oim/csv/metadata/parser.py
austinmatherne-wk: No strong reason. I just found it easier to read.

arelle/oim/_tc/metadata/parser.py
austinmatherne-wk: The choice was intentional. We've settled on early exit / guard clauses as the team convention for validation code. In isolation this one could go either way, but a lot of these functions accumulate many preconditions (fact valid, fact type X, concept valid, concept type Y, etc.) where inverting would nest the happy path several levels deep or require a very complex if statement. I find it easier to stay consistent across the codebase than to mix styles based on how many guards a given function happens to have today. Happy to revisit if you feel strongly.

arelle/oim/_tc/metadata/parser.py
austinmatherne-wk: Same reasoning as https://github.com/Arelle/Arelle/pull/2319#discussion_r3133424986, early exit is our convention for validation.

derekgengenbacher-wf (approved):

derekgengenbacher-wf:

PR #2320, opened 2026-04-23
Restrict Webserver Plugins

austinmatherne-wk:
#### Reason for change

Internally it's been understood that the web server should never be exposed to untrusted users, networks, or anyone you wouldn't otherwise grant shell access to, but this wasn't documented. Thankfully there aren't any documented instances of folks doing this.

#### Description of change

* Reject web server requests to load remote Python plugins. I don't think there's a compelling reason to support this, but we can restore it behind an opt-in flag if someone raises an issue.
* Document the lack of any real security for the web server.
* Emit security warning to the console at web server startup.

#### Steps to Test
* CI

**review**:
@Arelle/arelle


volodymyrtur-wk (approved):

aaroncameron-wk (approved):

derekgengenbacher-wf (approved):

derekgengenbacher-wf:

PR #2321, opened 2026-04-23
Type Hints: arelle.ValidateUtr

volodymyrtur-wk:
#### Description of change
This module is excluded from mypy type checking. We should remove it from the exclusion list and add types so the it passes mypy checks

#### Steps to Test
CI
**review**:
@Arelle/arelle

### DIFF
--- a/arelle/ValidateUtr.py
+++ b/arelle/ValidateUtr.py
@@ -150,7 +150,7 @@ class ValidateUtr:
         self.messageCode = messageCode
         if getattr(modelXbrl.modelManager.disclosureSystem, "utrItemTypeEntries", None) is None:
             loadUtr(modelXbrl)
-        self.utrItemTypeEntries: dict[str, dict[str, UtrEntry]] = modelXbrl.modelManager.disclosureSystem.utrItemTypeEntries  # type: ignore[attr-defined]
+        self.utrItemTypeEntries: dict[str | None, dict[str | None, UtrEntry]] = modelXbrl.modelManager.disclosureSystem.utrItemTypeEntries  # type: ignore[attr-defined]
 
     def validateFacts(self) -> None:
         modelXbrl = self.modelXbrl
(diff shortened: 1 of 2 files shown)


arelle/ValidateUtr.py
```diff
@@ -117,35 +126,36 @@ def loadUtr(modelXbrl, statusFilters=None): # Build a dictionary of item types t
...
 
-    def __init__(self, modelXbrl: ModelXbrl, messageLevel: str="ERROR", messageCode: str="utre:error-NumericFactUtrInvalid") -> None:
+    def __init__(self, modelXbrl: ModelXbrl, messageLevel: str = "ERROR", messageCode: str = "utre:error-NumericFactUtrInvalid") -> None:
         self.modelXbrl = modelXbrl
         self.messageLevel = messageLevel
         self.messageCode = messageCode
         if getattr(modelXbrl.modelManager.disclosureSystem, "utrItemTypeEntries", None) is None:
             loadUtr(modelXbrl)
-        self.utrItemTypeEntries: dict[str, dict[str, UtrEntry]] = modelXbrl.modelManager.disclosureSystem.utrItemTypeEntries
+        self.utrItemTypeEntries: dict[str, dict[str, UtrEntry]] = modelXbrl.modelManager.disclosureSystem.utrItemTypeEntries  # type: ignore[attr-defined]
```
aaroncameron-wk: Shouldn't this be the same as above?
```
utrItemTypeEntries: defaultdict[str | None, dict[str | None, UtrEntry]]
```
I think it would let us remove a `type: ignore[index]` below.

volodymyrtur-wk: Good point, thanks

derekgengenbacher-wf (approved):

austinmatherne-wk (approved):

derekgengenbacher-wf:

PR #2328, opened 2026-04-24
Adding EDINET validation EC8036W

allenbross-wf:
#### Description of change
```
EDINET.EC8036W - Detailed tagging of consolidated management indicators, etc. has not been done.
Please provide detailed tags for consolidated management indicators, etc. If you do not provide consolidated management indicators, etc., please confirm that the "Consolidated Financial Statements" field in the DEI information is correct.

Implementation: If jpdei_cor:WhetherConsolidatedFinancialStatementsArePreparedDEI = True, then there must be children of rol_BusinessResultsOfGroup
```

**AI generated conformance suite test case.** 

#### Steps to Test

CI

**review**:
@Arelle/arelle


arelle/plugin/validate/EDINET/rules/edinet.py
```diff
@@ -940,6 +941,41 @@ def rule_EC8034W(
...
+        *args: Any,
+        **kwargs: Any,
+) -> Iterable[Validation]:
+    """
+    EDINET.EC8036W: The consolidated business indicators, etc. have not been detailed tagged.
+    If WhetherConsolidatedFinancialStatementsArePreparedDEI is true, there must be children of
+    the extended link role rol_BusinessResultsOfGroup.
+    """
+    # if not hasattr(val, "modelXbrl"):
+    #     return
```
austinmatherne-wk: Remove commented out code, please

arelle/plugin/validate/EDINET/rules/edinet.py
```diff
@@ -940,6 +941,41 @@ def rule_EC8034W(
...
+        "http://disclosure.edinet-fsa.go.jp/role/jpcrp/rol_BusinessResultsOfGroup",
+    )
+    if relSet is not None and len(relSet.modelRelationships) > 0:
+        return
+    yield Validation.warning(
+        codes='EDINET.EC8036W',
+        msg=_("The consolidated business indicators, etc. have not been tagged in detail. "
+              "Please provide detailed tagging of the consolidated business indicators, etc. "
+              "If you do not provide consolidated business indicators, please confirm that "
+              'the "Consolidated Financial Statements" field in the DEI information is correct.'),
```
austinmatherne-wk: The validation logic is checking that concepts exist under `BusinessResultsOfGroup` in the presentation network, but the error message suggests that these concepts must actually be tagged. If that's the actual intent, then [ValidationUtil.hasPresentationalConceptsWithFacts](https://github.com/Arelle/Arelle/blob/0733db1f789202e3417d2b71dbd02d486da01299/arelle/utils/validate/ValidationUtil.py#L30) can be used.

arelle/plugin/validate/EDINET/rules/edinet.py
```diff
@@ -940,6 +941,41 @@ def rule_EC8034W(
                 )
 
 
+@validation(
+    hook=ValidationHook.XBRL_FINALLY,
```
austinmatherne-wk: Does this rule belong on the `XBRL_FINALLY` (executed for each XBRL model) or `COMPLETE` (executed once for all models) validation hook? The other rules surrounding this one use the `COMPLETE` hook.

allenbross-wf: I think this rule belongs to  'XBRL_FINALLY'. If it inspects all of the models it will exit the rule when it encounters any of the AuditDoc models.

austinmatherne-wk: My understanding of EDINET filings is limited, so I could be (probably am) wrong. But to be sure, all the models loaded from the EDINET manifest must include this tagging, not just one?

arelle/plugin/validate/EDINET/rules/edinet.py
```diff
@@ -941,6 +941,35 @@ def rule_EC8034W(
...
+    """
+    EDINET.EC8036W: The consolidated business indicators, etc. have not been detailed tagged.
+    If WhetherConsolidatedFinancialStatementsArePreparedDEI is true, there must be children of
+    the extended link role rol_BusinessResultsOfGroup.
+    """
+    deiQname = qname(pluginData.namespaces.jpdei, 'WhetherConsolidatedFinancialStatementsArePreparedDEI')
+    if not any(fact.xValue == True for fact in iterValidNonNilFactsByQname(val.modelXbrl, deiQname)):
+        return
+    if hasPresentationalConceptsWithFacts(val.modelXbrl, 'http://disclosure.edinet-fsa.go.jp/role/jpcrp/rol_BusinessResultsOfGroup'):
+                return
```
austinmatherne-wk: Indention is off here

arelle/plugin/validate/EDINET/rules/edinet.py
austinmatherne-wk: Or I guess the `deiQname` filter/early exit prevents that from being an issue (only the models that need to be validated include that fact)?

derekgengenbacher-wf (dismissed):

arelle/plugin/validate/EDINET/rules/edinet.py
```diff
@@ -941,6 +942,39 @@ def rule_EC8034W(
...
+    """
+    if not pluginData.hasDocumentType({DocumentType.ANNUAL_SECURITIES_REPORT, DocumentType.SEMI_ANNUAL_REPORT}):
+        return
+    deiQname = qname(pluginData.namespaces.jpdei, 'WhetherConsolidatedFinancialStatementsArePreparedDEI')
+    for modelXbrl in pluginData.loadedModelXbrls:
+        if not any(fact.xValue == True for fact in iterValidNonNilFactsByQname(modelXbrl, deiQname)):
+            continue
+        if hasPresentationalConceptsWithFacts(modelXbrl, 'http://disclosure.edinet-fsa.go.jp/role/jpcrp/rol_BusinessResultsOfGroup'):
+            continue
+        yield Validation.warning(
```
aaroncameron-wk: DEI values are only set on one IXDS but are relevant to the entire multi-IXDS. This implementation seems to check if each individual IXDS passes the validation, but I think it should probably check the DEI at a global level (see `getDeiValue`) and then check if `rol_BusinessResultsOfGroup` facts exist in _any_ instance. 

That might resolve the number of valid/existing filings firing the error based on your current implementation.

arelle/plugin/validate/EDINET/rules/edinet.py
aaroncameron-wk: > all the models loaded from the EDINET manifest must include this tagging, not just one

I'm not sure either way, but the fact that the current implementation is firing so widely might suggest the tagging only needs to exist on one of the IXDS. The DEI facts are only ever on one of the IXDS, so supplementary IXDS would never even be checked based on the current logic.

arelle/plugin/validate/EDINET/rules/edinet.py
aaroncameron-wk: `rule_EC8038W` below, for example, passes if any IXDS has an appropriate fact. The only difference with this rule is that we also need to first check the DEI value to ensure the presentation check applies.

arelle/plugin/validate/EDINET/rules/edinet.py
aaroncameron-wk: This rule follows a similar pattern as EC8038W and EC8039W, and for those rules we check the DEI value at a filing level and then pass the validation if any IXDS contains a matching fact, so purely for consistency we should do the same here.

arelle/plugin/validate/EDINET/rules/edinet.py
```diff
@@ -942,6 +942,40 @@ def rule_EC8034W(
...
+        *args: Any,
+        **kwargs: Any,
+) -> Iterable[Validation]:
+    """
+    EDINET.EC8036W: The consolidated business indicators, etc. have not been detailed tagged.
+    If WhetherConsolidatedFinancialStatementsArePreparedDEI is true, there must be children of
+    the extended link role rol_BusinessResultsOfGroup.
+    """
+    if not pluginData.hasDocumentType({DocumentType.ANNUAL_SECURITIES_REPORT, DocumentType.SEMI_ANNUAL_REPORT}):
+        return
```
aaroncameron-wk, line 962: Did this condition end up having an effect on the results?

allenbross-wf: It did not seem to have any impact on the conformance suite tests we have.  I don't think that is the same thing as not having an effect on the results, though.

austinmatherne-wk (approved):

derekgengenbacher-wf (approved):

derekgengenbacher-wf:

PR #2336, opened 2026-04-25
Windows installer: pre-calculate installation size and update GitHub Action

strr:
#### Reason for change
I just tried out the new installer for 2.39.10 and there is now a step (introduced by me in #2306) at the end of the install where it calculates the installation size. This step just took ages (~ 2 minutes) to complete. I'm not sure why I didn't notice it being slow at the time of the PR, but it is definitely slow now.

#### Description of change
Remove use of installer's `${GetSize}` command (which stats every file in the installation directory). Replace with a build time estimated size which can be passed in on the command line.

Update the GitHub actions to calculate the BUILD_PATH size in kb and pass it in.

#### Steps to Test

master spend time towards the end of the installation displaying "Size: x Kb Files: y Folders: z"
image

The PR should not display that step at all.

**review**:
@Arelle/arelle

### DIFF
--- a/.github/workflows/build-windows.yml
+++ b/.github/workflows/build-windows.yml
@@ -125,7 +125,8 @@ jobs:
     - name: Make installer
       run: |
         $installer_version = ("${{ env.BUILD_VERSION }}" -replace '[^0-9.].*$', '' -replace '\.$', '') + ".${{ github.run_number }}"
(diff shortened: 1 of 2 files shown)


austinmatherne-wk (approved):

austinmatherne-wk:
Thanks. I had chalked this up to my slow Windows VM when testing.

PR #2339, opened 2026-04-27
Move webserver security warning to CntlrCmdLine.py

ryangeiser1:
#### Reason for change
The web server security warning is currently invoked by calling the web server from the CLI and the wsgiApplication. I believe this warning should only appear if the built-in web server is started from the CLI utility.

#### Description of change
Move web server security warning to CntlrCmdLine.py from CntlrWebMain.py. Create boolean flag to check for "--webserver" argument in CLI call and print warning if true.

#### Steps to Test
1. `python arelleCmdLine.py --webserver=127.0.0.1:8080`
2. Call arelle.CntlrCmdLine:wsgiApplication using a dedicated WSGI server (gunicorn, uWSGI, waitress, etc...)

**review**:
@Arelle/arelle
@austinmatherne-wk

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -50,7 +50,7 @@ Crypto = [
     'pycryptodome>=3,<4',
 ]
 DB = [
-    'oracledb>=3,<4',
+    'oracledb>=3,<5',
     'pg8000>=1,<2',
     'PyMySQL>=1,<2',
     'pyodbc>=4,<6',
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -3,7 +3,7 @@ flake8-noqa==1.5.0
 
 mypy==1.20.2
 
-types-boto3[s3]==1.43.2
+types-boto3[s3]==1.43.3
 types-lxml==2026.2.16
 types-openpyxl==3.1.5.20260408
 types-PyMySQL==1.1.0.20260408
--- a/requirements-docs.txt
+++ b/requirements-docs.txt
@@ -4,6 +4,7 @@ myst-parser==5.0.0 ; python_version > "3.10"
 sphinx==8.1.3 ; python_version == "3.10"
 sphinx==8.2.3 ; python_version == "3.11"
 sphinx==9.1.0 ; python_version > "3.11"
-sphinx-autobuild==2025.8.25
+sphinx-autobuild==2024.10.3 ; python_version == "3.10"
+sphinx-autobuild==2025.8.25 ; python_version > "3.10"
 sphinx-autodoc2==0.5.0
 sphinx-copybutton==0.5.2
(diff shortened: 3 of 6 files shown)


austinmatherne-wk:
> I believe this warning should only appear if the built-in web server is started from the CLI utility.

Could you say more about why you think the warning shouldn't fire when Arelle is hosted via a WSGI server? My concern is that the things the warning is actually about (no authentication on any REST route, the file= parameter opening arbitrary local paths, the plugins= parameter executing Python modules) all live in the application itself, not in Bottle's HTTP server. Fronting the same app with gunicorn doesn't restrict those parameters, so the security posture described in the docs seems to apply identically in both deployment environments.

If anything, I'd expect the WSGI path to be the likelier scenario for an operator to expose the app on a real network interface, so it's the path where I'd most want the warning to print at startup.

Is the underlying motivation that the warning is noisy in container/worker logs? If so, would an explicit opt-out flag (e.g. `--suppressWebserverSecurityWarning`) suffice?

Happy to be wrong about any of this, just want to make sure we're not silencing it in the place it's most needed.

ryangeiser1:
Correct, my primary motivation is noisy logs. I think adding the warning is positive, especially with the soon-to-be published Docker images. You did a good job documenting the risks in the markdown. If someone were to spin up a container and expose port 8080, they would be opening a front door to the entire internet with root access. Not good. 

However, I assume anyone with the knowledge and ability to spin up a Docker container and expose a port to the internet is sufficiently advanced to understand the risks associated. When a new operator spins up a container for the first time, I assume they will run the container as-is from Docker hub (or GHCR) and see the warning message on startup. Likewise with someone running locally or with the python package, if they're interested in the web server, they'll start by using the provided web server. 

If they need a more advanced setup (custom Dockerfile, custom plugins, dedicated WSGI server, etc...), they are probably familiar with the risks associated with hosting in general, and specifically with Arelle. If they are using the wsgiApplication directly, they are probably a more advanced operator and already familiar with the risks after seeing the warning message while developing. 

An explicit opt-out flag would suffice, though I think adding another CLI arg to block one warning message is overkill. Six of one, half a dozen of the other. I think it comes down to preference, to which I will defer to you.

strr:
Sometimes you see these sorts of things silenced via an environment variable:
`ARELLE_NO_SECURITY_WARNINGS=1`

austinmatherne-wk:
> However, I assume anyone with the knowledge and ability to spin up a Docker container and expose a port to the internet is sufficiently advanced to understand the risks associated.

Unfortunately, I don't think we can assume this. It's been my experience from handling Arelle support requests that many users copy and paste commands from other users without understanding their purpose. Given the risks here, I think requiring an explicit opt-out is called for. Stuart's suggestion makes sense to me.

ryangeiser1:
Makes sense. I'm happy using the environment variable. I can make those changes if y'all haven't already.

austinmatherne-wk:
> I can make those changes if y'all haven't already

We haven't. Go ahead please :)

PR #2341, opened 2026-04-27
Type Hints: arelle.LocalViewer

volodymyrtur-wk:
#### Description of change

#### Steps to Test
CI
**review**:
@Arelle/arelle

### DIFF
--- a/arelle/LocalViewer.py
+++ b/arelle/LocalViewer.py
@@ -67,7 +67,7 @@ class LocalViewer:
                 ex, traceback.format_exc()), messageCode="localViewer:exception",level=logging.DEBUG)
             return None
 
-    def get(self, file: str | None = None, relpath: str | None = None) -> object:
+    def get(self, file: str | None = None, relpath: str | None = None) -> HTTPResponse | None:
         assert self.cntlr is not None, "Cntlr not initialized"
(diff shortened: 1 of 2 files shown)


derekgengenbacher-wf (dismissed):

arelle/LocalViewer.py
```diff
@@ -50,16 +56,19 @@ def init(self, cntlr, reportsFolder):
...
             return localhost
         except Exception as ex:
+            assert self.cntlr is not None, "Cntlr not initialized"
             self.cntlr.addToLog(_("{} exception: http://localhost:{} \nException: {} \nTraceback: {}").format(
                 self.title, self.port,
                 ex, traceback.format_exc()), messageCode="localViewer:exception",level=logging.DEBUG)
+            return None
 
-    def get(self, file=None, relpath=None):
+    def get(self, file: str | None = None, relpath: str | None = None) -> object:
```
austinmatherne-wk: Can the return type here be `HTTPResponse | None`, or does that not work?

derekgengenbacher-wf (approved):

aaroncameron-wk (approved):

austinmatherne-wk (approved):

PR #2342, opened 2026-04-29
Type Hints: arelle.Validate

volodymyrtur-wk:
#### Description of change

#### Steps to Test
CI
**review**:
@Arelle/arelle

### DIFF
--- a/arelle/Validate.py
+++ b/arelle/Validate.py
@@ -35,6 +35,7 @@ from arelle.ModelXbrl import (
     create as modelXbrlCreate,
     ModelXbrl
 )
+from arelle.ErrorManager import ErrorsType
 from arelle.ModelObject import ModelObject
 from arelle.ModelTestcaseObject import testcaseVariationsByTarget, ModelTestcaseVariation
 from arelle.ModelValue import (qname, QName)
@@ -142,7 +143,6 @@ class Validate:
                     _("Versioning report exception: %(error)s, testcase: %(reportFile)s"),
                     modelXbrl=self.modelXbrl,
                     reportFile=self.modelXbrl.modelDocument.basename, error=err,
-                    #traceback=traceback.format_exc(),
                     exc_info=True)
         elif self.modelXbrl.modelDocument.type == Type.RSSFEED:
             self.validateRssFeed()
@@ -156,7 +156,6 @@ class Validate:
                     modelXbrl=self.modelXbrl,
                     instance=self.modelXbrl.modelDocument.basename if hasattr(self.modelXbrl, "modelDocument") and hasattr(self.modelXbrl.modelDocument, "basename") else "(closed)",
                     error=err,
-                    # traceback=traceback.format_exc(),
                     exc_info=True)
         self.close()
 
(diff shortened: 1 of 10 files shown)


derekgengenbacher-wf (dismissed):

arelle/Validate.py
```diff
@@ -301,14 +329,26 @@ def _validateTestcaseVariation(self, testcase, modelTestcaseVariation):
...
-    def _testcaseLoadReadMeFirstUri(self, testcase, modelTestcaseVariation, index, readMeFirstUri, resultIsVersioningReport, resultIsTaxonomyPackage, inputDTSes, errorCaptureLevel, baseForElement, parameters):
-        preLoadingErrors = [] # accumulate pre-loading errors, such as during taxonomy package loading
-        loadedModels = []
-        filesource = None
+    def _testcaseLoadReadMeFirstUri(
+        self,
+        testcase: ModelDocument,
+        modelTestcaseVariation: ModelTestcaseVariation,
+        index: int,
+        readMeFirstUri: str,
```
austinmatherne-wk: This can also be a `tuple`. I *think* `str | tuple[str | QName, str]`

arelle/Validate.py
```diff
@@ -1,44 +1,62 @@
 '''
 See COPYRIGHT.md for copyright information.
 '''
+from __future__ import annotations
+
 import bisect
 import fnmatch
-import os, sys, traceback, logging
```
austinmatherne-wk, line 6: Why was the `logging` import removed? It's used?

arelle/Validate.py
```diff
@@ -652,17 +706,22 @@ def _testcaseValidateInputDTS(self, testcase, modelTestcaseVariation, errorCaptu
...
 
-    def noErrorCodes(self, modelTestcaseVariationActual):
+    def noErrorCodes(self, modelTestcaseVariationActual: Iterable[Any]) -> bool:
         return not any(not isinstance(actual,dict) for actual in modelTestcaseVariationActual)
 
-    def determineTestStatus(self, modelTestcaseVariation, errors, validateModelCount=None):
+    def determineTestStatus(
+            self,
+            modelTestcaseVariation: ModelTestcaseVariation,
+            errors: list[str | dict[str, tuple[int, int, int, int, int]]],
```
austinmatherne-wk: Out of curiosity, where is this called with errors: `list[dict[str, tuple[int, int, int, int, int]]]`?

arelle/Validate.py
```diff
@@ -301,14 +329,26 @@ def _validateTestcaseVariation(self, testcase, modelTestcaseVariation):
...
+        testcase: ModelDocument,
+        modelTestcaseVariation: ModelTestcaseVariation,
+        index: int,
+        readMeFirstUri: str,
+        resultIsVersioningReport: bool,
+        resultIsTaxonomyPackage: bool,
+        inputDTSes: dict[str | None, list[ModelXbrl]],
+        errorCaptureLevel: int,
+        baseForElement: str,
+        parameters: dict[str, tuple[None, list[ModelXbrl]]],
```
austinmatherne-wk: I think this should be something along the lines of:

`dict[QName, tuple[QName | None, Any]]`

arelle/Validate.py
volodymyrtur-wk: yes, I removed unused imports

austinmatherne-wk: It's used: https://github.com/Arelle/Arelle/blob/a21397cfebb631542a1aa2873e99e2085fa9419a/arelle/Validate.py#L147

It works because there's [another logging import at the bottom of this file](https://github.com/Arelle/Arelle/blob/a21397cfebb631542a1aa2873e99e2085fa9419a/arelle/Validate.py#L806). But it would be better to remove that import and keep it at the top of the module.

aaroncameron-wk (approved):

allenbross-wf (approved):

derekgengenbacher-wf (approved):

derekgengenbacher-wf:

PR #2345, opened 2026-04-29
Adding validation EDINET.EC8037W

allenbross-wf:
#### Description of change

Added EDINET validation EC8037W: `If jpdei_cor:WhetherConsolidatedFinancialStatementsArePreparedDEI = False, then there must be children of the extended link role rol_BusinessResultsOfReportingCompany`

#### Steps to Test

CI

**review**:
@Arelle/arelle

### DIFF
--- a/arelle/plugin/validate/EDINET/rules/edinet.py
+++ b/arelle/plugin/validate/EDINET/rules/edinet.py
@@ -28,7 +28,7 @@ from arelle.utils.PluginHooks import ValidationHook
 from arelle.utils.validate.Characters import findProhibitedCharacters
 from arelle.utils.validate.Decorator import validation
 from arelle.utils.validate.Document import checkDocumentEncoding
-from arelle.utils.validate.Facts import isValidNonNilFact
+from arelle.utils.validate.Facts import isValidNonNilFact, iterValidNonNilFactsByQname
 from arelle.utils.validate.Validation import Validation
 from arelle.utils.validate.Common import isExtensionUri
 from arelle.utils.validate.ValidationUtil import hasPresentationalConceptsWithFacts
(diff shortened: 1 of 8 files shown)


austinmatherne-wk (dismissed):

tests/resources/conformance_suites/edinet/index.xml
```diff
@@ -174,6 +174,7 @@
     <testcase uri="EC8033W/index.xml" />
     <testcase uri="EC8034W/index.xml" />
     <testcase uri="EC8036W/index.xml" />
+    <!-- <testcase uri="EC8037W/index.xml" /> Not necessary, valid testcases fire this warning. -->
```
austinmatherne-wk, line 177: This is also true for `EC8038W` below, but we include a test case anyway.

derekgengenbacher-wf (approved):

derekgengenbacher-wf:

arelle/plugin/validate/EDINET/rules/edinet.py
```diff
@@ -976,6 +976,40 @@ def rule_EC8036W(
...
+) -> Iterable[Validation]:
+    """
+    EDINET.EC8037W: The consolidated business indicators, etc. have not been detailed tagged.
+    If WhetherConsolidatedFinancialStatementsArePreparedDEI is false, there must be children of
+    the extended link role rol_BusinessResultsOfReportingCompany.
+    """
+    if not pluginData.hasDocumentType({DocumentType.ANNUAL_SECURITIES_REPORT, DocumentType.SEMI_ANNUAL_REPORT}):
+        return
+    statementsPrepared = pluginData.getDeiValue('WhetherConsolidatedFinancialStatementsArePreparedDEI')
+    if statementsPrepared != False:
```
austinmatherne-wk, line 998: This is fine (it matches what we've already done elsewhere in this module), but we should really be using `is` and `is not` when confirming that these multi type `TypeXValue` values are a specific type of boolean. Equality comparison of booleans can return unexpected results. For example: 

```python
1 == True
> True

1 is True
> False
```

PR #2352, opened 2026-05-01
avoid exceptions when an invalid typed member is used

freddym56:
#### Reason for change
Exceptions are raised if a context contains an invalid typed member

#### Description of change
Added additional condition to check that the typedMember is not None.

#### Steps to Test
The attached filing has an empty typedMember element. If we run it though Arelle with the following command it will result in an exception. This PR is meant to avoid the exception and only output the xbrldie:IllegalTypedDimensionContentError error.

`arelleCmdLine.py --xdgConfigHome $XDG_CONFIG_HOME --file "$PATH_TO_FILE/EXFILINGFEES.htm" -v --disclosureSystem efm-pragmatic --plugins EdgarRenderer`

[EXFILINGFEES.htm](https://github.com/user-attachments/files/27286847/EXFILINGFEES.htm)

**review**:
@Arelle/arelle

### DIFF
--- a/arelle/ModelXbrl.py
+++ b/arelle/ModelXbrl.py
@@ -699,7 +699,7 @@ class ModelXbrl:
                     dimElt = XmlUtil.addChild(contextElt, XbrlConst.xbrldi, "xbrldi:typedMember",
                                               attributes=dimAttr)
                     if isinstance(dimValue, (arelle.ModelInstanceObject.ModelDimensionValue, DimValuePrototype)) and dimValue.isTyped and dimValue.typedMember is not None:
-                        XmlUtil.copyNodes(dimElt, cast(ModelObject, dimValue.typedMember))
+                        XmlUtil.copyNodes(dimElt, dimValue.typedMember)
                 elif dimMemberQname:
                     dimElt = XmlUtil.addChild(contextElt, XbrlConst.xbrldi, "xbrldi:explicitMember",
                                               attributes=dimAttr,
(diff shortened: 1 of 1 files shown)


arelle/ModelXbrl.py
```diff
@@ -698,7 +698,7 @@ def createContext(
                 if cast('DimValuePrototype | ModelDimensionValue', dimValue).isTyped:  #Typing thinks that this can also be a QName
                     dimElt = XmlUtil.addChild(contextElt, XbrlConst.xbrldi, "xbrldi:typedMember",
                                               attributes=dimAttr)
-                    if isinstance(dimValue, (arelle.ModelInstanceObject.ModelDimensionValue, DimValuePrototype)) and dimValue.isTyped:
+                    if isinstance(dimValue, (arelle.ModelInstanceObject.ModelDimensionValue, DimValuePrototype)) and dimValue.isTyped and dimValue.typedMember:
```
austinmatherne-wk: Thanks for the patch. CI is cranky because `ModelObject` inherits the [unintuitive](https://github.com/lxml/lxml/blob/64ed06c1a0c1833bfac99f209f16c3bdfddfde79/src/lxml/etree.pyx#L1352-L1363) `__bool__` implementation from the `etree._Element` base class which returns `True` based on if the XML element has children. Here and below need to use an explicit `dimValue.typedMember is not None` instead of depending on the result of `bool(ModelObject | None)`

freddym56: Thank you for your comment Austin, I've gone ahead and made the change.

austinmatherne-wk (dismissed):

austinmatherne-wk (approved):

PR #2357, opened 2026-05-04
Add retries to address shallow file conflict during merge

aaroncameron-wk:
#### Reason for change
Intermittent failures during merge in CI ([example](https://github.com/Arelle/Arelle/actions/runs/25341592128/job/74300022929?pr=2356)).

#### Description of change
The merge action's two sequential git fetch --deepen calls can race over .git/shallow, causing a fatal: shallow file has changed since we read it error. This is a transient concurrency issue — the first fetch modifies the shallow file, and the second fetch occasionally reads it mid-write.

Wrap the second git fetch --deepen in a retry loop (up to 3 attempts with a 1-second delay) so the action recovers automatically when the race occurs.

#### Steps to Test
Watch for merge failures after this merges.

**review**:
@Arelle/arelle

### DIFF
--- a/.github/actions/merge/action.yml
+++ b/.github/actions/merge/action.yml
@@ -29,7 +29,16 @@ runs:
         git fetch --deepen="$DEPTH" "https://github.com/$REPO.git" "$SHA"
 
         echo "Deepening base branch history to depth $DEPTH..."
-        git fetch --deepen="$DEPTH" origin
+        max_attempts=3
+        for ((i=1; i<=max_attempts; i++)); do
+          git fetch --deepen="$DEPTH" origin && break
+          if ((i == max_attempts)); then
+            echo "::error::Failed to deepen base branch history after $max_attempts attempts."
+            exit 1
+          fi
+          echo "Attempt $i failed, retrying..."
+          sleep 1
+        done
 
         git merge "$SHA" --no-edit || {
           echo "::error::Merge failed. Ancestor not found within $DEPTH commits or a conflict exists."


.github/actions/merge/action.yml
```diff
@@ -29,7 +29,16 @@ runs:
...
+        max_attempts=3
+        for ((i=1; i<=max_attempts; i++)); do
+          git fetch --deepen="$DEPTH" origin && break
+          if ((i == max_attempts)); then
+            echo "::error::Failed to deepen base branch history after $max_attempts attempts."
+            exit 1
+          fi
+          echo "Attempt $i failed, retrying..."
+          sleep 1
+        done
```
austinmatherne-wk, line 41: arbitrary line:

I think the changes here are reasonable regardless, but I'm a little perplexed by what's happening. The two `git fetch` calls are sequential invocations, so fetch 1 should be fully exited before fetch 2 starts. My only thought is that fetch 1 is triggering background git gc which is rewriting the shallow file while fetch 2 runs. Might be worth configuring git gc in this action to either not run or to not auto detach. If on the other hand this is actually an internal race condition issue in `git fetch --deepen`, we might find after merging this that we need to apply the retry to the first fetch, as well.

austinmatherne-wk (approved):

.github/actions/merge/action.yml
aaroncameron-wk: Let's table this PR and go with the GC route first ([PR](https://github.com/Arelle/Arelle/pull/2359))

austinmatherne-wk: Doesn't look like disabling gc did the trick 😞 

https://github.com/Arelle/Arelle/actions/runs/25420678108/job/74561914334?pr=2361

Interestingly this seems to only occur on the macOS 26 runners.

PR #2362, opened 2026-05-06
Close global Package/Plugin managers before Session runs

aaroncameron-wk:
#### Reason for change
Using `PluginManager.reset` rather than `close` before Session runs meant that the global `PluginManager.pluginConfig` could have existing data that bleeds into the Session run (though the Session does call `PluginManager.close` _after_ execution so this generally isn't a concern for subsequent Session runs).

#### Description of change
Call `PluginManager.close` rather than `reset` before Session runs. Do the same for `PackageManager`, though there was little effective difference between `PackageManager.close` and `reset`.

#### Steps to Test
CI

**review**:
@Arelle/arelle

### DIFF
--- a/arelle/plugin_system/_plugin_manager.py
+++ b/arelle/plugin_system/_plugin_manager.py
@@ -145,10 +145,7 @@ class PluginManager:
     def close(self) -> None:  # close all loaded methods
         if self.pluginConfig is not None:
             self.pluginConfig.clear()
-        if self.modulePluginInfos is not None:
-            self.modulePluginInfos.clear()
-        if self.pluginMethodsForClasses is not None:
-            self.pluginMethodsForClasses.clear()
+        self.reset()
 
     ''' pluginInfo structure:
 
--- a/arelle/api/Session.py
+++ b/arelle/api/Session.py
@@ -129,7 +129,7 @@ class Session:
         with _session_lock:
             self._check_thread()
             PackageManager.getInstance().close()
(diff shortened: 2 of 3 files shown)


arelle/plugin_system/_plugin_manager.py
```diff
@@ -145,10 +145,7 @@ def save(self, cntlr: Cntlr) -> None:
     def close(self) -> None:  # close all loaded methods
         if self.pluginConfig is not None:
             self.pluginConfig.clear()
-        if self.modulePluginInfos is not None:
-            self.modulePluginInfos.clear()
-        if self.pluginMethodsForClasses is not None:
-            self.pluginMethodsForClasses.clear()
+        self.reset()
```
austinmatherne-wk, line 148: preexisting: should `self._isInitialized` be set back to `False` on close and/or reset?

austinmatherne-wk (approved):

arelle/plugin_system/_plugin_manager.py
aaroncameron-wk: Good thought, but that flag mainly exists just to ensure that `init` has been called at least once to ensure that certain properties are set (e.g. `cntlr`). Obviously those values might have stale values if someone is trying to use `PluginManager` across multiple `Cntlr` contexts, but that was already the case.

derekgengenbacher-wf (approved):

derekgengenbacher-wf:

PR #2364, opened 2026-05-07
Type Hints: arelle.ModelDtsObject

volodymyrtur-wk:
#### Description of change

#### Steps to Test
CI
**review**:
@Arelle/arelle

### DIFF
--- a/arelle/ModelDtsObject.py
+++ b/arelle/ModelDtsObject.py
@@ -58,12 +58,16 @@ Relationship sets can identify ineffective arcroles, which is a requirement for 
 validation.
 """
 from __future__ import annotations
+
+from abc import ABC, abstractmethod
 from collections import defaultdict
-from collections.abc import Collection
+from collections.abc import Collection, Iterator
 import decimal
 import os
 import sys
-from typing import TYPE_CHECKING
+from typing import TYPE_CHECKING, Any, Iterable
+# for python 3.10
+from typing_extensions import Self
 from arelle import (
     Locale,
     ModelValue,
@@ -77,7 +81,12 @@ from arelle.ModelObject import ModelObject
 from arelle.typing import ModelFactBase, ModelResourceBase
 
 if TYPE_CHECKING:
+    from arelle.ModelDocument import ModelDocument
     from arelle.ModelValue import QName
+    from arelle.typing import TypeGetText
+
+_: TypeGetText
+
 
 class ModelRoleType(ModelObject):
     """
@@ -88,31 +97,35 @@ class ModelRoleType(ModelObject):
     :param modelDocument: owner document
     :type modelDocument: ModelDocument
     """
-    def init(self, modelDocument):
+    _definition: str | None
+    _usedOns: set[QName]
+    _tableCode: str | None
+
+    def init(self, modelDocument: ModelDocument) -> None:
         super(ModelRoleType, self).init(modelDocument)
 
     @property
-    def isArcrole(self):
+    def isArcrole(self) -> bool:
         """(bool) -- True if ModelRoleType declares an arcrole type"""
         return self.localName == "arcroleType"
 
     @property
-    def roleURI(self):
+    def roleURI(self) -> str | None:
         """(str) -- Value of roleURI attribute"""
         return self.get("roleURI")
 
     @property
-    def arcroleURI(self):
+    def arcroleURI(self) -> str | None:
         """(str) -- Value of arcroleURI attribute"""
         return self.get("arcroleURI")
 
     @property
-    def cyclesAllowed(self):
+    def cyclesAllowed(self) -> str | None:
         """(str) -- Value of cyclesAllowed attribute"""
         return self.get("cyclesAllowed")
 
     @property
-    def definition(self):
+    def definition(self) -> str | None:
         """(str) -- Text of child definition element (stripped)"""
         try:
             return self._definition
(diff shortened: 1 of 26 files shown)


arelle/ModelDtsObject.py
```diff
@@ -122,37 +135,37 @@ def definition(self):
...
     @property
-    def definitionNotStripped(self):
+    def definitionNotStripped(self) -> str | None:
         """(str) -- Text of child definition element (not stripped)"""
         definition = XmlUtil.child(self, XbrlConst.link, "definition")
         return definition.textValue if definition is not None else None
 
     @property
-    def usedOns(self):
+    def usedOns(self) -> set[QName]:
```
aaroncameron-wk: Open to discussion, but if our intent is that we can trust type hints when they are provided, then this should be `set[TypeXValue]` since we don't actually filter out non-QName values from this set. Someone using `usedOns` should be made aware that there's really no guarantee that every value is a `QName` and they should be checking themselves if it's important. If we've already established a pattern where we are more optimistic with our type hints then this is not as necessary.

arelle/XmlValidate.py
```diff
@@ -2,6 +2,8 @@
 See COPYRIGHT.md for copyright information.
 '''
 from __future__ import annotations
+
+import sys
```
austinmatherne-wk: Was this added by mistake?

arelle/ValidateXbrl.py
```diff
@@ -194,8 +195,9 @@ def validate(self, modelXbrl: ModelXbrl, parameters: dict[Any, Any] | None = Non
                         fromConcept = modelRel.fromModelObject
                         toConcept = modelRel.toModelObject
                         if fromConcept is not None and toConcept is not None:
+                            assert isinstance(weight, float), "weight is not an integer"
```
austinmatherne-wk: Previously an attribute error would've been raised below when `weight` was `None` and both from and to concepts had a balance. Now an assertion is raised if `weight` is `None`, regardless of balance. We should update the `if fromBalance and toBalance:` line to be `if fromBalance and toBalance and weight is not None:` instead of adding an assertion.

arelle/plugin/validate/ESEF/ESEF_Current/DTS.py
```diff
@@ -343,6 +345,7 @@ def checkFilingDTS(val: ValidateXbrl, modelDocument: ModelDocument, esefNotesCon
                 if linkEltName in ("definitionLink", ) and esefDisclosureSystemYear == 2023 and val.authParam["validate1_9_1"] in ("true", "True", 1):
                     for locElt in linkElt.iterchildren("{http://www.xbrl.org/2003/linkbase}loc"):
                         refObject = locElt.dereference()
+                        assert refObject.qname is not None, "refObject.qname is None"
                         if (isinstance(refObject, ModelConcept)
```
austinmatherne-wk: To be safe, the `assert` needs to be after the `isinstance(refObject, ModelConcept)` guard (`refObject` itself could be `None`)? Or better, bake it in to the if condition:

```python
if (isinstance(refObject, ModelConcept)
    and refObject.qname is not None
    and refObject.qname.namespaceURI in ifrsNses
    ...
```

arelle/ModelDtsObject.py
```diff
@@ -1547,37 +1692,23 @@ def allowsNamespace(self, namespaceURI):
             self._isAny = (not self._namespaces) or "##any" in self._namespaces
             return self.allowsNamespace(namespaceURI)
 
-class ModelAnyAttribute(ModelObject):
+class ModelAny(ModelAnyAttribute, ModelParticle):
```
austinmatherne-wk: I appreciate the desire to deduplicate `allowsNamespace`. However, `ModelAny` (`xs:any` wildcard particle) isn't semantically a subtype of `ModelAnyAttribute` (`xs:anyAttribute` attribute wildcard) in xml schema. With this change `isinstance(modelAny, ModelAnyAttribute)` would return `True`, which doesn't match how these map to distinct objects in the xml schema spec. A shared mixin for the namespace matching logic would give us the deduplication without the incorrect type relationship.

arelle/XmlValidate.py
volodymyrtur-wk: yes, forgot to clean it up

aaroncameron-wk (dismissed):

arelle/ModelDtsObject.py
```diff
@@ -1556,10 +1729,10 @@ class ModelAnyAttribute(ModelObject):
     :param modelDocument: owner document
     :type modelDocument: ModelDocument
     """
-    def init(self, modelDocument):
+    def init(self, modelDocument: ModelDocument) -> None:
         super(ModelAnyAttribute, self).init(modelDocument)
 
-    def allowsNamespace(self, namespaceURI):
+    def allowsNamespace(self, namespaceURI: str | None) -> bool:
```
austinmatherne-wk: It looks like both classes override the `AllowsNamespaceMixin` method, so it's duplicated 3 times now instead of 2. Did you intend to remove the implementations from `ModelAnyAttribute` and `ModelAny` after adding the mixin?

arelle/ModelDtsObject.py
```diff
@@ -185,36 +223,42 @@ class ModelNamableTerm(ModelObject):
...
         parent = self.getparent()
+        assert parent is not None
         return parent.namespaceURI == XbrlConst.xsd and parent.localName == "schema"
 
-    def schemaNameQname(self, prefixedName, isQualifiedForm=True, prefixException=None):
+    def schemaNameQname(
+        self,
+        prefixedName: str | None,
+        isQualifiedForm: bool = True,
+        prefixException: Exception | None = None,
```
austinmatherne-wk: The usage of this is a little confusing, but it takes both an exception or an exception type:
```suggestion
        prefixException: Exception | type[Exception] | None = None,
```

arelle/XmlValidate.py
```diff
@@ -718,7 +719,7 @@ def validateFacetValueString(facetName: str, facetValue: str, baseXsdType: str)
 def validateAnyWildcard(qnElt: QName, qnAttr: QName, attributeWildcards: list[ModelAny]) -> bool:
     # note wildcard is a set of possibly multiple values from inherited attribute groups
     for attributeWildcard in attributeWildcards:
-        if attributeWildcard.allowsNamespace(qnAttr.namespaceURI):  # type: ignore[no-untyped-call]
+        if attributeWildcard.allowsNamespace(qnAttr.namespaceURI):
```
austinmatherne-wk: Preexisting, but presumably the type for `attributeWildcards` is wrong and it should be `list[ModelAnyAttribute]`

arelle/ModelDtsObject.py
volodymyrtur-wk: yes, I missed them. Removed

arelle/XmlValidate.py
volodymyrtur-wk: Looks like it's true, at least when I use tests

austinmatherne-wk (approved):

derekgengenbacher-wf (approved):

aaroncameron-wk (approved):

derekgengenbacher-wf:

PR #2367, opened 2026-05-08
EDINET 2026

aaroncameron-wk:
#### Reason for change
EDINET 2026 was released on November 11, 2025. Our plugin only supports 2025.

Note: It's unclear how we will maintain the conformance suite year-to-year going forward, updating all of the invalid test cases for the current taxonomy year is not a minor effort. We would likely want to design an automated build process. In the meantime, adding the new year samples to the conformance suite feels like an acceptable solution.

#### Description of change
- Configured 2025/2026 disclosure systems
- Added 2025-11-01 taxonomy files to standard taxonomy file list
- Built/published a proper taxonomy package based off of https://www.fsa.go.jp/search/20251111/1c_Taxonomy.zip
- Added 2026 samples to conformance suite
- Made some fixes for validation issues revealed by samples 16, 17, 18 which did not have counterparts for 2025.

#### Steps to Test
CI

**review**:
@Arelle/arelle


arelle/plugin/validate/EDINET/resources/config.xml
```diff
@@ -4,7 +4,23 @@
         xsi:noNamespaceSchemaLocation="../../../../config/disclosuresystems.xsd">
     <!-- see arelle/config/disclosuresystems.xml for full comments -->
     <DisclosureSystem
-            names="EDINET|edinet"
+            names="EDINET-2026|edinet-2026"
```
austinmatherne-wk: In the ESEF plugin, disclosure system names include the year because it corresponds to a specific filer manual version. The taxonomy version alone isn't enough to determine which validation rules apply without the user explicitly choosing a manual version. Is something analogous happening here, or is it a clean one to one mapping of taxonomy year to rules that could be inferred from filing namespaces automatically?

Assuming we do want multiple disclosure systems, removing `EDINET`/`edinet` would be a disruptive change for existing users (I think we have at least 1). For consistency with ESEF, aliasing the bare name to the latest non-draft year seems like the right approach. When we've migrated the ESEF disclosure system name to a new year I've bumped the Arelle minor version number. We could do the same thing here if we think that makes sense.

arelle/plugin/validate/EDINET/resources/edinet-taxonomies.xml
```diff
@@ -59,4 +59,44 @@
...
+    <Loc><Family>FSA</Family><Version>2025</Version><Href>http://disclosure.edinet-fsa.go.jp/taxonomy/jppfs/2025-11-01/label/jppfs_2025-11-01_gla.xml</Href><AttType>LAB</AttType><FileTypeName>Labels</FileTypeName><Elements>0</Elements></Loc>
+    <Loc><Family>FSA</Family><Version>2025</Version><Href>http://disclosure.edinet-fsa.go.jp/taxonomy/jppfs/2025-11-01/label/jppfs_2025-11-01_lab-en.xml</Href><AttType>LAB</AttType><FileTypeName>Labels</FileTypeName><Elements>0</Elements></Loc>
+    <Loc><Family>FSA</Family><Version>2025</Version><Href>http://disclosure.edinet-fsa.go.jp/taxonomy/jppfs/2025-11-01/label/jppfs_2025-11-01_lab.xml</Href><AttType>LAB</AttType><FileTypeName>Labels</FileTypeName><Elements>0</Elements></Loc>
+    <Loc><Family>FSA</Family><Version>2025</Version><Href>http://disclosure.edinet-fsa.go.jp/taxonomy/jpsps/2025-11-01/jpsps_cor_2025-11-01.xsd</Href><AttType>SCH</AttType><FileTypeName>Schema</FileTypeName><Elements>1</Elements> <Namespace>http://disclosure.edinet-fsa.go.jp/taxonomy/jpsps/2025-11-01/jpsps_cor</Namespace><Prefix>jppfs</Prefix></Loc>
```
austinmatherne-wk: Preexisting from the prior taxonomy and I'm not sure it even matters, but this prefix looks wrong (also a couple other copy pasta `jppfs` prefixes).

arelle/plugin/validate/EDINET/NamespaceConfig.py
```diff
@@ -4,26 +4,33 @@
...
+
 
 class NamespaceConfig:
-    def __init__(self) -> None:
-        self.jpcrpEsr = "http://disclosure.edinet-fsa.go.jp/taxonomy/jpcrp-esr/2024-11-01/jpcrp-esr_cor"
-        self.jpcrp = "http://disclosure.edinet-fsa.go.jp/taxonomy/jpcrp/2024-11-01/jpcrp_cor"
-        self.jpcrpSbr = "http://disclosure.edinet-fsa.go.jp/taxonomy/jpcrp-sbr/2024-11-01/jpcrp-sbr_cor"
-        self.jpctl = "http://disclosure.edinet-fsa.go.jp/taxonomy/jpctl/2024-11-01/jpctl_cor"
+    def __init__(self, disclosureSystemName: str | None) -> None:
+        assert (disclosureSystemName is not None and
```
austinmatherne-wk: Declaring an optional type and then asserting not `None` is surprising behavior. Probably better to type as a string and require the caller to handle the not `None` check.

arelle/plugin/validate/EDINET/resources/cover-item-requirements.json
```diff
@@ -763,17 +729,11 @@
     "jptoo-tor_cor:TelephoneNumberCoverPage",
     "jptoo-tor_cor:NameOfContactPersonCoverPage",
     "jptoo-tor_cor:NameOfLegalRepresentativeCoverPage",
-    "jptoo-tor_cor:NameOfLegalRepresentativeCoverPageNA",
```
austinmatherne-wk, line 766: Were these NA entries not needed for 2025?

arelle/plugin/validate/EDINET/resources/config.xml
aaroncameron-wk: Looking at NL disclosure systems `NT16...20`, the only purpose they serve is to change the namespaces/entrypoints to use in validation, without actually affecting which rules apply. That's the existing design this was modeled after.

I'll add the `EDINET/edinet` aliases to the 2026 disclosure system, and I've added the `minor-release` label to this PR.

arelle/plugin/validate/EDINET/resources/cover-item-requirements.json
aaroncameron-wk: This data is inferred from [ESE140114.xlsx](https://disclosure2dl.edinet-fsa.go.jp/guide/static/disclosure/download/ESE140114.xlsx), which has these "NA" (not applicable) entries listed, but I don't believe they're meant to be interpreted as actual concepts (they don't appear to exist).

arelle/plugin/validate/EDINET/resources/config.xml
austinmatherne-wk: That feels like a mistake in NL that we probably copied from the ESEF pattern without considering how it was different. Requiring the user to know which taxonomy is used in order to validate creates friction for bulk processing and requires managing multiple configurations for not much benefit. You're deeper in this than I am, so I'll defer to you, but we may want to come back and review all the validation plugin disclosure systems and see if we can streamline the user experience a bit.

austinmatherne-wk (approved):

derekgengenbacher-wf (approved):

derekgengenbacher-wf:

PR #2371, opened 2026-05-11
Type Hints: arelle.ModelRssItem, arelle.ModelRssObject

volodymyrtur-wk:
#### Description of change
Those modules are excluded from mypy type checking. We should remove them from the exclusion list and add types so that they pass mypy checks.

#### Steps to Test
CI
**review**:
@Arelle/arelle

### DIFF
--- a/arelle/ModelRssItem.py
+++ b/arelle/ModelRssItem.py
@@ -262,4 +262,3 @@ class ModelRssItem(ModelObject):
 
     def __repr__(self) -> str:
         return ("rssItem[{0}]{1})".format(self.objectId(), self.propertyView))
-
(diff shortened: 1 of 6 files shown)


arelle/ModelRssItem.py
```diff
@@ -24,11 +48,33 @@
...
+    _url: str | None
+    _htmURLs: list[tuple[str | None, str | None]]
+    _primaryDocumentURL: str | None
+
+    def init(self, modelDocument: ModelDocument) -> None:
         super(ModelRssItem, self).init(modelDocument)
         try:
-            if (self.modelXbrl.modelManager.rssWatchOptions.latestPubDate and
-                self.pubDate <= self.modelXbrl.modelManager.rssWatchOptions.latestPubDate):
+            assert self.modelXbrl is not None
```
austinmatherne-wk: I'm not sure about this assertion. The prior attribute access is within a try/except `AttributeError` block so previously it was caught and set status to "not tested" (whatever that means), the new assertion raises an unhandled `AssertionError` instead.

derekgengenbacher-wf (approved):

austinmatherne-wk (approved):

PR #2376, opened 2026-05-14
Type Hints: arelle.HashUtil 

volodymyrtur-wk:
#### Description of change

#### Steps to Test
CI
**review**:
@Arelle/arelle

### DIFF
--- a/arelle/HashUtil.py
+++ b/arelle/HashUtil.py
@@ -1,79 +1,94 @@
 '''
 See COPYRIGHT.md for copyright information.
 '''
-from hashlib import md5
-from arelle.ModelObject import ModelObject
-from arelle.ModelValue import QName, DateTime
+from __future__ import annotations
+
 from datetime import date, datetime
+from hashlib import md5
+from typing import Iterable, Union, TypeAlias
+
 from arelle import XmlUtil
+from arelle.ModelObject import ModelObject
+from arelle.ModelValue import QName
+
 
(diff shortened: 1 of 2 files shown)


aaroncameron-wk (dismissed):

arelle/HashUtil.py
```diff
@@ -1,79 +1,94 @@
...
         if not isinstance(other, Md5Sum):
+            if not isinstance(other, (int, str)):
+                return NotImplemented
             other = Md5Sum(other)
         return self.value == other.value
 
-    def __ne__(self, other):
-        return not (self.value == other.value)
+    def __ne__(self, other: object) -> bool:
+        return not self.__eq__(other)
```
austinmatherne-wk, line 47: This looks wrong. When `self.__eq__` returns `NotImplemented` this evaluates as `return not NotImplemented` which should throw a `TypeError` on Python 3.14+. You should be able to delete this method and use the default python `__ne__` implementation built off of `__eq__`.

volodymyrtur-wk: It shouldn't `return NotImplemented`, I meant `raise NotImplementedError`. Probably, this extra condition is unnecessary, because we handle this in the `__init__` method, so I'll remove it

austinmatherne-wk (approved):

derekgengenbacher-wf (approved):

derekgengenbacher-wf:

PR #2377, opened 2026-05-15
Type Hints: arelle.PrototypeDtsObject 

volodymyrtur-wk:
#### Description of change

#### Steps to Test
CI
**review**:
@Arelle/arelle

### DIFF
--- a/arelle/PrototypeDtsObject.py
+++ b/arelle/PrototypeDtsObject.py
@@ -40,7 +40,7 @@ class PrototypeObject(PrototypeObjectBase):
     def get(self, key: str, default: str | None = None) -> str | None:
         return self.attributes.get(key, default)
 
-    def itersiblings(self, **kwargs: Any) -> Iterator[ModelObject]:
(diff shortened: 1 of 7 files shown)


arelle/PrototypeDtsObject.py
```diff
@@ -1,200 +1,271 @@
...
+        self.qname: QName = qname
+        self.elementQname: QName = qname
+        self.namespaceURI: str | None = qname.namespaceURI
+        self.localName: str = qname.localName
+        self.role: str | None = role
         # children are arc and loc elements or prototypes
-        self.childElements = []
-        self.text = self.textValue = None
-        self.attributes = {"{http://www.w3.org/1999/xlink}type":"extended"}
+        self.childElements: list[ModelObject] = []
```
austinmatherne-wk: This also includes `PrototypeObject`?

derekgengenbacher-wf (approved):

derekgengenbacher-wf:

PR #2378, opened 2026-05-15
Refactor locale handling and improve test coverage

strr:
#### Reason for change
Fetching the list of supported system locales was not supported on Windows

#### Description of change

- Add support for fetching system locales on Windows
- As part of this refactor findCompatibleLocale, _candidatePosixLocales (and split out _compatibleSystemLocales) with the aim that no desired behaviour has changed and various undesired bits (e.g. sorting language codes in a non-delimiter aware way) removed
- Introduce class `_LocaleCode` and use it to drive all localeName and langTag string parsing, validation and conversion (BCP47 or POSIX) within the module (this is mostly centralising of the best variations of existing code)
- Introduce availableBCP47LangTags to make calling code simpler in DialogLanguage.py
- Add tests to pin down expected behaviour

#### Steps to Test

New unit tests work
Language list still appears as expected in the GUI drop-down.

**review**:
@Arelle/arelle


arelle/Locale.py
```diff
@@ -102,133 +201,137 @@ def getLanguageCodes(configLang: str | None = None) -> list[str]:
...
+    matches: set[_LocaleCode] = {
+        lc_sys
+        for lc_sys in _getSystem_LocaleCodes()
+        if lc_sys not in exclude and lc_sys.lang == lc.lang
+    }
+
+    def _sortKey(candidate: _LocaleCode) -> tuple[bool, bool, str]:
+        region_match = lc.region is None or candidate.region == lc.region
+        encoding_match = (candidate.encoding is None and lc.encoding is None) or \
+                     (candidate.encoding == (lc.encoding or POSIX_LOCALE_DEFAULT_ENCODING))
```
austinmatherne-wk: Should the encoding comparison here be lowercase normalized, the same way the assignment expression for `not_default_encoding` is?

strr: Yes, you're right. 22b2456 introduces `_LocaleCode.encoding_lc` property and uses it to simplify this code and `not_default_encoding` itself.

austinmatherne-wk (approved):

PR #2379, opened 2026-05-15
Update documentation on prefixedNamespaces and include more prefixes

strr:
#### Reason for change
`ModelXbrl.prefixedNamespaces` is list of the prefixes and namespaces used in a report and taxonomy. It is built from the self-declared prefixes in every schema document in a DTS.

#### Description of change

- Add documentation to note the incomplete nature of prefixedNamespaces:
  - Two or more namespaces requesting the same prefix: first one wins;
  - Prefix used on a given QName can be different to the relevant namespace's self-declared prefix
- Add all the namespace prefix bindings found on the root document as long as they don't conflict with those found already as it's a relatively cheap way to fill in a few more gaps.
- Convert to `cached_property` 

#### Steps to Test

Should just get the same dict as before when used with a sample document or the same + additional entries

**review**:
@Arelle/arelle

### DIFF
--- a/arelle/ModelXbrl.py
+++ b/arelle/ModelXbrl.py
@@ -3,18 +3,16 @@ See COPYRIGHT.md for copyright information.
 '''
 from __future__ import annotations
 
-from functools import cached_property
 import logging
 import os
 import sys
 import traceback
-from types import MappingProxyType
 import uuid
 from collections import defaultdict
 from typing import TYPE_CHECKING, Any, TypeVar, Union, cast, Optional
 
 import regex as re
-from collections.abc import Iterable, Iterator, Mapping
+from collections.abc import Iterable, Iterator
 
 import arelle
 from arelle import FileSource, ModelRelationshipSet, XmlUtil, ModelValue, XbrlConst, XmlValidate
--- a/arelle/ModelXbrl.py
+++ b/arelle/ModelXbrl.py
@@ -3,6 +3,7 @@ See COPYRIGHT.md for copyright information.
 '''
 from __future__ import annotations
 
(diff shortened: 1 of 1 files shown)


arelle/ModelXbrl.py
```diff
@@ -560,18 +561,38 @@ def saveInstance(self, **kwargs: Any) -> Any:
         assert self.modelDocument is not None
         self.modelDocument.save(**kwargs)
 
-    @property
+    @cached_property
```
austinmatherne-wk: I think this is going to be problematic. This returns a mutable dict that at least one [caller mutates](https://github.com/Arelle/Arelle/blob/2c3f6d2bc78dbcf1ed3e9ef6100aecabb73c76af/arelle/plugin/xbrlDB/XbrlOpenSqlDB.py#L1261-L1264). Currently each caller gets a fresh dict, but now everyone would get a modified dict if any caller mutates it. Also, [there's a reload method](https://github.com/Arelle/Arelle/blob/2c3f6d2bc78dbcf1ed3e9ef6100aecabb73c76af/arelle/ModelXbrl.py#L404) that calls init and clears `namespaceDocs` that this it built from, but doesn't reset the `prefixedNamespaces` cached property.

strr: Yuck! But I have a plan 😂

strr: I've just reverted it to `@property` in 3058b60 as it should not be called more than once per well written plugin.

I don't think it is worth worrying about the complexity of caching given the existing mutability and reload support required.

austinmatherne-wk (approved):

PR #2381, opened 2026-05-15
Split inline utils from inline validate and break import cycle

strr:
#### Reason for change
XmlUtil.py and XhtmlValidate.py depend on each other. This makes them hard to read due to the various tricks used to defer imports so as to work around the cycle. 

#### Description of change

- Disentangle XhtmlValidate.py, XmlUtil.py and friends by introducing XhtmlInlineUtil.py
  - Clean up the code moved across from XhtmlValidate.py to XhtmlInlineUtil.py
- Add tests
- Improve the performance of XmlUtil's escapedText() and selfCloseable() as they were easy wins

#### Steps to Test

**review**:
@Arelle/arelle


arelle/XhtmlValidate.py
```diff
@@ -108,16 +48,5 @@ def xhtmlValidate(modelXbrl: ModelXbrl, elt: ModelObject) -> None:
         ValidateFilingText.validateHtmlContent(modelXbrl, elt, elt, "InlineXBRL", valHtmlContentMsgPrefix, isInline=True)
 
 
-def containsNamespacedElements(elt: etree.ElementBase, namespace: str) -> bool:
+def _containsNamespacedElements(elt: etree.ElementBase, namespace: str) -> bool:
```
austinmatherne-wk: I doubt external code uses this function, but renaming to private is still a breaking change. Private would be the right call if we were writing it fresh, but I'm not sure there's enough upside to justify the rename right now.

arelle/XhtmlInlineUtil.py
```diff
@@ -0,0 +1,118 @@
...
+htmlEltUriAttrs: dict[str, set[str]] = { # attributes with URI content (for relative correction and %20 canonicalization
+    "a": {"href"},
+    "area": {"href"},
+    "blockquote": {"cite"},
+    "del": {"cite"},
+    "form": {"action"},
+    "input": {"src", "usemap"},
+    "ins": {"cite"},
+    "img": {"src", "longdesc", "usemap"},
+    "object": {"codebase", "classid", "data", "archive", "usemap"}, # codebase must be first to resolve others
```
austinmatherne-wk: > codebase must be first to resolve others

This comment was copied over, but the collection it's referring to was changed from an ordered tuple to unordered set, so it's no longer true. However, this was never actually depended upon (`escapedNode()` explicitly checks `codebase` first before iterating this collection). The unordered set is an improvement, but it would be good to remove the stale comment to avoid confusion.

austinmatherne-wk (commented):
Lots of great improvements here. Thank you!

arelle/XhtmlValidate.py
strr: I agree. Done in 6a232a7

arelle/XhtmlInlineUtil.py
strr: Removed unhelpful comment in 7cb7ffa.

austinmatherne-wk (approved):

PR #2383, opened 2026-05-19
Type Hints: arelle.ValidateXbrlDTS 

volodymyrtur-wk:
#### Description of change

#### Steps to Test
CI
**review**:
@Arelle/arelle

### DIFF
--- a/arelle/ValidateXbrlDTS.py
+++ b/arelle/ValidateXbrlDTS.py
@@ -1159,7 +1159,7 @@ def checkElements(val: ValidateXbrl, modelDocument: ModelDocument, parent: _Elem
                                     conceptTo=arcToConceptQname(elt))
                             try:
                                 weightAttr = elt.get("weight")
-                                if float(weightAttr) in (1, -1):  # type: ignore[operator,arg-type]
+                                if not float(weightAttr) in (1, -1):  # type: ignore[operator,arg-type]
                                     val.modelXbrl.error(("EFM.6.14.02", "GFM.1.07.02"),
                                         _("CalculationArc from %(xlinkFrom)s to %(xlinkTo)s weight %(weight)s must be 1 or -1"),
                                         modelObject=elt,
--- a/arelle/ValidateXbrlDTS.py
+++ b/arelle/ValidateXbrlDTS.py
@@ -189,19 +189,19 @@ def checkDTS(val: ValidateXbrl, modelDocument: ModelDocument, checkedModelDocume
                             elementHref=hrefElt.get("{http://www.w3.org/1999/xlink}href"))
             # check href'ed target if a linkbaseRef
             if hrefElt.namespaceURI == XbrlConst.link:
-                if hrefedElt is not None and hrefElt.localName == "linkbaseRef":
+                if hrefElt.localName == "linkbaseRef":
                     # check linkbaseRef target
                     if (hrefedDoc is None or
                         hrefedDoc.type < ModelDocumentType.firstXBRLtype or  # range of doc types that can have linkbase
                         hrefedDoc.type > ModelDocumentType.lastXBRLtype or
-                        hrefedElt.namespaceURI != XbrlConst.link or hrefedElt.localName != "linkbase"):
+                        hrefedElt.namespaceURI != XbrlConst.link or hrefedElt.localName != "linkbase"):  # type: ignore[union-attr]
                         val.modelXbrl.error("xbrl.4.3.2:linkbaseRefHref",
                             _("LinkbaseRef %(linkbaseHref)s does not identify an link:linkbase element"),
                             modelObject=(hrefElt, hrefedDoc),
                             linkbaseHref=hrefElt.get("{http://www.w3.org/1999/xlink}href"))
                     elif hrefElt.get("{http://www.w3.org/1999/xlink}role") is not None:
                         role = hrefElt.get("{http://www.w3.org/1999/xlink}role")
-                        for linkNode in hrefedElt.iterchildren():
+                        for linkNode in hrefedElt.iterchildren():  # type: ignore[union-attr]
                             if (isinstance(linkNode,ModelObject) and
                                 linkNode.get("{http://www.w3.org/1999/xlink}type") == "extended"):
                                 ln = linkNode.localName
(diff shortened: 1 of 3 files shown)


aaroncameron-wk (dismissed):

arelle/ValidateXbrlDTS.py
```diff
@@ -163,6 +172,7 @@ def checkDTS(val: ValidateXbrl, modelDocument: ModelDocument, checkedModelDocume
                 else:
                     hrefedElt = hrefedDoc.xmlRootElement
 
+            assert hrefedElt is not None, "hrefedElt is None"
```
aaroncameron-wk: Is this the correct assertion? It doesn't seem necessary in this specific spot, could be the cause of the conformance suite failures.

arelle/ValidateXbrlDTS.py
```diff
@@ -396,6 +409,7 @@ def linkbaseTopElts():
                             _("Element %(element)s attribute %(attribute)s '%(value)s' is not an NCname"),
                             modelObject=elt, element=elt.prefixedName, attribute=name, value=attrValue)
                     '''
+                    assert _valueItems is not None, "_valueItems is None"
```
aaroncameron-wk: Seems like this could be within the `if` block below.

arelle/ValidateXbrlDTS.py
```diff
@@ -180,7 +189,7 @@ def checkDTS(val: ValidateXbrl, modelDocument: ModelDocument, checkedModelDocume
                             elementHref=hrefElt.get("{http://www.w3.org/1999/xlink}href"))
             # check href'ed target if a linkbaseRef
             if hrefElt.namespaceURI == XbrlConst.link:
-                if hrefElt.localName == "linkbaseRef":
+                if hrefedElt is not None and hrefElt.localName == "linkbaseRef":
```
austinmatherne-wk: The `hrefedElt is not None` checks here and below are functional changes. Previously if `hrefedDoc` and `hrefedElt` were both `None` (when `hrefedDoc` is `None`, `hrefedElt` is also always `None`), then `xbrl.4.3.2` would be raised (attribute access on `hrefedElt` was short circuited by the `hrefedDoc` check). Now it's skipped.

arelle/ValidateXbrlDTS.py
```diff
@@ -1106,24 +1116,23 @@ def linkbaseTopElts():
...
+                             }.get(val.extendedElementName,(elt.qname,)):  # type: ignore[arg-type] # allow non-2.1 to be ok regardless per RH 2013-03-13
                             val.modelXbrl.error("SBR.NL.2.3.0.11",
                                 _("Resource element %(element)s may not be contained in a linkbase with %(element2)s"),
                                 modelObject=elt, element=elt.qname, element2=val.extendedElementName)
                 if xlinkType == "arc":
                     if elt.get("priority") is not None:
                         priority = elt.get("priority")
                         try:
-                            if int(priority) >= 10:
+                            if priority and int(priority) >= 10:
```
austinmatherne-wk: Some of these guards (`priority` and `weightAttr`) look like they may cause the `ValueError` exception handling to be skipped on empty strings.

arelle/ValidateXbrlDTS.py
```diff
@@ -1151,8 +1160,7 @@ def linkbaseTopElts():
                                     conceptTo=arcToConceptQname(elt))
                             try:
                                 weightAttr = elt.get("weight")
-                                weight = float(weightAttr)
-                                if not weight in (1, -1):
+                                if weightAttr and not float(weightAttr) in (1, -1):
```
austinmatherne-wk: Similar to priority before, the `weightAttr` guard here is bypassing the `ValueError` except handling that logs an error.

arelle/ValidateXbrlDTS.py
```diff
@@ -212,7 +221,7 @@ def checkDTS(val: ValidateXbrl, modelDocument: ModelDocument, checkedModelDocume
                                         modelObject=hrefElt,
                                         linkbaseHref=hrefElt.get("{http://www.w3.org/1999/xlink}href"),
                                         role=role, link=linkNode.prefixedName)
-                elif hrefElt.localName == "schemaRef":
+                elif hrefedElt is not None and hrefElt.localName == "schemaRef":
```
austinmatherne-wk: This also looks like a functional change. Previously when `hrefElt.localName == "schemaRef"` and `hrefedDoc.type != ModelDocumentType.SCHEMA` would short circuit the `hrefedElt` usage and log an error. Now it falls into the other elif branches.

arelle/ValidateXbrlDTS.py
austinmatherne-wk: It doesn't look like this has been addressed yet

derekgengenbacher-wf (dismissed):

derekgengenbacher-wf (dismissed):

arelle/ValidateXbrlDTS.py
```diff
@@ -1151,8 +1159,7 @@ def linkbaseTopElts():
                                     conceptTo=arcToConceptQname(elt))
                             try:
                                 weightAttr = elt.get("weight")
-                                weight = float(weightAttr)
-                                if not weight in (1, -1):
+                                if float(weightAttr) in (1, -1):  # type: ignore[operator,arg-type]
```
aaroncameron-wk: Looks like the polarity of this statement got flipped.

derekgengenbacher-wf (approved):

aaroncameron-wk (approved):

derekgengenbacher-wf:

pyproject.toml
```diff
@@ -176,6 +176,7 @@ module = [
     'arelle.UITkTable',
     'arelle.ValidateInfoset',
     'arelle.ValidateVersReport',
+    'arelle.ValidateXbrlCalcs',
```
austinmatherne-wk, line 179: Why was `arelle.ValidateXbrlCalcs` added to the exclusion list? Looks like something got crossed up here and `arelle.ValidateXbrlDTS` wasn't removed

volodymyrtur-wk: Looks like something went wrong while merging a bunch of PRs. I'll clean them up in the next PR

PR #2384, opened 2026-05-21
Fix ValueError found via INFO-SEMANTIC calls

derekgengenbacher-wf:
#### Reason for change
ParsePort shared with us an error they were receiving via the EDINET validations.

#### Description of change
Fixed the call of level

#### Steps to Test
CI

**review**:
@Arelle/arelle

### DIFF
--- a/arelle/Cntlr.py
+++ b/arelle/Cntlr.py
@@ -327,7 +327,7 @@ class Cntlr:
     def validation(self, val: Validation, fileSource: FileSource | None = None) -> None:
         """Same as `error`, but parameters passed in from Validation object
         """
-        self.error(codes=val.codes, msg=val.msg, level=val.level.name, fileSource=fileSource, **val.args)
+        self.error(codes=val.codes, msg=val.msg, level=val.level.value, fileSource=fileSource, **val.args)
 
     def error(self, codes: Any, msg: str, level: str = "ERROR", fileSource: FileSource | None = None, **args: Any) -> None:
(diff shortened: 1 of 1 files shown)


lancestaley-wk (approved):

arelle/Cntlr.py
```diff
@@ -327,7 +327,7 @@ def __init__(
     def validation(self, val: Validation, fileSource: FileSource | None = None) -> None:
         """Same as `error`, but parameters passed in from Validation object
         """
-        self.error(codes=val.codes, msg=val.msg, level=val.level.name, fileSource=fileSource, **val.args)
+        self.error(codes=val.codes, msg=val.msg, level=val.level.value, fileSource=fileSource, **val.args)
```
austinmatherne-wk, line 330: Looks like we have the same issue in [ModelXbrl](https://github.com/Arelle/Arelle/blob/e587765be14afb62c88d104521c5b34f7d9b859a/arelle/ModelXbrl.py#L1070) and [ValidateFileSource](https://github.com/Arelle/Arelle/blob/e587765be14afb62c88d104521c5b34f7d9b859a/arelle/ValidateFileSource.py#L30). Can you fix those, too?

austinmatherne-wk (approved):

arelle/Cntlr.py
aaroncameron-wk: https://github.com/Arelle/Arelle/pull/2385

PR #2386, opened 2026-05-25
Type Hints: arelle.ValidateXbrlCalcs

volodymyrtur-wk:
#### Description of change

#### Steps to Test
CI
**review**:
@Arelle/arelle


austinmatherne-wk (approved):

derekgengenbacher-wf (approved):

derekgengenbacher-wf:

PR #2391, opened 2026-05-26
Type Hints: arelle.ValidateXbrlDimensions 

volodymyrtur-wk:
#### Description of change

#### Steps to Test
CI
**review**:
@Arelle/arelle

### DIFF
--- a/arelle/ValidateXbrlDimensions.py
+++ b/arelle/ValidateXbrlDimensions.py
@@ -1,28 +1,48 @@
 '''
 See COPYRIGHT.md for copyright information.
 '''
-import os, sys
+from __future__ import annotations
+
+import os
+import sys
 from collections import defaultdict
-from arelle import (UrlUtil, XbrlConst)
-from arelle.ModelObject import ModelObject
+from typing import TYPE_CHECKING
+
+from arelle import UrlUtil, XbrlConst
 from arelle.ModelDtsObject import ModelConcept
+from arelle.ModelObject import ModelObject
 from arelle.PrototypeInstanceObject import ContextPrototype, DimValuePrototype
-from typing import cast
+from arelle.typing import TypeGetText
+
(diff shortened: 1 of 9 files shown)


derekgengenbacher-wf (approved):

arelle/ValidateXbrlDimensions.py
```diff
@@ -160,53 +180,79 @@ def checkBaseSet(val, arcrole, ELR, relsSet) -> None:
                                 _("Domain-Member relationship from %(source)s to %(target)s in link role %(linkrole)s must have a domain primary item or domain member target"),
                                 modelObject=domMbrRel, source=priItemConcept.qname, target=toConcept.qname, linkrole=ELR)
 
-def domainTargetRoles(val, fromELR, rels, fromConcepts=None, ELRs=None):
+
+def domainTargetRoles(
+    val: ModelXbrl,
+    fromELR: str,
+    rels: list[ModelRelationship],
+    fromConcepts: set[ModelObject] | None = None,
```
aaroncameron-wk, line 188: Have you tried `ModelConcept` here (and elsewhere, "primary items", "members")? I'm curious what is preventing us from being able to more strictly hint these as concepts as the name suggests.

volodymyrtur-wk: Technically, it must be `ModelConcept`, but we take it from `ModelRelationship.toModelObject`, and it might contain different subclasses of `ModelObject`

aaroncameron-wk: 👍 In the future we could potentially filter out `toModelObject` values that are not `ModelConcept`, but I'd agree that's out of scope for a type hint PR.

aaroncameron-wk (approved):

derekgengenbacher-wf:

PR #2398, opened 2026-05-29
Remove log handlers on close.

stevenbronson-wk:
**review**:
@Arelle/arelle

### DIFF
--- a/arelle/Cntlr.py
+++ b/arelle/Cntlr.py
@@ -18,6 +18,7 @@ import os
 import subprocess
 import sys
 import tempfile
+import traceback
 from typing import TYPE_CHECKING, Any, cast, Union, Literal
 
 import regex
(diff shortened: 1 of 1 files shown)


arelle/Cntlr.py
```diff
@@ -549,7 +550,8 @@ def close(self, saveConfig: bool = False) -> None:
             try:
                 self.logHandler.close()
             except Exception: # fails on some earlier pythons (3.1)
-                pass
+                traceback.print_exc()
```
brettkail-wk, line 553: That's a rather old Python version referenced :-).  Is your thinking that the `try`/`except` is still useful, but it's better to log in case it happens to fail on more recent?

brettkail-wk (approved):

arelle/Cntlr.py
stevenbronson-wk: A log handler failure to close seems plausible (logging to the network?) and arguably not worth raising an exception, but I'm fine with removing try/except if people prefer.

brettkail-wk: I have no opinion on whether to keep swallowing (vs moving the `removeHandler` up, or `try`/`finally`).  If we want to keep the `except`+swallow, I suppose the comment could clarify the intent.

aaroncameron-wk (approved):

PR #2408, opened 2026-06-08
Table Constraints Restriction Validation

austinmatherne-wk:
#### Description of change

1000+ test case variations were introduced in an updated conformance suite. These tests uncovered a number of limitations in the existing xml value validation code. The enum test cases are intentionally not handled by this PR. They require potentially more controversial changes (such as how to handle valid XML date values outside the Python datetime min/max value range) and I don't won't to bury those changes in a large PR.

#### Steps to Test
* CI

**review**:
@Arelle/arelle


arelle/oim/_tc/metadata/validate.py
```diff
@@ -89,3 +92,16 @@ def _validate_column_order(self) -> Generator[TCMetadataValidationError, None, N
...
                     ),
                 )
+
+    def _validate_value_constraints(self) -> Generator[TCMetadataValidationError, None, None]:
+        for template_id, tc in self._tc_metadata.template_constraints.items():
+            for col_name, constraint in tc.constraints.items():
+                for error in validate_value_constraint(constraint, self._namespaces):
+                    error.prepend_path(
+                        TABLE_TEMPLATES_KEY, template_id, COLUMNS_KEY, col_name, TC_CONSTRAINTS_PROPERTY_NAME
+                    )
```
volodymyrtur-wk, line 102:
```suggestion
                    error.prepend_path(
                        TABLE_TEMPLATES_KEY, 
                        template_id, 
                        COLUMNS_KEY, 
                        col_name, 
                        TC_CONSTRAINTS_PROPERTY_NAME,
                    )
```

arelle/oim/_tc/metadata/value_constraint_validation.py
```diff
@@ -0,0 +1,290 @@
...
+        "half",
+        "quarter",
+        "week",
+        "month",
+        "day",
+        "instant",
+    }
+)
+
+
```
volodymyrtur-wk: Too many blank lines

arelle/oim/_tc/metadata/value_constraint_validation.py
```diff
@@ -0,0 +1,290 @@
...
+    min_exc_result = _parse_bounds_facet_value(TCRestriction.MIN_EXCLUSIVE, min_exclusive, base_xsd_type)
+    max_inc_result = _parse_bounds_facet_value(TCRestriction.MAX_INCLUSIVE, max_inclusive, base_xsd_type)
+    max_exc_result = _parse_bounds_facet_value(TCRestriction.MAX_EXCLUSIVE, max_exclusive, base_xsd_type)
+
+    restriction_parsed_values = [
+        (TCRestriction.MIN_INCLUSIVE, min_inclusive, min_inc_result),
+        (TCRestriction.MAX_INCLUSIVE, max_inclusive, max_inc_result),
+        (TCRestriction.MIN_EXCLUSIVE, min_exclusive, min_exc_result),
+        (TCRestriction.MAX_EXCLUSIVE, max_exclusive, max_exc_result),
+    ]
```
volodymyrtur-wk: In this case, a tuple of tuples suits better than a list of tuples

arelle/oim/_tc/metadata/value_constraint_validation.py
```diff
@@ -0,0 +1,290 @@
...
+
+    min_inc = _comparable_value(min_inc_result)
+    min_exc = _comparable_value(min_exc_result)
+    max_inc = _comparable_value(max_inc_result)
+    max_exc = _comparable_value(max_exc_result)
+
+    if min_inc is not None and max_inc is not None and min_inc > max_inc:
+        yield _bounds_ordering_error(
+            TCRestriction.MIN_INCLUSIVE, min_inclusive, "<=", TCRestriction.MAX_INCLUSIVE, max_inclusive
+        )
```
volodymyrtur-wk, line 220: When you drag the function parameters to the next line, it would be nice to split them line by line. You can put the comma after the last one to prevent changes when using the code formatter

austinmatherne-wk: This formatting was intentional. To my eye, keeping the parameters on one line makes it easier to compare the four different if statements against each other.

arelle/oim/_tc/metadata/validate.py
austinmatherne-wk: This is subjective, so it could just be me, but I don't think multiline is easier to read for paths like this. Reading left to right is marginally better? It's similar to writing something like this:

```
Users, username, Downloads, file
```
vs this:
```
Users,
username,
Downloads,
file
```

derekgengenbacher-wf (approved):

volodymyrtur-wk (approved):

arelle/XmlValidate.py
```diff
@@ -36,6 +36,21 @@
 cNameChar = r"[_\-\.:"   "\xB7A-Za-z0-9\xC0-\xD6\xD8-\xF6\xF8-\xFF\u0100-\u02FF\u0370-\u037D\u037F-\u1FFF\u200C-\u200D\u2070-\u218F\u2C00-\u2FEF\u3001-\uD7FF\uF900-\uFDCF\uFDF0-\uFFFD\u0300-\u036F\u203F-\u2040]"
 cMinusCNameChar = r"[_\-\."   "\xB7A-Za-z0-9\xC0-\xD6\xD8-\xF6\xF8-\xFF\u0100-\u02FF\u0370-\u037D\u037F-\u1FFF\u200C-\u200D\u2070-\u218F\u2C00-\u2FEF\u3001-\uD7FF\uF900-\uFDCF\uFDF0-\uFFFD\u0300-\u036F\u203F-\u2040]"
 
+def _raiseOnNonXsdRegexSyntax(p: str) -> None:
```
brettkail-wk, line 39: This seems like very narrow restriction based on a specific test failure.  I know that `XmlValidate.py` has `XsdPattern` that handles other cases too, so it seems like it would be nice to somehow combine them.  Well, both are a bit ad hoc; it seems like at some point we should have a full XSD pattern parser that converts to Python syntax.

austinmatherne-wk: > This seems like very narrow restriction based on a specific test failure.

It is. I suspect that the test cases were added because the author knew Arelle was incorrectly accepting the patterns. If you're aware of additional pattern syntax we should reject, I'm happy to cover it.

> I know that XmlValidate.py has XsdPattern that handles other cases too, so it seems like it would be nice to somehow combine them

Maybe I'm misunderstanding what you mean by combine them, but this is called a few lines below in `XsdPattern.compile`, right before `regex.compile` is called on the pattern (which is used to raise `ValueError` on invalid patterns). I just chose to extract it as a named function instead of inlining.

> it seems like at some point we should have a full XSD pattern parser that converts to Python syntax.

Agreed. I'll create a ticket (#2413)

brettkail-wk: > If you're aware of additional pattern syntax we should reject, I'm happy to cover it.

I'm not very familiar with Python RE.  An LLM can probably enumerate better than me the syntax that Python supports that XSD doesn't :-).

> Maybe I'm misunderstanding what you mean by combine them

Sorry, I got lost in the PR and thought `_raiseOnNonXsdRegexSyntax` was added to (and called by) TC-specific code rather than the core `XmlValidate.py`.  The change you made looks fine.

allenbross-wf:
QA +10
All tests passed

PR #2414, opened 2026-06-10
Improve ix11.10.1.2 negative number detection for iXBRL 1.1

aaroncameron-wk:
This PR fixes false positives in the ix11.10.1.2 validation rule for iXBRL 1.1 documents during DTS validation.

## Problem

The existing check for negative values in `ix:nonFraction` and `ix:numerator`/`ix:denominator` elements (when no `format` attribute is present) used a simple `'-' in text` check on the inner text content. This produced false positives for non-numeric text that happened to contain a hyphen (e.g., date strings like "2023-01-01" or hyphenated words).

## Changes

- **arelle/ValidateXbrlDTS.py**: Replaced the naive hyphen-contains check with a new `_isNegativeDecimal` helper that uses `Decimal` parsing to determine if the raw text value is actually a negative number. This eliminates false positives on non-numeric text containing hyphens while still correctly detecting genuinely negative numeric values.

- **tests/unit_tests/arelle/test_validatexbrldts.py**: Added unit tests for the `_isNegativeDecimal` helper covering:
  - Negative numeric strings that should return True (e.g., "-1", "-0.5")
  - Non-negative numeric strings that should return False (e.g., "0", "1", "100")
  - Non-numeric text containing hyphens that should return False — the false positive cases this PR fixes (e.g., "2023-01-01", "hello-world")

## Note

This check only applies to the raw inner text of elements that have **no** `format` attribute. Per the iXBRL spec, when a `format` attribute is present, the transformation handles sign interpretation and this check is skipped entirely.

### DIFF
--- a/arelle/ValidateXbrlDTS.py
+++ b/arelle/ValidateXbrlDTS.py
@@ -1453,7 +1453,7 @@ def checkIxContinuationChain(val: ValidateXbrl, elt: ModelObject, chain: list[Mo
 def _isNegativeDecimal(text: str) -> bool:
     """Return True if text is a valid decimal number and is negative."""
     try:
-        return Decimal(text.strip()) < 0
+        return Decimal(text) < 0
     except (InvalidOperation, ValueError):
         return False
 
--- a/arelle/ValidateXbrlDTS.py
+++ b/arelle/ValidateXbrlDTS.py
@@ -2,6 +2,7 @@
 See COPYRIGHT.md for copyright information.
 '''
 from __future__ import annotations
+from decimal import Decimal, InvalidOperation
 from enum import Enum
 from typing import TYPE_CHECKING, Any, Iterable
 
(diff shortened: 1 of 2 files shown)


austinmatherne-wk:
The code looks correct, but the PR description is confusing. It reads like negative values aren't allowed at all, and that the validator inspects the format and sign attributes to make sure they aren't used for that purpose. The code doesn't (and per spec shouldn't). It also claims the unit tests cover the format and sign attributes, but they only test raw value strings.

derekgengenbacher-wf (dismissed):

austinmatherne-wk (dismissed):

volodymyrtur-wk (dismissed):

arelle/ValidateXbrlDTS.py
```diff
@@ -1449,6 +1450,14 @@ def checkIxContinuationChain(val: ValidateXbrl, elt: ModelObject, chain: list[Mo
                     chain.append(elt)
                 checkIxContinuationChain(val, contAt, chain)  # type: ignore[arg-type]
 
+def _isNegativeDecimal(text: str) -> bool:
+    """Return True if text is a valid decimal number and is negative."""
+    try:
+        return Decimal(text.strip()) < 0
```
brettkail-wk: Out of curiosity, why is `.strip()` required?

It's a little unfortunate that we duplicate the `Decimal` parsing that `ModelInlineValueObject.value` has already done.  I wonder if checking `>= VALID` and then `value.startswith("-")` would be nicer.

brettkail-wk (dismissed):

arelle/ValidateXbrlDTS.py
aaroncameron-wk: The `.strip()` is redundant, I'll update.

`ModelInlineValueObject.value` incorporates the `sign` value, so in the case of `<ix:nonFraction sign="-">100</ix:nonFraction>` would result in `"-100"` and fire a false positive here. The rule is specifically concerned about cases where a negative sign is used in a raw text value rather than using the `sign` attribute (or a `format`). Also, I'm not sure that `xValid` is actually set at this stage of processing.

austinmatherne-wk (approved):

arelle/ValidateXbrlDTS.py
brettkail-wk: > `ModelInlineValueObject.value` incorporates the `sign` value

You're right, that was a poor suggestion, sorry.

> Also, I'm not sure that `xValid` is actually set at this stage of processing.

Sure, though it seems like the value processing could be moved earlier, or this message could be moved later.  It just seems unfortunate that we can't somehow avoid parsing the `Decimal` twice.  Anyway...

derekgengenbacher-wf (approved):

derekgengenbacher-wf:

PR #2417, opened 2026-06-11
Type Hints: arelle.ModelFormulaObject 

volodymyrtur-wk:
#### Description of change

#### Steps to Test
CI
**review**:
@Arelle/arelle


arelle/ModelFormulaObject.py
```diff
@@ -354,553 +390,607 @@ def evaluateRule(self, xpCtx, aspect):
...
-    def __repr__(self):
-        return ("modelFactVariable[{0}]{1})".format(self.objectId(),self.propertyView))
+    def __repr__(self) -> str:
+        return "modelFactVariable[{0}]{1})".format(self.objectId(), self.propertyView)
 
     @property
-    def viewExpression(self):
-        return ("fallbackValue =" + self.fallbackValue) if self.fallbackValue else ""
+    def viewExpression(self) -> str:
+        return "fallbackValue =" + self.fallbackValue or ""  # type: ignore[operator]
```
austinmatherne-wk: This isn't equivalent to the original ternary.

austinmatherne-wk (commented):
One issue with `viewExpression`, but otherwise looks good!

austinmatherne-wk (approved):

PR #2424, opened 2026-06-17
Type Hints: arelle.PrototypeInstanceObject 

volodymyrtur-wk:
#### Description of change

#### Steps to Test
CI
**review**:
@Arelle/arelle

### DIFF
--- a/arelle/PrototypeInstanceObject.py
+++ b/arelle/PrototypeInstanceObject.py
@@ -16,6 +16,7 @@ from arelle.ModelObject import ModelObject
 if TYPE_CHECKING:
     from arelle.ModelManager import ModelManager
     from arelle.ModelXbrl import ModelXbrl
+    from arelle.ModelInstanceObject import ModelContext
 
 Aspect: Any = None
 
@@ -75,9 +76,9 @@ class FactPrototype:  # behaves like a fact for dimensional validity testing
         return self.parent
 
     @property
-    def propertyView(self) -> tuple[tuple[str, str] | tuple[()], ...]:
+    def propertyView(self) -> tuple[tuple[str, str], tuple[str, str, tuple[tuple[str, str], ...]] | tuple[()]]:
         dims = self.context.qnameDims  # type: ignore[union-attr]
-        return (("concept", str(self.qname) if self.concept is not None else "not specified"),  # type: ignore[return-value]
+        return (("concept", str(self.qname) if self.concept is not None else "not specified"),
                 ("dimensions", "({0})".format(len(dims)),
                   tuple(dimVal.propertyView if dimVal is not None else (str(dim.qname), "None")
                         for dim, dimVal in sorted(dims.items(), key=lambda i:i[0])
@@ -203,20 +204,20 @@ class ContextPrototype:  # behaves like a context
     def nonDimValues(self, contextElement: str | int) -> list[ModelObject]:
         return self._nonDimValues.get(contextElement, [])
 
-    def isEntityIdentifierEqualTo(self, cntx2: ContextPrototype) -> bool:
+    def isEntityIdentifierEqualTo(self, cntx2: ModelContext) -> bool:
         return self.entityIdentifierHash is None or self.entityIdentifierHash == cntx2.entityIdentifierHash
 
-    def isPeriodEqualTo(self, cntx2: ContextPrototype) -> bool:
+    def isPeriodEqualTo(self, cntx2: ModelContext) -> bool:
         if self.isForeverPeriod:
-            return cntx2.isForeverPeriod
+            return cntx2.isForeverPeriod  # type: ignore[no-any-return]
         elif self.isStartEndPeriod:
             if not cntx2.isStartEndPeriod:
                 return False
-            return self.startDatetime == cntx2.startDatetime and self.endDatetime == cntx2.endDatetime
+            return self.startDatetime == cntx2.startDatetime and self.endDatetime == cntx2.endDatetime  # type: ignore[no-any-return]
         elif self.isInstantPeriod:
             if not cntx2.isInstantPeriod:
                 return False
-            return self.instantDatetime == cntx2.instantDatetime
+            return self.instantDatetime == cntx2.instantDatetime  # type: ignore[no-any-return]
         else:
             return False
 
--- a/arelle/PrototypeInstanceObject.py
+++ b/arelle/PrototypeInstanceObject.py
@@ -204,26 +204,26 @@ class ContextPrototype:  # behaves like a context
     def nonDimValues(self, contextElement: str | int) -> list[ModelObject]:
         return self._nonDimValues.get(contextElement, [])
 
-    def isEntityIdentifierEqualTo(self, cntx2: ModelContext) -> bool:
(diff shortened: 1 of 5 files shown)


arelle/PrototypeInstanceObject.py
```diff
@@ -196,16 +220,29 @@ def isPeriodEqualTo(self, cntx2):
         else:
             return False
 
-class DimValuePrototype():
-    typedMember: ModelConcept | None
-    def __init__(self, v, dimConcept, dimQname, mem, contextElement):
+
+class DimValuePrototype:
+    typedMember: ModelObject | QName | None
```
austinmatherne-wk: I don't think this can be a `QName`

arelle/PrototypeInstanceObject.py
```diff
@@ -170,19 +194,19 @@ def dimValue(self, dimQname):
...
             return self.segDimVals if contextElement == "segment" else self.scenDimVals
         else:
             return self.scenDimVals if contextElement == "segment" else self.segDimVals
 
-    def nonDimValues(self, contextElement):
+    def nonDimValues(self, contextElement: str | int) -> list[ModelObject]:
         return self._nonDimValues.get(contextElement, [])
 
-    def isEntityIdentifierEqualTo(self, cntx2):
+    def isEntityIdentifierEqualTo(self, cntx2: ContextPrototype) -> bool:
```
austinmatherne-wk: I'd expect the param here can also be `ModelContext`

arelle/PrototypeInstanceObject.py
```diff
@@ -170,19 +194,19 @@ def dimValue(self, dimQname):
...
         return self.entityIdentifierHash is None or self.entityIdentifierHash == cntx2.entityIdentifierHash
 
-    def isPeriodEqualTo(self, cntx2):
+    def isPeriodEqualTo(self, cntx2: ContextPrototype) -> bool:
```
austinmatherne-wk: Same here, can also be `ModelContext`

arelle/PrototypeInstanceObject.py
```diff
@@ -4,87 +4,111 @@
...
+    def getparent(self) -> Any:
         return self.parent
 
     @property
-    def propertyView(self):
-        dims = self.context.qnameDims
-        return (("concept", str(self.qname) if self.concept is not None else "not specified"),
+    def propertyView(self) -> tuple[tuple[str, str] | tuple[()], ...]:
+        dims = self.context.qnameDims  # type: ignore[union-attr]
+        return (("concept", str(self.qname) if self.concept is not None else "not specified"),  # type: ignore[return-value]
```
austinmatherne-wk: The return type doesn't match the return value?

volodymyrtur-wk: No, it is. Because of complicated tuple "comprehension" with conditions in it, mypy can't interpret it correctly

arelle/PrototypeInstanceObject.py
volodymyrtur-wk: Based on [the comment](https://github.com/Arelle/Arelle/blob/11aeb9d9f959874f3f3c4d8e26332265fd2f7b8f/arelle/PrototypeInstanceObject.py#L93), I assume `ContextPrototype` must behave like `ModelContext`

arelle/PrototypeInstanceObject.py
volodymyrtur-wk: Based on [the comment](https://github.com/Arelle/Arelle/blob/11aeb9d9f959874f3f3c4d8e26332265fd2f7b8f/arelle/PrototypeInstanceObject.py#L93), I assume `ContextPrototype` must behave like `ModelContext`

arelle/PrototypeInstanceObject.py
austinmatherne-wk: Yes, that's why the param should be typed for both. `ContextPrototype` and `ModelContext` behave the same and are both passed here.

arelle/PrototypeInstanceObject.py
austinmatherne-wk: Same as https://github.com/Arelle/Arelle/pull/2424#discussion_r3429937254

arelle/PrototypeInstanceObject.py
austinmatherne-wk: But the complicated `dimensions` comprehension produces nested tuples? mypy tells me it's returning this, which ~sadly~ looks correct:

`tuple[tuple[str, str], tuple[str, str, tuple[tuple[str, str], ...]] | tuple[()]]`

volodymyrtur-wk: Oh yes, this part produces nested tuples:
```python
tuple(dimVal.propertyView if dimVal is not None else (str(dim.qname), "None")
for dim, dimVal in sorted(dims.items(), key=lambda i:i[0])
if hasattr(dimVal, "propertyView")))
```

arelle/PrototypeInstanceObject.py
```diff
@@ -170,42 +195,55 @@ def dimValue(self, dimQname):
...
+            return self.instantDatetime == cntx2.instantDatetime  # type: ignore[no-any-return]
         else:
             return False
 
-class DimValuePrototype():
-    typedMember: ModelConcept | None
-    def __init__(self, v, dimConcept, dimQname, mem, contextElement):
+
+class DimValuePrototype:
+    typedMember: QName | None
```
austinmatherne-wk: The wrong type was removed:
```suggestion
    typedMember: ModelObject | None
```

arelle/PrototypeInstanceObject.py
```diff
@@ -170,42 +195,55 @@ def dimValue(self, dimQname):
...
             return self.segDimVals if contextElement == "segment" else self.scenDimVals
         else:
             return self.scenDimVals if contextElement == "segment" else self.segDimVals
 
-    def nonDimValues(self, contextElement):
+    def nonDimValues(self, contextElement: str | int) -> list[ModelObject]:
         return self._nonDimValues.get(contextElement, [])
 
-    def isEntityIdentifierEqualTo(self, cntx2):
+    def isEntityIdentifierEqualTo(self, cntx2: ModelContext) -> bool:
```
austinmatherne-wk: In my previous comment I meant this should be a union:
```suggestion
    def isEntityIdentifierEqualTo(self, cntx2: ContextPrototype | ModelContext) -> bool:
```

arelle/PrototypeInstanceObject.py
```diff
@@ -170,42 +195,55 @@ def dimValue(self, dimQname):
...
         return self.entityIdentifierHash is None or self.entityIdentifierHash == cntx2.entityIdentifierHash
 
-    def isPeriodEqualTo(self, cntx2):
+    def isPeriodEqualTo(self, cntx2: ModelContext) -> bool:
```
austinmatherne-wk: Same here:
```suggestion
    def isPeriodEqualTo(self, cntx2: ContextPrototype | ModelContext) -> bool:
```

austinmatherne-wk (approved):

aaroncameron-wk (approved):

derekgengenbacher-wf (approved):

derekgengenbacher-wf:

PR #2425, opened 2026-06-17
ESEF: Initial migration to `ValidationPlugin`

aaroncameron-wk:
#### Reason for change
The ESEF plugin predates the `ValidationPlugin` architecture and doesn't benefit from some of the design patterns introduced in it.

#### Description of change
This PR is intended to be a least-friction pathway towards using the `ValidationPlugin` architecture while having as little as possible effect on behavior.

#### Steps to Test
- [x] CI
- [x] GUI testing

**review**:
@Arelle/arelle

### DIFF
--- a/arelle/plugin/validate/ESEF/rules/base.py
+++ b/arelle/plugin/validate/ESEF/rules/base.py
@@ -1,0 +1,86 @@
+import re
+from typing import Any
+
+from collections.abc import Iterable
+
+from arelle.ValidateXbrl import ValidateXbrl
+from arelle.typing import TypeGetText
+from arelle.utils.PluginHooks import ValidationHook
+from arelle.utils.validate.Decorator import validation
+from arelle.utils.validate.Validation import Validation
+from ..ESEF_2021.ValidateXbrlFinally import validateXbrlFinally as validateXbrlFinally2021
+from ..ESEF_Current.ValidateXbrlFinally import validateXbrlFinally as validateXbrlFinallyCurrent
+from ..PluginValidationDataExtension import PluginValidationDataExtension
+from ..Util import getDisclosureSystemYear, shouldRunEsefValidationRules
+
+_: TypeGetText
+
+ixErrorPattern = re.compile(r"ix11[.]|xmlSchema[:]|(?!xbrl.5.2.5.2|xbrl.5.2.6.2)xbrl[.]|xbrld[ti]e[:]|utre[:]")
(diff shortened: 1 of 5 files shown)


arelle/plugin/validate/ESEF/rules/base.py
```diff
@@ -0,0 +1,86 @@
+import re
```
austinmatherne-wk, line 4: Copyright header is missing.

arelle/plugin/validate/ESEF/rules/base.py
```diff
@@ -0,0 +1,86 @@
...
+    hook=ValidationHook.FINALLY,
+)
+def rule_finally(
+        pluginData: PluginValidationDataExtension,
+        val: ValidateXbrl,
+        *args: Any,
+        **kwargs: Any,
+) -> Iterable[Validation]:
+    """
+    """
```
austinmatherne-wk, line 38: Empty docstrings here and below

arelle/plugin/validate/ESEF/rules/base.py
```diff
@@ -0,0 +1,86 @@
...
+    if not shouldRunEsefValidationRules(val):
```
austinmatherne-wk, line 39: This isn't needed within the validation function, is it? The validation plugin framework/`_executeModelValidations` already checks if a disclosure system from the plugin is selected when deciding to invoke the validation function?

volodymyrtur-wk (dismissed):

arelle/plugin/validate/ESEF/rules/base.py
aaroncameron-wk: Just to be cautious, I wanted to retain the [`validateDisclosureSystem`](https://github.com/aaroncameron-wk/arelle-public/blob/c893c3467f7102bb29f0387836aa6cb8d5aab0ae/arelle/plugin/validate/ESEF/Util.py#L211-L212) check and [`ESEFplugin`](https://github.com/aaroncameron-wk/arelle-public/blob/c893c3467f7102bb29f0387836aa6cb8d5aab0ae/arelle/plugin/validate/ESEF/Util.py#L207) property check which would not otherwise be considered by the `ValidationPlugin` implementation. Those checks probably could be omitted/consolidated but I wanted to defer that to future efforts.

austinmatherne-wk (approved):

allenbross-wf:
QA +10
All local tests and integration tests passed.

PR #2427, opened 2026-06-18
Addressing issue where unit id is being inspected without verifying existence of unit.

allenbross-wf:
#### Description of change

Addressing issue where unit id is being inspected without verifying existence of unit.

#### Steps to Test

CI

**review**:
@Arelle/arelle

### DIFF
--- a/arelle/plugin/validate/DBA/rules/fr.py
+++ b/arelle/plugin/validate/DBA/rules/fr.py
@@ -1107,6 +1107,8 @@ def rule_fr74(
         liabilityOtherFact = None
         provisionFact = None
         for fact in facts:
+            if fact.unit is None:
+                continue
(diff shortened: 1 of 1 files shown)


arelle/plugin/validate/DBA/rules/fr.py
```diff
@@ -1107,6 +1107,8 @@ def rule_fr74(
         liabilityOtherFact = None
         provisionFact = None
         for fact in facts:
+            if fact is None or fact.unit is None:
```
austinmatherne-wk: This is fine, but the `fact is None` check isn't needed. `fact` can never be `None` in these paths.

austinmatherne-wk (dismissed):

aaroncameron-wk (dismissed):

austinmatherne-wk (approved):

PR #2432, opened 2026-06-23
Type Hints: arelle.ModelInstanceObject 

volodymyrtur-wk:
#### Description of change

#### Steps to Test
CI
**review**:
@Arelle/arelle

### DIFF
--- a/arelle/ModelInstanceObject.py
+++ b/arelle/ModelInstanceObject.py
@@ -43,6 +43,7 @@ from arelle.ValidateXbrlCalcs import inferredPrecision, inferredDecimals, roundV
 from arelle.XmlValidateConst import UNVALIDATED, INVALID, VALID
 from arelle.XmlValidate import validate as xmlValidate
 from arelle.XmlUtil import collapseWhitespace
+from arelle.ModelValue import QName
 from arelle.PrototypeInstanceObject import DimValuePrototype
 from arelle.ModelDtsObject import ModelResource
 from arelle.typing import ModelFactBase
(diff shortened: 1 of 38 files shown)


arelle/ModelInstanceObject.py
```diff
@@ -1122,45 +1199,45 @@ def dimValues(self, contextElement):
...
 
-    def hasDimension(self, dimQname):
+    def hasDimension(self, dimQname: QName) -> bool:
         """(bool) -- True if dimension concept qname is reported by context (in either context element), not including defaulted dimensions."""
         return dimQname in self.qnameDims
 
     # returns ModelDimensionValue for instance dimensions, else QName for defaults
-    def dimValue(self, dimQname):
+    def dimValue(self, dimQname: QName) -> ModelDimensionValue | QName | None:
         """(ModelDimension or QName) -- ModelDimension object if dimension is reported (in either context element), or QName of dimension default if there is a default, otherwise None"""
```
aaroncameron-wk, line 1208: Pre-existing, but a few comments refer to the non-existent `ModelDimension`. Could be corrected to `ModelDimensionValue`. Not worth another push if nothing else needs changing, though.

aaroncameron-wk (approved):

arelle/ModelInstanceObject.py
```diff
@@ -1307,53 +1385,54 @@ def isEqualTo_(self, cntx2, dimensionalAspectModel):
...
+                if not XbrlUtil.sEqual(self.modelXbrl, self.scenario, cntx2.scenario):  # type: ignore[arg-type]
                     return False
             elif cntx2.hasScenario:
                 return False
 
         return True
 
     @property
-    def propertyView(self):
+    def propertyView(self) -> tuple[tuple[str, str, tuple[tuple[str, str]]] | tuple[str, str], ...]:  # type: ignore[override]
```
austinmatherne-wk, line 1413: I don't think it should be addressed in this PR, but it does look like some of the types on our base classes are declared too narrow (extending classes use similar but different types).

austinmatherne-wk (approved):
There's a lot here. Not an easy module to add type hints too, good work!

derekgengenbacher-wf (approved):

derekgengenbacher-wf:

PR #2433, opened 2026-06-23
GUI: Multi-select packages and save on close

aaroncameron-wk:
#### Reason for change
Selecting packages through the "Locally" menu requires selecting one package at a time. 

Additionally, package selections/settings aren't saved to disk when closing the package manager, which can result in loss of package configuration changes depending on how the application ultimately exits. 

#### Description of change
- Configure local package selection dialog to be multi-select.
- Save package config changes when closing (not cancelling) the package manager dialog.

#### Steps to Test
- [x] Tested on Mac OS and Windows.

**review**:
@Arelle/arelle

### DIFF
--- a/arelle/DialogPackageManager.py
+++ b/arelle/DialogPackageManager.py
@@ -384,8 +384,9 @@ class DialogPackageManager(Toplevel):
         initialdir = self.cntlr.pluginDir # default plugin directory
         #if not self.cntlr.isMac: # can't navigate within app easily, always start in default directory
         initialdir = self.cntlr.config.setdefault("packageOpenDir", initialdir)
-        filename = self.cntlr.uiFileDialog("open",
+        filenames = self.cntlr.uiFileDialog("open",
                                            parent=self,
+                                           multiple=True,
                                            title=_("Choose taxonomy package file"),
                                            initialdir=initialdir,
                                            filetypes=[(_("Taxonomy package files (*.zip)"), "*.zip"),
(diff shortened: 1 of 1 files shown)


arelle/DialogPackageManager.py
```diff
@@ -384,16 +385,17 @@ def findLocally(self):
...
+                                           multiple=True,
                                            title=_("Choose taxonomy package file"),
                                            initialdir=initialdir,
                                            filetypes=[(_("Taxonomy package files (*.zip)"), "*.zip"),
                                                       (_("PWD Manifest (taxonomyPackage.xml)"), "taxonomyPackage.xml"),
                                                       (_("pre-PWD Manifest (*.taxonomyPackage.xml)"), "*.taxonomyPackage.xml"),
                                                       (_("pre-PWD Oasis Catalog (*catalog.xml)"), "*catalog.xml")],
                                            defaultextension=".zip")
-        if filename:
+        for filename in filenames:
```
austinmatherne-wk, line 398: What happens if the user selects multiple files and some are valid packages and others aren't? Should we have exception handling here so subsequent files are processed even if an earlier one fails? Does the log inform the user which file(s) are invalid?

volodymyrtur-wk (approved):

arelle/DialogPackageManager.py
aaroncameron-wk: The error handling in [`packageInfo`](https://github.com/Arelle/Arelle/blob/f1a69b4cd614b57bc3b368c80101a359f5ecf0aa/arelle/packages/_package_manager.py#L532) seems sufficient. An invalid package does not prevent the loading of subsequent packages. 

Each invalid package will [yield a warning dialog](https://github.com/Arelle/Arelle/blob/f1a69b4cd614b57bc3b368c80101a359f5ecf0aa/arelle/DialogPackageManager.py#L426-L429). 
Screenshot 2026-06-24 at 10 59 04 AM

The messages pane in the main window also provides a more detailed error message.
Screenshot 2026-06-24 at 10 58 37 AM

austinmatherne-wk (approved):

derekgengenbacher-wf (approved):

derekgengenbacher-wf:

PR #2434, opened 2026-06-24
Table Constraints Enum Values Validation

austinmatherne-wk:
#### Reason for change

Implements the remaining table constraint enumeration value validation rules. The same value validation code will be reused to validate xBRL-CSV report values.

The TC conformance suite uncovered a limitation in Arelle's date handling. The `DateTime` class extends `datetime.datetime`, which restricts years to the range 1–9999 (`datetime.MINYEAR` to `datetime.MAXYEAR`). This is narrower than the XML `xs:date` and `xs:dateTime` value space. Correctly supporting the full range has broader ramifications beyond table constraints and will likely require an Arelle 3.0 release. That work will be handled separately.

#### Description of change
* Implement the remaining of the 1000+ TC metadata value test cases.

#### Steps to Test
* CI

**review**:
@Arelle/arelle


arelle/oim/_tc/value_validator.py
```diff
@@ -0,0 +1,353 @@
...
+
+    def _is_valid_unit(self, value: str) -> bool:
+        unit_qnames = PREFIXED_QNAME_PATTERN.findall(value)
+        if not unit_qnames:
+            return False
+        substituted = PREFIXED_QNAME_PATTERN.sub(UNIT_QNAME_SUBSTITUTION_CHAR, value)
+        if UNIT_PATTERN.fullmatch(substituted) is None:
+            return False
+        for unit_qname in unit_qnames:
+            qname_validation_result = self._validate_base_type(QNAME, unit_qname)
```
aaroncameron-wk, line 224: The intent here is to validate the _part_ (numerator/denominator) as a QName, correct? `_validate_base_type` will internally also apply the unit-level facet validations, which is incorrect by my understanding, but maybe not actually a concern in practice.

aaroncameron-wk (dismissed):

arelle/oim/_tc/value_validator.py
austinmatherne-wk: Good catch!

derekgengenbacher-wf (approved):

aaroncameron-wk (approved):

PR #2436, opened 2026-06-24
Type Hints: arelle.ViewFileFactList 

volodymyrtur-wk:
#### Description of change

#### Steps to Test
CI
**review**:
@Arelle/arelle

### DIFF
--- a/arelle/ViewFileFactList.py
+++ b/arelle/ViewFileFactList.py
@@ -3,7 +3,7 @@ See COPYRIGHT.md for copyright information.
 '''
 from __future__ import annotations
 
-from typing import TYPE_CHECKING, Any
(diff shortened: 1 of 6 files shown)


arelle/ViewFileFactList.py
```diff
@@ -1,16 +1,37 @@
 '''
 See COPYRIGHT.md for copyright information.
 '''
+from __future__ import annotations
+
+from typing import TYPE_CHECKING, Any
```
austinmatherne-wk: nit: `Any` import is unused

austinmatherne-wk (dismissed):

aaroncameron-wk (dismissed):

austinmatherne-wk (approved):

PR #2442, opened 2026-06-25
Build tcl 9 libs

austinmatherne-wk:
#### Reason for change

We built and packaged tktable libs for tcl/tk 8, but not 9. Users running Arelle with Python built against tcl/tk 9 get an exception that tktable lib can't be found when opening taxonomies including a table linkbase in the GUI. The existing tcl files are already configured to locate the tcl/tk 9 libs.

#### Description of change

* Update build-tktable.yml workflow to build libs for both tcl/tk 8 and 9.
* Check in tktable lib [files created](https://github.com/Arelle/Arelle/actions/runs/28194664239) from updated workflow.

#### Steps to Test
* [x] open reports with table linkbase 

**review**:
@Arelle/arelle


aaroncameron-wk (approved):

derekgengenbacher-wf (approved):

derekgengenbacher-wf:

PR #2443, opened 2026-06-26
Type Hints: arelle.ViewWinTree 

volodymyrtur-wk:
#### Description of change

#### Steps to Test
CI
**review**:
@Arelle/arelle

### DIFF
--- a/arelle/ViewWinTree.py
+++ b/arelle/ViewWinTree.py
@@ -3,10 +3,9 @@ See COPYRIGHT.md for copyright information.
 '''
 from __future__ import annotations
 import os
-from typing import TYPE_CHECKING
-
-from tkinter import *
-from tkinter.ttk import *
+from tkinter import HORIZONTAL, VERTICAL, N, S, E, W, StringVar, TclError, Menu, PhotoImage
+from tkinter.ttk import Frame, Scrollbar, Treeview, Notebook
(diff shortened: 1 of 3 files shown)


arelle/ViewWinTree.py
```diff
@@ -271,7 +287,13 @@ def menuAddUnitDisplay(self):
                                     modelObject=self.modelXbrl.modelDocument, title=self.tabTitle, error=str(ex))
                 self.menu = None
 
-    def menuAddViews(self, addClose=True, tabWin=None, additionalViews=None, additionalViewMethod=None):
+    def menuAddViews(
+        self,
+        addClose: bool = True,
+        tabWin: Notebook | None = None,
+        additionalViews: Iterable[tuple[Any, ...]] | None = None,
+        additionalViewMethod: Callable[[ModelXbrl, tuple[Any, ...], Iterable[tuple[Any, ...]] | None], None] | None = None,
```
austinmatherne-wk: This type doesn't look quite right. The `tuple[Any, ...]` part is unpacked via `*b` on line 316 below, so the declared parameter count doesn't match the actual call convention. `Callable[..., None] | None` would be more accurate without implying a specific (wrong) signature.

austinmatherne-wk (approved):

derekgengenbacher-wf (approved):

derekgengenbacher-wf:

Issue #2445, opened 2026-06-26
Base XSD value validation fixes


aaroncameron-wk:
### What should we change and why?

> [!IMPORTANT]
> I was able to convert a subset of the W3C XSD tests published and maintained [here](https://github.com/w3c/xsdtests) into an XBRL conformance suite that can be consumed by Arelle. The results yielded the following validation issues. I'll PR a solution for each individually so each can be evaluated independently. Finally, I'll configure the transformed conformance suite to run in CI.

# Base XSD value validation fixes

Corrections to Arelle's simple-type value validation. Most live in
`arelle/XmlValidate.py::_validateValueStringOrRaise`; one (Fix 6) is in the
`xs:duration` value object in `arelle/ModelValue.py`. Each entry states what was
wrong, the change, and the **normative basis** so the fix can be verified against
the W3C specifications alone.

Specifications referenced:

- **XML 1.0** — *Extensible Markup Language (XML) 1.0 (Fifth Edition)*, W3C
  Recommendation. <https://www.w3.org/TR/xml/>
- **XSD 2** — *XML Schema Part 2: Datatypes (Second Edition)*, W3C Recommendation.
  <https://www.w3.org/TR/xmlschema-2/>
- **XSD 1.1 D** — *W3C XML Schema Definition Language (XSD) 1.1 Part 2: Datatypes*,
  W3C Recommendation. <https://www.w3.org/TR/xmlschema11-2/> (same semantics; cited
  where convenient.)

---

- [x] Fix 1 — `Name` permits a leading colon
- [x] Fix 2 — XSD regex escape `\i` includes the colon
- [x] Fix 3 — ordering facets enforced on date/time (and other ordered) types
- [x] Fix 4 — `length` facets count octets for `hexBinary` / `base64Binary`
- [x] Fix 5 — `length` facets are vacuous for `QName` / `NOTATION`
- [x] Fix 6 — `xs:duration` greater-than comparison (`IsoDuration.__gt__`)
- [x] Fix 7 — `totalDigits` must not count the sign
- [x] Fix 8 — `enumeration` is matched in the value space, not lexically
- [x] Fix 9 — the `g*` date/time types (`gYearMonth`, `gYear`, `gMonthDay`, `gMonth`,
`gDay`) silently discard their timezone

PR #2446, opened 2026-06-26
Base XSD type validation: `Name` permits a leading colon

aaroncameron-wk:
Part of #2445 

## Fix 1 — `Name` permits a leading colon

**Where:** `namePattern` (module-level compiled regex used as the lexical pattern
for `baseXsdType == "Name"`).

**Symptom:** the value space of `xsd:Name` was under-accepted — a `Name` beginning
with `:` (e.g. `:foo`) was rejected, although it is lexically valid.

**Change:** add `:` to the *first* character class of `namePattern` (the
NameStartChar position). The trailing class (NameChar) already permitted `:`.

```text
before:  ^[_A-Za-z…][ _-.: … ]*$
after:   ^[:_A-Za-z…][ _-.: … ]*$
              ^ colon added to the start-character class
```

**Normative basis:**

- XML 1.0 §2.3 *Common Syntactic Constructs*, production [4]:

  ```
  NameStartChar ::= ":" | [A-Z] | "_" | [a-z] | [#xC0-#xD6] | …
  NameChar      ::= NameStartChar | "-" | "." | [0-9] | #xB7 | …
  ```

  `":"` is the **first** alternative of `NameStartChar`, so a `Name` may begin with
  a colon. <https://www.w3.org/TR/xml/#NT-NameStartChar>

- XSD 2 §3.3.7 *Name*: "Name represents XML Names. The value space of Name is the
  set of all strings which match the Name production of [XML 1.0]."
  <https://www.w3.org/TR/xmlschema-2/#Name>

- Contrast XSD 2 §3.3.8 *NCName*: "the set of all strings which match the NCName
  production" — i.e. `Name` *minus the colon*. The existence of a separate, colon-
  excluding `NCName` is precisely because `Name` itself **includes** the colon.
  <https://www.w3.org/TR/xmlschema-2/#NCName>

**Independent check:** the first character of `:foo` is `:`, which appears verbatim
in production [4] `NameStartChar`; the remaining characters `f`, `o`, `o` are
`[a-z]` ⊂ `NameChar`. Therefore `:foo` matches `Name ::= NameStartChar NameChar*`
and must be accepted. (NCName, by contrast, must still reject `:foo` — see Fix 2.)

#### Steps to Test
CI

**review**:
@Arelle/arelle

### DIFF
--- a/tests/unit_tests/arelle/test_xmlvalidate.py
+++ b/tests/unit_tests/arelle/test_xmlvalidate.py
@@ -323,6 +323,7 @@ BASE_XSD_TYPES = {
     ],
     "NCName": [
         {"value": "*invalid", "expected": ("=", None, INVALID)},
+        {"value": ":invalid", "expected": ("=", None, INVALID)},
         {"value": "valid", "expected": ("=", "=", VALID)},
     ],
     "negativeInteger": [
--- a/tests/unit_tests/arelle/test_xmlvalidate.py
+++ b/tests/unit_tests/arelle/test_xmlvalidate.py
@@ -319,6 +319,7 @@ BASE_XSD_TYPES = {
     "Name": [
         {"value": "*invalid", "expected": ("=", None, INVALID)},
         {"value": "valid", "expected": ("=", "=", VALID)},
+        {"value": ":valid", "expected": ("=", "=", VALID)},
     ],
     "NCName": [
         {"value": "*invalid", "expected": ("=", None, INVALID)},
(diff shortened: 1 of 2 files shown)


tests/unit_tests/arelle/test_xmlvalidate.py
```diff
@@ -319,6 +319,7 @@
     "Name": [
         {"value": "*invalid", "expected": ("=", None, INVALID)},
         {"value": "valid", "expected": ("=", "=", VALID)},
+        {"value": ":valid", "expected": ("=", "=", VALID)},
```
austinmatherne-wk, line 322: Adding an invalid case for this value to `NCName` would be nice

austinmatherne-wk (dismissed):

derekgengenbacher-wf (dismissed):

austinmatherne-wk (approved):

PR #2453, opened 2026-06-29
Enforce ordering facets on date/time and other ordered types

aaroncameron-wk:
Part of #2445

## Fix 3 — ordering facets enforced on date/time (and other ordered) types

**Where:** the `else` arm of `_validateValueStringOrRaise` that parses the
lexical-pattern datatypes (`date`, `dateTime`, `time`, `gYearMonth`, `gYear`,
`gMonthDay`, `gMonth`, `gDay`, `duration`, and the XBRL date union).

**Symptom:** the four bounding facets `maxInclusive`, `maxExclusive`,
`minInclusive`, `minExclusive` were only enforced in the numeric branches
(`decimal`/`float`/`double` and the integer family). For ordered non-numeric types
the value was parsed into a comparable object but the bounds were never checked, so
an out-of-range value (e.g. a `date` greater than a `maxExclusive` bound) was
accepted.

**Change:** after the value is parsed (so `xValue` is a comparable
date/time/duration object), apply the same four bounds already enforced by the
numeric branches, routed through a helper `_orderedComparison(value, bound)` that
returns `-1`/`0`/`1` for `<`/`=`/`>`, or `None` when the order between `value` and
`bound` is indeterminate:

```python
if facets:
    if "maxInclusive" in facets and _orderedComparison(xValue, facets["maxInclusive"]) not in (-1, 0):
        raise ValueError(…)
    if "maxExclusive" in facets and _orderedComparison(xValue, facets["maxExclusive"]) != -1:
        raise ValueError(…)
    if "minInclusive" in facets and _orderedComparison(xValue, facets["minInclusive"]) not in (0, 1):
        raise ValueError(…)
    if "minExclusive" in facets and _orderedComparison(xValue, facets["minExclusive"]) != 1:
        raise ValueError(…)
```

Each check requires a **determinate proof** that the value satisfies the facet;
anything else — a determinate violation, or an indeterminate order — is rejected
(see Fix 3a for why indeterminate orders arise and why they must be rejected, not
accepted). The facet bound is itself parsed through the same datatype path, so each
comparison is between like-typed, ordered values. Non-ordered lexical types
(`language`, `base64Binary`, `hexBinary`) never carry these facets, so gating on
facet presence needs no extra type guard.

**Normative basis:**

- XSD 2 §4.3 *Constraining Facets* defines the bounds and their facet-validity
  rules. For a value `v`:

  - §4.3.7 `maxInclusive`: valid iff `v ≤ maxInclusive`.
    <https://www.w3.org/TR/xmlschema-2/#rf-maxInclusive>
  - §4.3.8 `maxExclusive`: valid iff `v < maxExclusive`.
    <https://www.w3.org/TR/xmlschema-2/#rf-maxExclusive>
  - §4.3.9 `minExclusive`: valid iff `v > minExclusive`.
    <https://www.w3.org/TR/xmlschema-2/#rf-minExclusive>
  - §4.3.10 `minInclusive`: valid iff `v ≥ minInclusive`.
    <https://www.w3.org/TR/xmlschema-2/#rf-minInclusive>

  Each check above requires the corresponding relation to hold determinately
  (`_orderedComparison` result restricted to the value(s) that prove it); anything
  else raises.

- These facets apply to any datatype whose fundamental facet **ordered** is not
  `false`. The date/time and duration primitives are `ordered = partial`, and each
  lists `maxInclusive`/`maxExclusive`/`minInclusive`/`minExclusive` among its
  applicable Constraining facets — see, e.g.:

  - §3.2.7 `dateTime` <https://www.w3.org/TR/xmlschema-2/#dateTime>
  - §3.2.9 `date` <https://www.w3.org/TR/xmlschema-2/#date>
  - §3.2.8 `time`, §3.2.10 `gYearMonth`, §3.2.11 `gYear`, §3.2.12 `gMonthDay`,
    §3.2.13 `gDay`, §3.2.14 `gMonth`, §3.2.6 `duration`
  - Summary of applicable facets per datatype: Appendix B / §4.1.5.
    <https://www.w3.org/TR/xmlschema-2/#facets>

- The comparison itself is well-defined: XSD 2 §3.2.7 gives the **order relation**
  for `dateTime` (anchor `#dateTime-order`; the other date/time types order
  analogously), and §3.2.6 gives it for `duration` (anchor `#duration-order`).
  <https://www.w3.org/TR/xmlschema-2/#dateTime-order>

**Independent check:** take a `date`-derived type with `maxExclusive = 2011-10-16`
and the candidate value `2027-07-04`. By the §3.2.9 order relation,
`2027-07-04 > 2011-10-16`, hence `2027-07-04 < maxExclusive` is **false**, so by
§4.3.8 the value is **not** facet-valid and must be rejected. Symmetrically, a value
strictly before the bound (e.g. `2009-01-01`) is `< maxExclusive` and remains valid.

### Fix 3a — timezone-aware vs. timezone-naive ordering: no crash, and indeterminate orders are rejected

**Where:** the same ordering-facet block, plus the helpers `_orderedComparison` /
`_comparableInstant` in `arelle/XmlValidate.py`.

**Symptom:** two distinct problems in the naive four-comparison approach:

1. `DateTime` and `Time` subclass Python's `datetime.datetime` / `datetime.time`,
   and Python **raises `TypeError`** ("can't compare offset-naive and
   offset-aware datetimes/times") when ordering a value that carries a timezone
   against one that does not. So a `dateTime`/`date`/`time` value like
   `2025-01-02T03:04:05Z` compared against an un-timezoned facet bound (or vice
   versa) raised `TypeError`. `validateValue` only catches
   `ValueError`/`InvalidOperation`, so the `TypeError` escaped and crashed
   validation rather than producing a value error. (The gregorian types
   `gYear`/`gYearMonth`/`gMonth`/`gMonthDay`/`gDay` are unaffected — they discard
   the timezone at parse time and compare plain integers.)
2. Even once ordering is made not to crash, the order relation on `dateTime` (and
   analogously `date`/`time`) is only a **partial** order: when exactly one operand
   is timezoned, its untimezoned counterpart's instant is only known to within
   ±14 h, and the true order can be genuinely indeterminate. A naive
   `<`/`>`/`>=`/`<=` rewrite of the facet tests has no way to represent "neither
   provably true nor provably false", and must decide one way or the other for
   values that fall in that uncertain band.

**Change:** `_orderedComparison(value, bound)` returns `-1`/`0`/`1` for
`<`/`=`/`>`, or `None` when the order is indeterminate. It first uses the operands'
own comparison operators; on the `TypeError` from a naive-vs-aware `datetime`/`time`
it instead compares the underlying instants under the XSD timezone-straddling rule,
returning `-1`/`1` only when the ±14 h uncertainty band lies wholly to one side of
the other operand, and `None` when it straddles it. Every caller of
`_orderedComparison` (Fix 3) then requires a determinate proof of compliance, so
`None` is treated as **failing** the facet, on par with a determinate violation:
neither crashes, but neither is accepted.

**Normative basis:**

- XSD 2 §3.2.7.4 *Order relation on dateTime* (with §3.2.7.3): a `dateTime` (and,
  analogously, `date`/`time`) **without** a timezone is compared by considering the
  recoverable timezone offsets, i.e. as if it could lie anywhere in `[−14:00,
  +14:00]`. Concretely, normalize the timezoned operand to UTC; the un-timezoned
  operand's instant is then only known to within ±14 h. The order is **determinate**
  when that ±14 h band lies wholly to one side of the other operand, and
  **indeterminate** when it straddles it. <https://www.w3.org/TR/xmlschema-2/#dateTime-order>

- §3.2.6.3 *Facet Comparison for durations*: "In comparing duration values with
  minInclusive, minExclusive, maxInclusive and maxExclusive facet values
  indeterminate comparisons should be considered as 'false'."
  <https://www.w3.org/TR/xmlschema-2/#facet-comparison-for-durations> The same
  bounding facets, defined identically in §4.3.7–§4.3.10 for every ordered
  datatype, govern `dateTime`/`date`/`time`/the `g*` types; there is no basis for a
  different resolution of indeterminacy for these types than for `duration` — both
  share the fundamental facet `ordered = partial` and the same four Constraining
  facets.

- §3.2.4 `float` / §3.2.5 `double` state the equivalent rule for their own
  **incomparable** value (`NaN`): "Any value incomparable with the value used for
  the four bounding facets … will be excluded from the resulting restricted value
  space." <https://www.w3.org/TR/xmlschema-2/#float> Indeterminate/incomparable is
  consistently treated as exclusion (rejection), never inclusion, everywhere the
  specification is explicit about it.

**Independent check:** with `maxInclusive = 2025-01-02T03:04:05` (no timezone) and
the candidate `2025-01-04T00:00:00Z`, the bound's instant is at most
`2025-01-02T17:04:05Z` (offset −14:00), and the candidate exceeds that, so
`value > maxInclusive` is **determinate** and the value is rejected. The candidate
`2025-01-02T03:04:05Z` against the same bound falls inside the ±14 h band, so the
order is indeterminate; per §3.2.6.3 that is "false", so the value is rejected too
(without raising). By contrast, the candidate `2025-01-01T00:00:00` (no timezone)
against `maxInclusive = 2025-01-02T03:04:05Z` is provably more than 14 h before the
earliest instant the bound could represent, so the order is determinately `<` and
the value is **accepted**.

#### Steps to Test
- CI

**review**:
@Arelle/arelle

### DIFF
--- a/arelle/XmlValidate.py
+++ b/arelle/XmlValidate.py
@@ -475,7 +475,9 @@ def _orderedComparison(value: Any, bound: Any) -> int | None:
     """Order ``value`` against an ordering-facet ``bound``.
 
     Returns ``-1``/``0``/``1`` when ``value`` is less than/equal to/greater than
-    ``bound``, or ``None`` when the order is indeterminate.
+    ``bound``, or ``None`` when the order is indeterminate. Callers should treat
+    ``None`` as failing whichever relation the facet requires (Datatypes 3.2.6.3:
+    "indeterminate comparisons should be considered as 'false'").
 
     Ordinarily the operands' own comparison operators are used. For an xs:date/time
     value where exactly one operand carries a timezone, Python raises ``TypeError``
@@ -712,15 +714,17 @@ def _validateValueStringOrRaise(
             sValue = value
             if facets: # ordering facets on date/time/gYear/... (xValue is a comparable type)
                 # _orderedComparison tolerates xs:date/time values that mix timezoned and
-                # untimezoned operands (which Python won't order); an indeterminate order is
-                # treated as satisfying the facet rather than crashing or wrongly rejecting.
-                if "maxInclusive" in facets and _orderedComparison(xValue, facets["maxInclusive"]) == 1:
+                # untimezoned operands (which Python won't order); per XSD Datatypes 3.2.6.3
+                # ("indeterminate comparisons should be considered as 'false'") an
+                # indeterminate order fails the facet test, so the value is rejected rather
+                # than accepted or crashing.
+                if "maxInclusive" in facets and _orderedComparison(xValue, facets["maxInclusive"]) not in (-1, 0):
                     raise ValueError(" > maxInclusive {0}".format(facets["maxInclusive"]))
-                if "maxExclusive" in facets and _orderedComparison(xValue, facets["maxExclusive"]) in (0, 1):
+                if "maxExclusive" in facets and _orderedComparison(xValue, facets["maxExclusive"]) != -1:
                     raise ValueError(" >= maxExclusive {0}".format(facets["maxExclusive"]))
-                if "minInclusive" in facets and _orderedComparison(xValue, facets["minInclusive"]) == -1:
+                if "minInclusive" in facets and _orderedComparison(xValue, facets["minInclusive"]) not in (0, 1):
                     raise ValueError(" < minInclusive {0}".format(facets["minInclusive"]))
-                if "minExclusive" in facets and _orderedComparison(xValue, facets["minExclusive"]) in (0, -1):
+                if "minExclusive" in facets and _orderedComparison(xValue, facets["minExclusive"]) != 1:
                     raise ValueError(" <= minExclusive {0}".format(facets["minExclusive"]))
     return XmlValidationResult(sValue=sValue, xValue=xValue, xValid=xValid)
 
--- a/arelle/XmlValidate.py
+++ b/arelle/XmlValidate.py
@@ -653,6 +653,15 @@ def _validateValueStringOrRaise(
             else: # no lexical pattern, forget compiling value
                 xValue = value
             sValue = value
+            if facets: # ordering facets on date/time/gYear/... (xValue is a comparable type)
+                if "maxInclusive" in facets and xValue > facets["maxInclusive"]:
+                    raise ValueError(" > maxInclusive {0}".format(facets["maxInclusive"]))
+                if "maxExclusive" in facets and xValue >= facets["maxExclusive"]:
+                    raise ValueError(" >= maxExclusive {0}".format(facets["maxExclusive"]))
+                if "minInclusive" in facets and xValue < facets["minInclusive"]:
+                    raise ValueError(" < minInclusive {0}".format(facets["minInclusive"]))
+                if "minExclusive" in facets and xValue <= facets["minExclusive"]:
+                    raise ValueError(" <= minExclusive {0}".format(facets["minExclusive"]))
     return XmlValidationResult(sValue=sValue, xValue=xValue, xValid=xValid)
 
 
--- a/arelle/XmlValidate.py
+++ b/arelle/XmlValidate.py
@@ -3,6 +3,7 @@ See COPYRIGHT.md for copyright information.
 '''
 from __future__ import annotations
 
+import datetime
 from dataclasses import dataclass
 import logging
 from collections.abc import Callable, Mapping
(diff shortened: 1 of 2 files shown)


volodymyrtur-wk (dismissed):

tests/unit_tests/arelle/test_xmlvalidate.py
```diff
@@ -702,6 +702,43 @@ def test_validateValue_facets_minMaxExclusive(value: str, expected: tuple):
...
+        # dateTime
+        ("dateTime", "2025-01-02T03:04:05", {"maxInclusive": DateTime(2025, 1, 2, 3, 4, 5)}, VALID),
+        ("dateTime", "2025-01-02T03:04:06", {"maxInclusive": DateTime(2025, 1, 2, 3, 4, 5)}, INVALID),
+        # gYear
+        ("gYear", "0001", {"maxInclusive": gYear(5)}, VALID),
+        ("gYear", "0009", {"maxInclusive": gYear(5)}, INVALID),
+        # duration whose bounds differ in the years/months/days part (independent of the
+        # seconds tie-break, which is Fix 6)
+        ("duration", "P6M", {"maxExclusive": isoDuration("P1Y")}, VALID),
+        ("duration", "P2Y", {"maxExclusive": isoDuration("P1Y")}, INVALID),
```
austinmatherne-wk, line 757: It would be nice to include tests for all covered types (time, gMonth, gMonthDay, gDay). They use different classes, so I think the coverage is justified. Also, I'm not sure what Arelle does currently, but it would be good to make sure we don't crash on timezone aware vs naive DateTime comparisons.

aaroncameron-wk: Good catch on timezone vs. naive comparisons, that's an interesting problem to handle but the [spec](https://www.w3.org/TR/xmlschema-2/#dateTime-order) does address it directly.

austinmatherne-wk: I don't think that's a problem we want to address here. XPath and XBRL both further complicate this (this is a large part of why I bailed on implementing my own classes in the table constraints work). For this PR I think it would be good to just know this code doesn't raise an unhandled exception.

aaroncameron-wk: Thoughts on the latest commit? I suppose to be safe, the re-raise in `_orderedComparison` could be replaced with a `return None`

volodymyrtur-wk (dismissed):

arelle/XmlValidate.py
```diff
@@ -653,6 +710,18 @@ def _validateValueStringOrRaise(
             else: # no lexical pattern, forget compiling value
                 xValue = value
             sValue = value
+            if facets: # ordering facets on date/time/gYear/... (xValue is a comparable type)
+                # _orderedComparison tolerates xs:date/time values that mix timezoned and
+                # untimezoned operands (which Python won't order); an indeterminate order is
+                # treated as satisfying the facet rather than crashing or wrongly rejecting.
```
austinmatherne-wk: The spec isn't explicit for dateTime, but [for duration](https://www.w3.org/TR/xmlschema-2/#facet-comparison-for-durations) it says `indeterminate comparisons should be considered as "false".`. For float and double incomparable values are also excluded from the restricted value space. This makes sense to me, the facets define the valid value space, not the invalid one, so a value that can't be proven within bounds should be rejected. I tested with both Xerces and libxml2. They both reject timezone indeterminate values against ordering facets.

aaroncameron-wk: Good point. I didn't catch that pattern for duration/float/double. Updated.

arelle/XmlValidate.py
```diff
@@ -451,6 +452,64 @@ def fractionValidateValue(value: str, fractionValue: tuple[str, str]) -> XmlVali
...
+
+def _comparableInstant(value: datetime.datetime | datetime.time) -> datetime.datetime:
+    # Express an xs:dateTime/date/time value as a single datetime so values can be
+    # ordered: timezone-aware values are normalized to (naive) UTC; an xs:time is
+    # anchored to an arbitrary fixed date (XSD Datatypes 3.2.8).
+    # Timezone-naive values are left as-is.
+    if isinstance(value, datetime.datetime):
+        dt = value
+    else:  # datetime.time (xs:time)
+        dt = datetime.datetime(2000, 1, 1, value.hour, value.minute, value.second, value.microsecond, value.tzinfo)
```
austinmatherne-wk, line 468: I think anchoring time to an arbitrary fixed date is safe today because all callers guarantee the facet and value share the same base type (both values are time and fixed to the same arbitrary date). But `validateValueString` doesn't enforce that itself. If a future caller passes a time value with a datetime facet (or vice versa), `_orderedComparison` could silently produce wrong results by converting the time value to a datetime with the arbitrary date and comparing that to the real datetime value.

I think `validateValueString` should validate facet value types against `baseXsdType` at the entry point so mismatches are caught before any comparison logic runs, but that's beyond the scope of this PR. We can address that in a followup. You could add a type guard at the top of `_orderedComparison` to be safe, but this looks fine as is for existing usage.

austinmatherne-wk (dismissed):

austinmatherne-wk (approved):

PR #2456, opened 2026-07-01
Type Hints: arelle.ViewWinDiffs 

volodymyrtur-wk:
#### Description of change

#### Steps to Test
CI
**review**:
@Arelle/arelle

### DIFF
--- a/arelle/ViewWinDiffs.py
+++ b/arelle/ViewWinDiffs.py
@@ -101,13 +101,6 @@ class ViewWinDiffs:
                         self.fromDTS.viewModelObject(modelObject.fromRelationship.objectId())  # type: ignore[union-attr]
                     else:
                         self.toDTS.viewModelObject(modelObject.fromRelationship.objectId())  # type: ignore[union-attr]
-                elif isinstance(modelObject, (ModelVersObject.ModelInstanceAspect, ModelVersObject.ModelRelatedConcept)):
-                    if modelObject.isFromDTS:
-                        self.fromDTS.viewModelObject(modelObject.sourceDtsObject.objectId())  # type: ignore[union-attr]
-                    else:
-                        self.toDTS.viewModelObject(modelObject.sourceDtsObject.objectId())  # type: ignore[union-attr]
-                elif modelObject.modelXbrl != self.modelVersReport:  # propogate event to versiong report's modelXbrl
-                    self.modelVersReport.viewModelObject(modelObject)  # type: ignore[arg-type,union-attr]
             except Exception:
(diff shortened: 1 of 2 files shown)


arelle/ViewWinDiffs.py
```diff
@@ -86,16 +98,16 @@ def viewModelObject(self, modelObject):
                         self.toDTS.viewModelObject(toConcept.objectId())
                 elif isinstance(modelObject, ModelVersObject.ModelRelationships):
                     if modelObject.isFromDTS:
-                        self.fromDTS.viewModelObject(modelObject.fromRelationship.objectId())
+                        self.fromDTS.viewModelObject(modelObject.fromRelationship.objectId())  # type: ignore[union-attr]
                     else:
-                        self.toDTS.viewModelObject(modelObject.fromRelationship.objectId())
-                elif isinstance(modelObject, (ModelVersObject.ModelInstanceAspect, ModelVersObject.ModelInstanceMemberAspect)):
+                        self.toDTS.viewModelObject(modelObject.fromRelationship.objectId())  # type: ignore[union-attr]
+                elif isinstance(modelObject, (ModelVersObject.ModelInstanceAspect, ModelVersObject.ModelRelatedConcept)):
```
aaroncameron-wk: So `ModelInstanceMemberAspect` hasn't existed since https://github.com/Arelle/Arelle/commit/24c0be9d798ba58c63049ccbd76c198ccbbff36e, which suggests any time this code has been running an exception would be thrown here and ignored. So this risk of "breaking" anything here is low, but based on the commit it seems `ModelInstanceMemberAspect` was effectively replaced with not just `ModelRelatedConcept`, but also:
- `ModelAspectProperty`
- `ModelMeasureAspect`
- `ModelPeriodAspect`

Without a better understanding of this diff tooling, I can't say for sure what's "correct", but just thought I'd point that out. The correct thing might also be to remove this `elif` and the one below it, since that is effectively what has been running for 10+ years.

aaroncameron-wk (dismissed):

austinmatherne-wk (dismissed):

arelle/ViewWinDiffs.py
volodymyrtur-wk: I saw that commit, and based on it, I decided to choose the `ModelRelatedConcept` class because the `ModelInstanceMemberAspect` was renamed to it. I'll remove those two `elif` blocks

aaroncameron-wk (approved):

austinmatherne-wk (approved):

PR #2460, opened 2026-07-01
Fix IsoDuration.__gt__ seconds/microseconds tie-break

aaroncameron-wk:
Part of #2445

## Fix 6 — `xs:duration` greater-than comparison (`IsoDuration.__gt__`)

**Where:** `IsoDuration.__gt__` in `arelle/ModelValue.py` (the value object for
`xs:duration`). This underpins the ordering-facet checks added in Fix 3.

**Symptom:** for two durations with an equal "average days" component (i.e. equal
years/months/days) that differ only in the time-of-day part, `a > b` always returned
`False` — even when `a` was strictly greater. A copy/paste error tested the wrong
branch condition:

```python
if self.avgdays > other.avgdays:
    return True
elif self.avgdays > other.avgdays:        # bug: duplicates the '>' test
    … seconds / microseconds tie-break …  # therefore unreachable
return False
```

The second condition can never be true after the first, so the seconds /
microseconds tie-break was dead code and `__gt__` under-reported greater-than. (As a
consequence `maxInclusive`/`maxExclusive` bounds on `duration`, which use `>` / `>=`,
were not enforced when the bound and value shared the same year/month/day part.)

**Change:** mirror the already-correct `__lt__`: test equality in the `elif`, and add
the same `isinstance` guard `__lt__` uses.

```python
elif self.avgdays == other.avgdays:       # fixed: equal -> fall through to tie-break
```

**Normative basis:**

- XSD 2 §3.2.6.2 *Order relation on duration* defines that durations compare by their
  effect when added to specific reference `dateTime`s. Two durations differing only
  in seconds (e.g. `…T10H36M30S` vs `…T10H36M29S`, with identical
  years/months/days) order unambiguously: the `…30S` value is the greater under any
  reference instant. <https://www.w3.org/TR/xmlschema-2/#duration-order>

- §4.3.7 *maxInclusive*: a value is facet-valid iff `value ≤ maxInclusive`; a value
  strictly greater than the bound must be rejected.
  <https://www.w3.org/TR/xmlschema-2/#rf-maxInclusive>

  (Arelle approximates the partial duration order with an "average days" magnitude —
  see the `IsoDuration` docstring — but the seconds tie-break is still required for
  durations that agree on years/months/days, which is exactly the case this fix
  restores.)

**Independent check:** with `maxInclusive = P2004Y12M06DT10H36M29S` and the candidate
`P2004Y12M06DT10H36M30S`, the two agree on every field except seconds (30 vs 29);
by the §3.2.6.2 order relation the candidate is greater, so `value ≤ maxInclusive`
is false and the value must be rejected. Symmetrically, `…M29S` itself (equal to the
bound) and `…M28S` (less than the bound) remain valid.

#### Steps to Test
- CI

**review**:
@Arelle/arelle

### DIFF
--- a/arelle/ModelValue.py
+++ b/arelle/ModelValue.py
@@ -1013,7 +1013,7 @@ class IsoDuration(isodate.Duration): # type: ignore[misc]
     def __gt__(self,other: Any) -> bool:
         if self.avgdays > other.avgdays:
             return True
-        elif self.avgdays > other.avgdays:
+        elif self.avgdays == other.avgdays:
             if self.tdelta.seconds > other.tdelta.seconds:
                 return True
             elif self.tdelta.seconds == other.tdelta.seconds:
--- a/arelle/ModelValue.py
+++ b/arelle/ModelValue.py
@@ -1011,6 +1011,9 @@ class IsoDuration(isodate.Duration): # type: ignore[misc]
     def __le__(self, other: Any) -> bool:
         return self.__lt__(other) or self.__eq__(other)
     def __gt__(self,other: Any) -> bool:
+        if not isinstance(other, IsoDuration):
+            return NotImplemented
+
         if self.avgdays > other.avgdays:
             return True
         elif self.avgdays == other.avgdays:
--- a/tests/unit_tests/arelle/test_xmlvalidate.py
+++ b/tests/unit_tests/arelle/test_xmlvalidate.py
@@ -1217,3 +1217,23 @@ class TestTimezoneValidation:
         result = validateValueString(base_xsd_type, value)
         assert result.xValid == INVALID
         assert not result.isXValid
+
+
+class TestIsoDurationComparison:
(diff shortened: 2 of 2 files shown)


volodymyrtur-wk (dismissed):

arelle/ModelValue.py
```diff
@@ -1013,7 +1013,7 @@ def __le__(self, other: Any) -> bool:
     def __gt__(self,other: Any) -> bool:
         if self.avgdays > other.avgdays:
             return True
-        elif self.avgdays > other.avgdays:
+        elif self.avgdays == other.avgdays:
```
austinmatherne-wk, line 1019: > Change: mirror the already-correct __lt__: test equality in the elif, and add
the same isinstance guard __lt__ uses.

The `isinstance` guard hasn't been added?

austinmatherne-wk (approved):

PR #2462, opened 2026-07-02
Type Hints: arelle.ViewWinRelationshipSet 

volodymyrtur-wk:
#### Description of change

#### Steps to Test
CI
**review**:
@Arelle/arelle

### DIFF
--- a/arelle/ViewWinRelationshipSet.py
+++ b/arelle/ViewWinRelationshipSet.py
@@ -70,8 +70,6 @@ def viewRelationshipSet(
 
 
 class ViewRelationshipSet(ViewWinTree.ViewTree):
-    tabName: str
-
     def __init__(
         self,
         modelXbrl: ModelXbrl,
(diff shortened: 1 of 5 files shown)


aaroncameron-wk (dismissed):

arelle/ViewWinRelationshipSet.py
```diff
@@ -43,14 +70,29 @@ def viewRelationshipSet(modelXbrl, tabWin, arcrole,
 
 
 class ViewRelationshipSet(ViewWinTree.ViewTree):
-    def __init__(self, modelXbrl, tabWin,
-                 arcrole, linkrole=None, linkqname=None, arcqname=None, lang=None,
-                 treeColHdr=None, showLinkroles=True, showRelationships=True, showColumns=True,
-                 expandAll=False, hasTableIndex=False):
-        if isinstance(arcrole, (list,tuple)):
+    tabName: str
```
austinmatherne-wk: `tabName` is never set as an attribute on the class?

austinmatherne-wk (commented):
Can you rebase/merge master into this branch? I suspect that the other recently merged type hint PRs will cause mypy to complain about unused `type: ignore` comments when this is merged with latest master

austinmatherne-wk (approved):

PR #2466, opened 2026-07-07
Match enumeration facets in the value space, not lexically

aaroncameron-wk:
Part of #2445

## Fix 8 — `enumeration` is matched in the value space, not lexically

**Where:** the `enumeration` facet check in `_validateValueStringOrRaise`, plus a
new `_EnumerationFacet` (a `dict` subclass) in `arelle/ModelDtsObject.py` that
`ModelType.facets` now uses to hold each enumeration and carry its cached value
space.

**Symptom:** enumeration membership was tested as `value not in facets["enumeration"]`
— a comparison of the *lexical* string against the lexical forms of the enumeration
members. That wrongly rejects a value that equals a member in the **value space** but
is spelled differently. For example, a `dateTime`-derived type whose only member is
`2002-01-01T12:01:01-00:00` rejected the values `2002-01-01T12:01:01+00:00` and
`2002-01-01T12:01:01Z`, even though all three denote the same instant. The same class
of false rejection applies to any type whose lexical space is many-to-one onto its
value space (`decimal` `1.0` vs `1`, `boolean` `1` vs `true`, etc.).

**Change:** keep the cheap lexical membership test as a fast path, but when it misses,
fall back to a value-space comparison. The check is moved to after `xValue` (the parsed
value-space object) is computed. On a miss the enumeration's members are parsed once into
a map from each member's parsed `xValue` to its lexical form; membership is then a lookup
of the candidate's `xValue` in that map. Each member is parsed through the same datatype
path (facets omitted, so the parse cannot recurse back into the enumeration facet) and
with its **own schema facet's** `nsmap` (a `QName`-lexical member's namespace bindings are
fixed at the schema, not the instance)

The parsed value space is built **lazily** (only on a lexical miss) and cached on the
enumeration facet object itself: `ModelType.facets` returns the enumeration as an
`_EnumerationFacet`, a `dict` subclass that stores the map in a `__slots__` attribute
(`valueSpace`) kept off the dict's own keys/items. Because `ModelType.facets` is memoized,
the map is computed at most once per type, while the facet still behaves as a plain
`dict[str, facetElt]` for existing consumers (type-equivalence checks, ViewX rendering, DB
export, sample-instance generation). Synthetic set/dict enumerations built elsewhere in
the module lack the attribute (the cache write is swallowed) and simply recompute.

Because the lexical fast path still accepts every previously-accepted value, the change
can only *accept* values that were formerly (wrongly) rejected; it never newly rejects a
value. Robustness: a member that fails to parse is skipped rather than aborting the check,
and an unhashable `xValue` (e.g. a list-valued type) falls back to a linear equality scan
instead of raising.

**Normative basis:**

- XSD 2 §4.3.5 *enumeration*: "enumeration constrains the **value space** to a specified
  set of values." Its facet-validity rule is value-space membership: a value is
  facet-valid with respect to enumeration iff the value is **equal** (in the value space)
  to one of the enumerated values — not iff its lexical form matches.
  <https://www.w3.org/TR/xmlschema-2/#rf-enumeration>

- Equality is the datatype's value-space identity, which is independent of lexical
  form. For `dateTime`, §3.2.7 *Order relation on dateTime* makes timezoned values that
  denote the same instant equal, and §3.2.7.1 notes that the timezones `Z`, `+00:00`
  and `-00:00` are equivalent. <https://www.w3.org/TR/xmlschema-2/#dateTime-order>
  Likewise `decimal` (§3.2.3) treats `1`, `1.0`, `1.00` as the same value, and `boolean`
  (§3.2.2) has the two values denoted by `{true, 1}` and `{false, 0}`.

- XSD 1.1 D §3.2.7 / §4.3.5 state the same value-space semantics for enumeration.
  <https://www.w3.org/TR/xmlschema11-2/#rf-enumeration>

**Independent check:** take a type derived from `dateTime` with the single enumerated
member `2002-01-01T12:01:01-00:00`. By the §3.2.7 order relation, `-00:00` denotes UTC,
so the value `2002-01-01T12:01:01+00:00` (also UTC) is **equal** to the member in the
value space; by §4.3.5 it is therefore facet-valid and must be accepted. A value at a
different instant (e.g. `2002-01-01T12:01:02+00:00`) is not equal to any member and must
still be rejected.

#### Steps to Test
CI

**review**:
@Arelle/arelle

### DIFF
--- a/arelle/XmlValidate.py
+++ b/arelle/XmlValidate.py
@@ -192,9 +192,40 @@ _XSD_TYPE_INHERENT_INCLUSIVE_BOUNDS: dict[str, tuple[int | None, int | None]] = 
 }
 _INTEGER_BASE_XSD_TYPES = frozenset(_XSD_TYPE_INHERENT_INCLUSIVE_BOUNDS.keys() | {"integer"})
 
+
+class _EnumerationFacetMember:
+    """Stand-in for a schema enumeration facet element (a ``ModelObject``), used for
+    enumeration facets synthesized in this module (e.g. the predefined xml:space
+    attribute type, the whiteSpace facet's own permitted values) rather than derived
+    from a schema. Provides the ``nsmap`` a QName-lexical enumeration member would be
+    resolved against; empty by default since none of this module's synthetic
+    enumerations are QName-lexical types."""
+    __slots__ = ("nsmap",)
+
+    def __init__(self, nsmap: Mapping[str | None, str] | None = None) -> None:
+        self.nsmap = nsmap or {}
+
+
+class EnumerationFacet(dict[str, ModelObject | _EnumerationFacetMember]):
+    """dict of an enumeration facet's lexical value -> facet element (a schema
+    ``ModelObject`` for schema-derived enumerations, or an ``_EnumerationFacetMember``
+    for enumerations synthesized in this module), with a lazily-populated, out-of-band
+    value-space cache (xValue -> lexical value) attached via the ``valueSpace``
+    attribute. Kept off the dict's own keys/items so existing consumers that treat this
+    as a plain ``dict[str, ModelObject]`` (equivalence checks, ViewX rendering, etc.)
+    are unaffected."""
+    __slots__ = ("valueSpace",)
+    valueSpace: dict[Any, str] | None
+
+    def __init__(self, *args: Any, **kwargs: Any) -> None:
+        super().__init__(*args, **kwargs)
+        self.valueSpace = None
+
+
 predefinedAttributeTypes = {
     qname("{http://www.w3.org/XML/1998/namespace}xml:lang"):("languageOrEmpty",None),
-    qname("{http://www.w3.org/XML/1998/namespace}xml:space"):("NCName",{"enumeration":{"default","preserve"}})}
+    qname("{http://www.w3.org/XML/1998/namespace}xml:space"):("NCName", {"enumeration": EnumerationFacet(
+        {"default": _EnumerationFacetMember(), "preserve": _EnumerationFacetMember()})})}
 xAttributesSharedEmptyDict: dict[str, ModelAttribute] = {}
 
 def validate(
@@ -778,39 +809,26 @@ def _validateValueStringOrRaise(
             # XSD 1.0 Datatypes 4.3.5: the enumeration facet constrains the value space, so
             # a value is facet-valid when it equals a member in the value space even if its
             # lexical form differs (e.g. dateTime 12:01:01+00:00 vs -00:00, decimal 1.0 vs 1).
+            # Every enumeration facet (schema-derived, via ModelType.constrainingFacets, or
+            # synthesized in this module, e.g. for whiteSpace) is an EnumerationFacet: a
+            # dict of lexical value -> facet element, with a lazily-populated valueSpace cache.
             enumeration = facets["enumeration"]
-            # Only a schema-derived enumeration facet (a dict of lexical value -> facetElt,
-            # possibly an _EnumerationFacet) supports the lazily-populated valueSpace cache;
-            # synthetic enumerations built in this module (e.g. for whiteSpace) are plain
-            # sets/dicts without facetElts and always fall through to a fresh parse below.
-            valueSpace = getattr(enumeration, "valueSpace", None)
+            valueSpace = enumeration.valueSpace
             if valueSpace is None:
                 valueSpace = {}
-                members = (
-                    enumeration.items()
-                    if isinstance(enumeration, dict)
-                    else ((m, None) for m in enumeration)
-                )
-                for member, facetElt in members:
+                for member, facetElt in enumeration.items():
                     try:
                         # Use the enumeration facet's own schema-document nsmap (not the
                         # validated instance's) since a QName-lexical enumeration member's
                         # namespace bindings are fixed at the schema, per XSD Part 2 §3.2.18.
-                        memberNsmap = facetElt.nsmap if facetElt is not None else nsmap
-                        parsedMember = _validateValueStringOrRaise(baseXsdType, member, nsmap=memberNsmap)
+                        parsedMember = _validateValueStringOrRaise(baseXsdType, member, nsmap=facetElt.nsmap)
                         valueSpace[_hashableXValue(parsedMember.xValue)] = member
                     except (ValueError, InvalidOperation, TypeError):
                         pass
-                try: # only an _EnumerationFacet (schema-derived enumeration) supports this
-                    enumeration.valueSpace = valueSpace
-                except AttributeError:
-                    pass
+                enumeration.valueSpace = valueSpace
             found = _hashableXValue(xValue) in valueSpace
             if not found:
-                raise ValueError("{0} is not in {1}".format(value, (
-                    facets["enumeration"].keys()
-                    if isinstance(facets["enumeration"], dict)
-                    else facets["enumeration"])))
+                raise ValueError("{0} is not in {1}".format(value, enumeration.keys()))
     return XmlValidationResult(sValue=sValue, xValue=xValue, xValid=xValid)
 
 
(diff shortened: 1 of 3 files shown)


arelle/XmlValidate.py
```diff
@@ -767,6 +765,46 @@ def _validateValueStringOrRaise(
...
+                members = (
+                    enumeration.items()
+                    if isinstance(enumeration, dict)
+                    else ((m, None) for m in enumeration)
+                )
+                for member, facetElt in members:
+                    try:
+                        # Use the enumeration facet's own schema-document nsmap (not the
+                        # validated instance's) since a QName-lexical enumeration member's
+                        # namespace bindings are fixed at the schema, per XSD Part 2 §3.17.2.
```
austinmatherne-wk: @aaroncameron-wk fyi hallucinated spec section, should review this carefully.

tests/unit_tests/arelle/test_xmlvalidate.py
```diff
@@ -558,6 +559,170 @@ def test_validateValue_facets_enumeration(value: str, expected: tuple):
...
+        ("dateTime", "2002-01-01T12:01:01-00:00", "2002-01-01T12:01:02+00:00", INVALID),  # different instant
+        # date: timezone spellings that denote the same day are equal
+        ("date", "2002-01-01-00:00", "2002-01-01Z", VALID),
+        ("date", "2002-01-01-00:00", "2002-01-01+00:00", VALID),
+        ("date", "2002-01-01-00:00", "2002-01-02Z", INVALID),
+        # time: Z and -00:00 denote the same instant of day
+        ("time", "12:00:00-00:00", "12:00:00Z", VALID),
+        ("time", "12:00:00-00:00", "13:00:00Z", INVALID),
+        # gYearMonth discards the timezone at parse time, so tz spellings collapse
+        ("gYearMonth", "2002-01Z", "2002-01", VALID),
```
austinmatherne-wk: Preexisting behavior, but this test is encoding incorrect behavior.

tests/unit_tests/arelle/test_xmlvalidate.py
```diff
@@ -558,6 +559,170 @@ def test_validateValue_facets_enumeration(value: str, expected: tuple):
...
+        ("double", "1.5", "1.50", VALID),
+        ("double", "1.5", "1.6", INVALID),
+        # integer: leading sign and leading zeros are not part of the value
+        ("integer", "1", "+1", VALID),
+        ("integer", "1", "01", VALID),
+        ("integer", "1", "2", INVALID),
+        # boolean: {true, 1} and {false, 0}
+        ("boolean", "1", "true", VALID),
+        ("boolean", "1", "1", VALID),
+        ("boolean", "1", "false", INVALID),
```
austinmatherne-wk: Enumeration isn't a permitted constraining facet of boolean.

arelle/XmlValidate.py
aaroncameron-wk: Good catch. The conclusion was correct, but the citation was not. I've corrected it and [double checked](https://www.w3.org/TR/xmlschema-2/#QName):
> The mapping between literals in the [·lexical space·](https://www.w3.org/TR/xmlschema-2/#dt-lexical-space) and values in the [·value space·](https://www.w3.org/TR/xmlschema-2/#dt-value-space) of QName requires a namespace declaration to be in scope for the context in which QName is used.

tests/unit_tests/arelle/test_xmlvalidate.py
aaroncameron-wk: I've updated these tests to test correct behavior and [added](https://github.com/Arelle/Arelle/issues/2445) Fix 9 to address the incorrect behavior.

arelle/XmlValidate.py
```diff
@@ -767,6 +765,46 @@ def _validateValueStringOrRaise(
...
                     raise ValueError(" <= minExclusive {0}".format(facets["minExclusive"]))
+        if facets and "enumeration" in facets and value not in facets["enumeration"]:
+            # XSD 1.0 Datatypes 4.3.5: the enumeration facet constrains the value space, so
+            # a value is facet-valid when it equals a member in the value space even if its
+            # lexical form differs (e.g. dateTime 12:01:01+00:00 vs -00:00, decimal 1.0 vs 1).
+            enumeration = facets["enumeration"]
+            # Only a schema-derived enumeration facet (a dict of lexical value -> facetElt,
+            # possibly an _EnumerationFacet) supports the lazily-populated valueSpace cache;
+            # synthetic enumerations built in this module (e.g. for whiteSpace) are plain
+            # sets/dicts without facetElts and always fall through to a fresh parse below.
```
brettkail-wk: Out of curiosity, why not change those synthetic cases to use `_EnumerationFacet` (with a real `dict`, with an `_EnumerationFacetMember` value that defines `nsmap`) and remove all the special cases?

arelle/XmlValidate.py
```diff
@@ -767,6 +765,46 @@ def _validateValueStringOrRaise(
...
+                    except (ValueError, InvalidOperation, TypeError):
+                        pass
+                try: # only an _EnumerationFacet (schema-derived enumeration) supports this
+                    enumeration.valueSpace = valueSpace
+                except AttributeError:
+                    pass
+            try:
+                found = xValue in valueSpace
+            except TypeError: # xValue is unhashable (e.g. a list); fall back to linear equality scan
+                found = any(xValue == v for v in valueSpace)
```
brettkail-wk: Out of curiosity, did you consider converting list->tuple when building/looking up from the dict?

arelle/XmlValidate.py
aaroncameron-wk: That's a good idea to simplify this code. I've implemented it.

arelle/XmlValidate.py
aaroncameron-wk: No, and that's a minor change so I'll implement it here.

arelle/XmlValidate.py
```diff
@@ -767,6 +805,30 @@ def _validateValueStringOrRaise(
...
+            if valueSpace is None:
+                valueSpace = {}
+                for member, facetElt in enumeration.items():
+                    try:
+                        # Use the enumeration facet's own schema-document nsmap (not the
+                        # validated instance's) since a QName-lexical enumeration member's
+                        # namespace bindings are fixed at the schema, per XSD Part 2 §3.2.18.
+                        parsedMember = _validateValueStringOrRaise(baseXsdType, member, nsmap=facetElt.nsmap)
+                        valueSpace[_hashableXValue(parsedMember.xValue)] = member
+                    except (ValueError, InvalidOperation, TypeError):
```
brettkail-wk: Out of curiosity, is `TypeError` still required as of `_hashableXValue`?

I assume the `except` `ValueError` is still required for `_validateValueStringOrRaise`.  I wonder if it's useful to narrow `try:` `parsedMember = _validateValueStringOrRaise` `except ValueError:` `continue` to make that more obvious?

Out of curiosity, what does the `InvalidOperation` do?

arelle/XmlValidate.py
```diff
@@ -834,8 +896,8 @@ def validateValue(
         elt.sValue = sValue
 
 
-def _facetTypeAndFacets(facetName: str, baseXsdType: str) -> tuple[str, dict[str, int | set[str]] | None]:
-    facets: dict[str, int | set[str]] | None
+def _facetTypeAndFacets(facetName: str, baseXsdType: str) -> tuple[str, dict[str, int | EnumerationFacet] | None]:
+    facets: dict[str, int | EnumerationFacet] | None
```
volodymyrtur-wk: You can simplify the code by defining the `facets` variable as `None` and removing pieces of code like `facets = None` below inside conditions:

```python
facets: dict[str, int | EnumerationFacet] | None = None
```

volodymyrtur-wk (dismissed):

arelle/XmlValidate.py
aaroncameron-wk: Those are mostly out of defensiveness rather than a specific known scenario. 

For `InvalidOperation`, I considered wrapping the [only possible call site](https://github.com/Arelle/Arelle/blob/7715cbf6a240db21c25023941c8d8713ef436381/arelle/XmlValidate.py#L633-L635) that can trigger it and re-raising as `ValueError`, but decided against it because it has slightly larger surface area. I'm not actually able to trigger `InvalidOperation` there because the pattern check is strict enough that it appears impossible, but catching `InvalidOperation` here is consistent with other callers [here](https://github.com/Arelle/Arelle/blob/7715cbf6a240db21c25023941c8d8713ef436381/arelle/XmlValidate.py#L542-L544) and [here](https://github.com/Arelle/Arelle/blob/7715cbf6a240db21c25023941c8d8713ef436381/arelle/XmlValidate.py#L792-L794).

`TypeError` is not as defensible. It really just protects against an unhashable value other than `list` passed to `_hashableXValue`, but we probably want that to fail loudly rather than silence it, so I'll take your suggestion of narrowing the try/except.

brettkail-wk (approved):

austinmatherne-wk (approved):

PR #2469, opened 2026-07-10
Type Hints: arelle.ViewWinFactGrid 

volodymyrtur-wk:
#### Description of change

#### Steps to Test
CI
**review**:
@Arelle/arelle

### DIFF
--- a/arelle/ViewWinFactGrid.py
+++ b/arelle/ViewWinFactGrid.py
@@ -25,8 +25,8 @@ from arelle.UiUtil import (
 from arelle.typing import TypeGetText
 
 if TYPE_CHECKING:
-    from tkinter import Widget, Event
-    from tkinter.ttk import Treeview
+    from tkinter import Event
+    from tkinter.ttk import Treeview, Notebook
     from arelle.ModelXbrl import ModelXbrl
     from arelle.ModelValue import QName
     from arelle.ModelDtsObject import ModelConcept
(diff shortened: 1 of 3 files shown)


arelle/ViewWinFactGrid.py
```diff
@@ -30,93 +64,115 @@ def viewFactsGrid(modelXbrl, tabWin, header="Fact Grid", arcrole=XbrlConst.paren
...
         # relationship set based on linkrole parameter, to determine applicable linkroles
-        relationshipSet = self.modelXbrl.relationshipSet(self.arcrole, self.linkrole, self.linkqname, self.arcqname)
+        relationshipSet = self.modelXbrl.relationshipSet(self.arcrole, self.linkrole, self.linkqname, self.arcqname)  # type: ignore[union-attr,arg-type]
         if not relationshipSet:
-            self.modelXbrl.modelManager.addToLog(_("no relationships for {0}").format(self.arcrole))
+            self.modelXbrl.modelManager.addToLog(_("no relationships for {0}").format(self.arcrole))  # type: ignore[union-attr]
             return False
 
-        factConcepts = set(fact.concept for fact in self.modelXbrl.factsInInstance).append(fact)
+        factConcepts = set(fact.concept for fact in self.modelXbrl.factsInInstance).append(fact)  # type: ignore[union-attr,attr-defined,name-defined]
```
austinmatherne-wk: I think this entire module is unused/dead code (much of this code would crash if actually run). We should probably just delete this entire module, but if not I think the intent of this is: `factConcepts = {fact.concept for fact in self.modelXbrl.factsInInstance if fact.concept is not None}`

austinmatherne-wk (dismissed):

austinmatherne-wk (commented):
@volodymyrtur-wk can you update this branch with the latest changes on master? It has a couple of mypy errors when combined with the other recently merged type hint PRs.

austinmatherne-wk (approved):

PR #2471, opened 2026-07-10
Type Hints: arelle.ViewUtil 

volodymyrtur-wk:
#### Description of change

#### Steps to Test
CI
**review**:
@Arelle/arelle

### DIFF
--- a/arelle/ViewUtil.py
+++ b/arelle/ViewUtil.py
@@ -7,7 +7,7 @@ from __future__ import annotations
 import os
 from collections import Counter
 from operator import attrgetter
-from typing import TYPE_CHECKING
+from typing import TYPE_CHECKING, Sequence
 
 from arelle import XbrlConst, XmlUtil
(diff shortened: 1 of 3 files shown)


aaroncameron-wk (dismissed):

arelle/ViewUtil.py
```diff
@@ -97,7 +99,13 @@ def referenceURI(concept):
     )
 
 
-def groupRelationshipSet(modelXbrl, arcrole, linkrole, linkqname, arcqname):
+def groupRelationshipSet(
+    modelXbrl: ModelXbrl,
+    arcrole: str | list[str] | tuple[str, list[str]],
```
austinmatherne-wk: This doesn't look quite right:

```suggestion
    arcrole: str | tuple[str, Sequence[str]],
```

austinmatherne-wk (approved):

PR #2475, opened 2026-07-13
Type Hints: arelle.ViewWinFactList 

volodymyrtur-wk:
#### Description of change

#### Steps to Test
CI
**review**:
@Arelle/arelle

### DIFF
--- a/arelle/ViewWinFactList.py
+++ b/arelle/ViewWinFactList.py
@@ -3,7 +3,8 @@ See COPYRIGHT.md for copyright information.
 """
 from __future__ import annotations
 
-from typing import TYPE_CHECKING, Any, Sequence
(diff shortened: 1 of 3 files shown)


arelle/ViewWinFactList.py
```diff
@@ -1,12 +1,28 @@
-'''
+"""
 See COPYRIGHT.md for copyright information.
-'''
+"""
+from __future__ import annotations
+
+from typing import TYPE_CHECKING, Any, Sequence
```
austinmatherne-wk: nit: should use `Sequence` from `collections.abc`

austinmatherne-wk (dismissed):

austinmatherne-wk (approved):

PR #2480, opened 2026-07-15
Preserve timezone on gYearMonth/gYear/gMonthDay/gMonth/gDay

aaroncameron-wk:
Part of https://github.com/Arelle/Arelle/issues/2445

## Fix 9 — the `g*` date/time types (`gYearMonth`, `gYear`, `gMonthDay`, `gMonth`, `gDay`) silently discard their timezone

**Symptom:** the lexical patterns for all five types capture an optional trailing
timezone (`Z`, `+hh:mm`, or `-hh:mm`) — e.g. `"gYearMonth": re_compile(r"...(Z|(\+|-)
(...))?$")` — and `_validateValueStringOrRaise` even unpacks the timezone groups out of
the regex match (`month, day, zSign, zHrMin, zHr, zMin = match.groups()`), but the
captured groups were never used: every constructor call passed only the numeric
fields (`gMonthDay(month, day)`, `gYear(year)`, etc.), and none of the five classes had
anywhere to put a timezone even if one had been passed. Consequently:

- **Equality** (`__eq__`, and therefore the `EnumerationFacet` value-space lookup added
  in Fix 8) compared only the numeric fields, so an untimezoned value, a UTC value
  (`Z`/`+00:00`/`-00:00`), and a value in *any other timezone* with the same numeral
  fields were all treated as the identical value — e.g. `gYear` parsed from `"2002"`,
  `"2002Z"`, and `"2002+09:00"` compared equal to each other, even though the first has
  no defined instant at all and the third denotes a starting instant 9 hours away from
  the second's.
- **Ordering** (`__lt__` etc., and therefore the `maxInclusive`/`maxExclusive`/
  `minInclusive`/`minExclusive` facet checks added in Fix 3) had the same blind spot:
  `_orderedComparison` never saw a `TypeError` for a timezone mismatch on these types, 
  so a genuinely indeterminate order (per §3.2.7.4, applied to these five types by 
  §3.2.10.2/§3.2.11.2/§3.2.12.2/§3.2.13.2/§3.2.14.2) was silently treated as determined 
  by the numeral fields alone.

**Change:** give each of the five classes an optional `tzinfo: datetime.timezone | None`
field, populate it at parse time from the already-captured (previously discarded) regex
group, and route equality/ordering through helpers that treat a value's
timezone-presence as part of its identity, exactly as `dateTime`/`date`/`time` already
do.

**Normative basis:**

- XSD 2 §3.2.7 defines `dateTime` values as having "a boolean timezoned
  property" that is part of the value itself — not a display detail. §3.2.10 (`gYearMonth`),
  §3.2.11 (`gYear`), §3.2.12 (`gMonthDay`), §3.2.13 (`gDay`), and §3.2.14 (`gMonth`) each
  state: "Since the lexical representation allows an optional time zone indicator, \[this
  type's] values are partially ordered because it may not be possible to unequivocally
  determine the order of two values one of which has a time zone and the other does
  not." <https://www.w3.org/TR/xmlschema-2/#gYearMonth> ·
  <https://www.w3.org/TR/xmlschema-2/#gYear> ·
  <https://www.w3.org/TR/xmlschema-2/#gMonthDay> ·
  <https://www.w3.org/TR/xmlschema-2/#gDay> · <https://www.w3.org/TR/xmlschema-2/#gMonth>

- Each of those same sections continues: "the order relation on \[this type's] values is
  the order relation on their starting instants[, discussed in] Order relation on
  dateTime (§3.2.7.4)" — i.e. these five types do not get their own bespoke order
  relation; they explicitly reuse dateTime's, timezone handling included.
  <https://www.w3.org/TR/xmlschema-2/#dateTime-order>

- §3.2.7.3 *Timezones*: the lexical timezone ranges over `+14:00` to `-14:00` (a `hh:mm`
  duration with `hh <= 14`, and `hh == 14` only paired with `mm == 0`), which is the
  origin of the `+-14h` uncertainty band already established for `dateTime`/`date`/`time`
  in Fix 3a and reused here unchanged. <https://www.w3.org/TR/xmlschema-2/#dateTime>

- §3.2.12 *gMonthDay* and §3.2.13 *gDay* explicitly sanction picking an arbitrary
  reference year/month to reason about ordering — "in an arbitrary leap year" and "in an
  arbitrary month that has 31 days," respectively — which is exactly the anchoring
  `_comparableInstant` performs. <https://www.w3.org/TR/xmlschema-2/#gMonthDay> ·
  <https://www.w3.org/TR/xmlschema-2/#gDay>

- §3.2.7.1 *Lexical representation* / *Timezones*: `Z`, `+00:00`, and `-00:00` are three
  spellings of the same (UTC) timezone, so a value carrying any of the three must compare
  equal to a value carrying either of the others — the same equivalence already relied on
  for `dateTime` in Fix 8 now also holds for these five types.

#### Steps to Test
CI

**review**:
@Arelle/arelle

### DIFF
--- a/arelle/ModelValue.py
+++ b/arelle/ModelValue.py
@@ -671,6 +671,11 @@ class Time(datetime.time):
         return time
 
 
+GTYPE_ANCHOR_YEAR = 2000  # arbitrary leap year (XSD Datatypes 3.2.12: "in an arbitrary leap year")
+GTYPE_ANCHOR_MONTH = 1  # arbitrary month with 31 days (XSD Datatypes 3.2.13)
+GTYPE_ANCHOR_DAY = 1  # arbitrary day <= 28
+
+
 def _gTypeEqual(
     selfFields: tuple[int, ...], selfTz: datetime.timezone | None,
     otherFields: tuple[int, ...], otherTz: datetime.timezone | None,
@@ -691,17 +696,19 @@ def _gTypeOrder(
     selfFields: tuple[int, ...], selfTz: datetime.timezone | None,
     otherFields: tuple[int, ...], otherTz: datetime.timezone | None,
 ) -> int:
-    """Order two like-typed g* values (gYear/gYearMonth/gMonthDay/gMonth/gDay), whose only
-    lexical content is numeric fields plus an optional timezone.
+    """Order two like-typed g* values whose only lexical content is numeric fields plus
+    an optional timezone -- for ``gYear``/``gYearMonth``/``gMonth`` *only*. (``gMonthDay``/
+    ``gDay`` instead compare real anchored ``datetime`` instants via ``_gMonthDayInstant``/
+    ``_gDayInstant`` below: see the note there.)
 
     Returns -1/0/1 for less-than/equal/greater-than. Raises ``TypeError`` when exactly one
     operand carries a timezone and the numeric fields are equal: over the +-14h timezone
     range (XSD Datatypes 3.2.7.3) that difference can never bridge a whole field unit (a
-    year/month/day), so unequal ``selfFields``/``otherFields`` alone always determines the
+    year or month), so unequal ``selfFields``/``otherFields`` alone always determines the
     order; only a field tie leaves the order genuinely dependent on the missing timezone
-    (3.2.7.4, applied to g* types per 3.2.10.2/3.2.11.2/3.2.12.2/3.2.13.2/3.2.14.2), which
-    XmlValidate's ``_orderedComparison`` resolves with the +-14h straddling rule, mirroring
-    how it already handles offset-naive vs. offset-aware ``dateTime``/``date``/``time``.
+    (3.2.7.4, applied to g* types per 3.2.10.2/3.2.11.2/3.2.14.2), which XmlValidate's
+    ``_orderedComparison`` resolves with the +-14h straddling rule, mirroring how it
+    already handles offset-naive vs. offset-aware ``dateTime``/``date``/``time``.
     """
     if selfFields != otherFields:
         return -1 if selfFields < otherFields else 1
@@ -718,6 +725,14 @@ def _gTypeOrder(
     # occurs *earlier* in UTC: 00:00+05:00 is 19:00 UTC the previous day, earlier than
     # 00:00+04:00's 20:00 UTC.
     return -1 if selfOffset > otherOffset else 1
+
+
+def _gMonthDayInstant(month: int, day: int, tz: datetime.timezone | None) -> datetime.datetime:
+    return datetime.datetime(GTYPE_ANCHOR_YEAR, month, day, tzinfo=tz)
+
+
+def _gDayInstant(day: int, tz: datetime.timezone | None) -> datetime.datetime:
+    return datetime.datetime(GTYPE_ANCHOR_YEAR, GTYPE_ANCHOR_MONTH, day, tzinfo=tz)
 
 
 class gYearMonth:
@@ -775,7 +790,11 @@ class gMonthDay:
         self.month = int(month)
         self.day = int(day)
         self.tzinfo = tzinfo
-        self._hash = hash((self.month, self.day))
+        # Hashed via the anchor instant (not the raw fields): two different (month, day)
+        # pairs with different timezones can denote the *same* instant (see the note on
+        # _gTypeOrder), and Python's own aware-datetime hash already makes such equal
+        # instants hash equal regardless of offset, keeping __hash__/__eq__ consistent.
+        self._hash = hash(_gMonthDayInstant(self.month, self.day, self.tzinfo))
 
     def __repr__(self) -> str:
         return self.__str__()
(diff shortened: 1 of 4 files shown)


volodymyrtur-wk (dismissed):

arelle/XmlValidate.py
```diff
@@ -501,15 +502,31 @@ def fractionValidateValue(value: str, fractionValue: tuple[str, str]) -> XmlVali
...
+    # single datetime so values can be ordered: timezone-aware values are normalized to
+    # (naive) UTC. xs:time and the g* types have no full calendar date of their own, so
+    # each is anchored to an arbitrary fixed point (XSD Datatypes 3.2.8 for time;
+    # 3.2.10.2/3.2.11.2/3.2.12.2/3.2.13.2/3.2.14.2 analogously for the g* types via their
+    # "starting instant", 3.2.7.4). Timezone-naive values are left as-is.
     if isinstance(value, datetime.datetime):
         dt = value
-    else:  # datetime.time (xs:time)
+    elif isinstance(value, datetime.time):  # xs:time
         dt = datetime.datetime(2000, 1, 1, value.hour, value.minute, value.second, value.microsecond, value.tzinfo)
```
brettkail-wk: Out of curiosity why add `_GTYPE_ANCHOR_YEAR`/`_GTYPE_ANCHOR_31DAY_MONTH` but leave `2000, 1, 1` here?

(It's arbitrary, but January somehow seems more natural than December for the arbitrary 31-day month.)

arelle/XmlValidate.py
```diff
@@ -764,26 +782,28 @@ def _validateValueStringOrRaise(
                     xValue = time(value, castException=ValueError)
                     sValue = value
                 elif baseXsdType == "gMonthDay":
+                    # zSign is the whole timezone token ("Z"/"+hh:mm"/"-hh:mm"/None); the
+                    # remaining groups decompose it further but aren't needed here.
```
brettkail-wk, line 785: Out of curiosity, are they used anywhere, or could the regexes be updated to use non-capturing groups?

aaroncameron-wk: The patterns themselves are used in [FunctionXs.py](https://github.com/Arelle/Arelle/blob/64d59575fda52c8a697f90a2c55a69c414569fa8/arelle/FunctionXs.py#L397), but the time capture groups are similarly not used. Still, I think for the same reason FunctionXs.py re-uses the patterns, Python API users may rely on it as well, so changing it risks a breaking change without significant gain.

arelle/ModelValue.py
```diff
@@ -671,10 +671,60 @@ def __new__(cls, hour: int = 0, minute: int = 0, second: int = 0, microsecond: i
...
+def _gTypeOrder(
+    selfFields: tuple[int, ...], selfTz: datetime.timezone | None,
+    otherFields: tuple[int, ...], otherTz: datetime.timezone | None,
+) -> int:
+    """Order two like-typed g* values (gYear/gYearMonth/gMonthDay/gMonth/gDay), whose only
+    lexical content is numeric fields plus an optional timezone.
+
+    Returns -1/0/1 for less-than/equal/greater-than. Raises ``TypeError`` when exactly one
+    operand carries a timezone and the numeric fields are equal: over the +-14h timezone
+    range (XSD Datatypes 3.2.7.3) that difference can never bridge a whole field unit (a
```
austinmatherne-wk, line 710: > over the +-14h timezone range (XSD Datatypes 3.2.7.3) that difference can never bridge a whole field unit

That assumes only one value has a timezone offset. If two values include an offset (+-28h combined) they can reach beyond day fields

arelle/ModelValue.py
```diff
@@ -719,9 +771,10 @@ def __hash__(self) -> int:
 
 
 class gMonthDay:
-    def __init__(self, month: int | str, day: int | str) -> None:
+    def __init__(self, month: int | str, day: int | str, tzinfo: datetime.timezone | None = None) -> None:
         self.month = int(month)
         self.day = int(day)
+        self.tzinfo = tzinfo
```
austinmatherne-wk, line 797: hash needs to be updated? Also str and repr?

arelle/ModelValue.py
```diff
@@ -671,10 +671,60 @@ def __new__(cls, hour: int = 0, minute: int = 0, second: int = 0, microsecond: i
...
 
+def _gTypeEqual(
+    selfFields: tuple[int, ...], selfTz: datetime.timezone | None,
+    otherFields: tuple[int, ...], otherTz: datetime.timezone | None,
+) -> bool:
+    # XSD Datatypes 3.2.7 gives dateTime (and, analogously per 3.2.10.2/3.2.11.2/etc, the
+    # g* types) a boolean "timezoned" property that is part of the value: a timezoned value
+    # is never equal to an untimezoned one, regardless of field values.
+    if selfFields != otherFields:
+        return False
```
austinmatherne-wk, line 691: I think this also needs to account for day based values and the potential +-28h combined timezone offsets

arelle/ModelValue.py
```diff
@@ -671,241 +675,312 @@ def __new__(cls, hour: int = 0, minute: int = 0, second: int = 0, microsecond: i
...
+    return datetime.datetime(GTYPE_ANCHOR_YEAR, GTYPE_ANCHOR_MONTH, day, tzinfo=tz)
+
+
+def _gTypeHash(fields: tuple[int, ...], tz: datetime.timezone | None) -> int:
+    return hash((fields, tz.utcoffset(None) if tz is not None else None))
+
+
 class gYearMonth:
-    def __init__(self, year: int | str, month: int | str) -> None:
+    def __init__(self, year: int | str, month: int | str, tzinfo: datetime.timezone | None = None) -> None:
```
austinmatherne-wk, line 747: Doesn't need to be in this PR, but [FunctionXs.py](https://github.com/Arelle/Arelle/blob/74f6b8b3ee53c3f00a2fcf8f46f15de65162eefe/arelle/FunctionXs.py#L348) already captures timezone groups and should take advantage of this now

austinmatherne-wk (approved):

derekgengenbacher-wf (approved):

derekgengenbacher-wf:

PR #2485, opened 2026-07-17
Type Hints: arelle.UITkTable 

volodymyrtur-wk:
#### Description of change

#### Steps to Test
CI
**review**:
@Arelle/arelle

### DIFF
--- a/arelle/UITkTable.py
+++ b/arelle/UITkTable.py
@@ -22,10 +22,8 @@ from tkinter import (
     RIGHT,
     StringVar,
     TclError,
-    Frame,
-    Scrollbar,
 )
-from tkinter.ttk import Combobox as _Combobox
+from tkinter.ttk import Frame, Scrollbar, Combobox as _Combobox
 
 from arelle import TkTableWrapper
 from arelle.typing import TypeGetText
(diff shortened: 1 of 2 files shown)


arelle/UITkTable.py
```diff
@@ -1,93 +1,110 @@
...
+    S,
+    W,
+    E,
+    CENTER,
+    LEFT,
+    RIGHT,
+    StringVar,
+    TclError,
+    Frame,
+    Scrollbar,
```
austinmatherne-wk: `Frame` and `Scrollbar` were previously pulled in using the `tkinter.ttk` wildcard import (newer themed widgets). Importing them from `tkinter` will degrade the GUI appearance.

arelle/UITkTable.py
```diff
@@ -1,93 +1,110 @@
...
 
-
-    def __eq__(self, other):
+    def __eq__(self, other: object) -> bool:
+        if not isinstance(other, Coordinate):
+            return NotImplemented
         return self.x == other.x and self.y == other.y
 
+    def __ne__(self, other: object) -> bool:
+        return not self.__eq__(other)
```
austinmatherne-wk: The `isinstance`/`return NotImplemented` guard added to `__eq__` will cause `__ne__` to raise a `TypeError` here (evaluating `not NotImplemented`) on Python 3.14. Better to just delete `__ne__` and let Python auto derive it from `__eq__`

arelle/UITkTable.py
austinmatherne-wk: Looks like the `Scrollbar` and `Frame` imports were also changed in #2374 for `arelle/UiUtil.py`

austinmatherne-wk (approved):

PR #2486, opened 2026-07-17
Add xbrl_xsdtests conformance suite config

aaroncameron-wk:
#### Description of change
Add xbrl_xsdtests conformance suite config

#### Steps to Test
CI

**review**:
@Arelle/arelle

### DIFF
--- a/tests/integration_tests/validation/conformance_suite_configurations/xbrl_xsdtests.py
+++ b/tests/integration_tests/validation/conformance_suite_configurations/xbrl_xsdtests.py
@@ -26,5 +26,5 @@ config = ConformanceSuiteConfig(
     info_url='https://github.com/Arelle/xbrl-xsdtests',
     name=PurePath(__file__).stem,
     test_case_result_options='match-all',
-    shards=4,
+    shards=3,  # Fits on single macos GHA runner's 3 cores
 )
--- a/tests/integration_tests/validation/conformance_suite_configurations/xbrl_xsdtests.py
+++ b/tests/integration_tests/validation/conformance_suite_configurations/xbrl_xsdtests.py
@@ -1,0 +1,30 @@
+from pathlib import Path, PurePath
+
+from tests.integration_tests.validation.conformance_suite_config import (
+    ConformanceSuiteAssetConfig,
+    ConformanceSuiteConfig, AssetSource,
+)
+
+config = ConformanceSuiteConfig(
+    assets=[
+        ConformanceSuiteAssetConfig.conformance_suite(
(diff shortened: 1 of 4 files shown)


tests/integration_tests/validation/conformance_suite_configurations/xbrl_xsdtests.py
```diff
@@ -0,0 +1,30 @@
...
+        # The intent of the source testcase is that there are 2 invalid values,
+        # and Arelle correctly fires 2 errors, so this "failure" is accepted.
+        "anyURI/anyURI-21837395-testcase.xml:Microsoft_anyURI_b005_1355_anyURI_b005_1355.i": {
+            "xmlSchema:valueError": 1
+        },
+    }.items()},
+    info_url='https://github.com/Arelle/xbrl-xsdtests',
+    name=PurePath(__file__).stem,
+    test_case_result_options='match-all',
+    shards=4,
```
austinmatherne-wk: Out of curiosity, how much time does this add to the fast suites config in CI if it's bundled with the fast/no shards test suites? I'm wondering if it's worth including so it runs on windows. If it adds considerable time, it would still be worth a one off test run on a windows runner (or locally on a windows vm) to make sure the generated test suite runs as expected on all supported platforms.

austinmatherne-wk (dismissed):

tests/integration_tests/validation/conformance_suite_configurations/xbrl_xsdtests.py
aaroncameron-wk: It's not an incredibly fast suite, so would slow down "fast suites" quite a bit. I've added macos/windows with latest python, similar to EFM.

austinmatherne-wk: Only measuring from a couple CI runs, but the suite takes significantly longer to run on windows. Any idea why?
macOS: ~3 min
Ubuntu: ~4 min
Windows: ~12 min

aaroncameron-wk: No, but I'll investigate with some Windows runs locally and if there's anything that can be done I can write up an issue. I'm curious if there's some overhead unique to Windows that scales with testcase count. I can't recall how 9,451 testcases compares to existing conformance suites. An EFM shard looks to be < 1,000.

austinmatherne-wk (approved):

derekgengenbacher-wf (approved):

derekgengenbacher-wf:

PR #2496, opened 2026-07-21
Type Hints: arelle.ViewWinRenderedGrid 

volodymyrtur-wk:
#### Description of change

#### Steps to Test
CI
**review**:
@Arelle/arelle

### DIFF
--- a/arelle/ViewWinRenderedGrid.py
+++ b/arelle/ViewWinRenderedGrid.py
@@ -1,22 +1,26 @@
-'''
+"""
 See COPYRIGHT.md for copyright information.
-'''
-import os, threading, time, logging, sys, traceback
-from collections import OrderedDict, defaultdict
-from tkinter import Menu, BooleanVar, font as tkFont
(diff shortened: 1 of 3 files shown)


aaroncameron-wk (dismissed):

arelle/ViewWinRenderedGrid.py
```diff
@@ -913,104 +907,110 @@ def newFactOpenAspects(self, factObjectId):
...
             if dimConcept.isExplicitDimension:
                 if headersList: # has entries, add all-memembers at end
                     headersList.append("(all members)")
                 else:  # empty list, just add all members anyway
                     return self.explicitDimensionFilterMembers(structuralNode, structuralNode)
         return headersList
 
-    def onAspectComboboxSelection(self, event):
-        gridCombobox = event.widget
+    def onAspectComboboxSelection(self, event: _Combobox) -> None:
```
austinmatherne-wk: This is an event?

```suggestion
    def onAspectComboboxSelection(self, event: Event) -> None:
```

austinmatherne-wk (approved):

PR #2497, opened 2026-07-21
Pass formula params from testcase context to instance validation

aaroncameron-wk:
#### Reason for change
In some test suite scenarios (specifically the upcoming UKSEF validation suite implementation), it is desirable to pass formula parameters (e.g. `authority=UKFRC`) from the top-level test suite context into instance loading/validation.

#### Description of change
Use test suite context formula parameters as base for parameters passed into testcase variation instance context, allowing `ModelTestcaseVariation.parameters` to add/overwrite values.

#### Steps to Test
CI

**review**:
@Arelle/arelle

### DIFF
--- a/arelle/Validate.py
+++ b/arelle/Validate.py
@@ -280,7 +280,8 @@ class Validate:
             errorCaptureLevel = logging._checkLevel("WARNING")  # type: ignore[attr-defined]
         else:
             errorCaptureLevel = modelTestcaseVariation.severityLevel # default is INCONSISTENCY
-        parameters = modelTestcaseVariation.parameters.copy()
+        parameters = self.modelXbrl.modelManager.formulaOptions.typedParameters(self.modelXbrl.prefixedNamespaces)
+        parameters |= modelTestcaseVariation.parameters
         loadedModels = []
         for i, readMeFirstUri in enumerate(modelTestcaseVariation.readMeFirstUris):
             loadedModels.extend(self._testcaseLoadReadMeFirstUri(


arelle/Validate.py
```diff
@@ -280,7 +280,8 @@ def _validateTestcaseVariation(self, testcase: ModelDocument, modelTestcaseVaria
             errorCaptureLevel = logging._checkLevel("WARNING")  # type: ignore[attr-defined]
         else:
             errorCaptureLevel = modelTestcaseVariation.severityLevel # default is INCONSISTENCY
-        parameters = modelTestcaseVariation.parameters.copy()
+        parameters = self.modelXbrl.modelManager.formulaOptions.typedParameters(self.modelXbrl.prefixedNamespaces)
```
austinmatherne-wk, line 283: Passing options unrelated to XBRL formula as formula parameters is an unfortunate historical Arelle hack. Ideally we'd use the new dedicated ESEF authority arg `--esefAuthority UKFRC` for running the test suite

austinmatherne-wk (approved):

arelle/Validate.py
aaroncameron-wk: There is unfortunately not a clean way to pass `options` values to the testcase instance models. Doing so would require some hacks that might creep beyond the testcase runner infrastructure, so I thought this was a temporary, low-surface-area solution until the new test engine is adopted (at which point passing any option is trivial).

derekgengenbacher-wf (approved):

derekgengenbacher-wf:

PR #2498, opened 2026-07-23
Type Hints: arelle.ViewWinTkTable 

volodymyrtur-wk:
#### Description of change

#### Steps to Test
CI
**review**:
@Arelle/arelle

### DIFF
--- a/arelle/ViewWinTkTable.py
+++ b/arelle/ViewWinTkTable.py
@@ -4,8 +4,9 @@ See COPYRIGHT.md for copyright information.
 """
 from __future__ import annotations
 
+from collections.abc import Callable
 from tkinter import Menu
-from typing import TYPE_CHECKING, Any, Callable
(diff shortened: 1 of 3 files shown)


arelle/ViewWinTkTable.py
```diff
@@ -1,29 +1,47 @@
-'''
+"""
 @author: Acsone S. A.
 See COPYRIGHT.md for copyright information.
-'''
+"""
+from __future__ import annotations
+
+from tkinter import Menu
+from typing import TYPE_CHECKING, Any, Callable
```
austinmatherne-wk: Should import `Callable` from `collections.abc`

austinmatherne-wk (dismissed):

austinmatherne-wk (approved):

aaroncameron-wk (approved):

PR #2502, opened 2026-07-24
Implement UKSEF validation framework within ESEF plugin

aaroncameron-wk:
##### Authority-specific validation framework
- Establish framework for authority-specific validations in the ESEF plugin. 
- This framework supports running ESEF _with_ authority-specific validations by selecting an ESEF disclosure system and specifying an `esefAuthority` through various means. (`--disclosureSystem esef-2022 --esefAuthority UKFRC`)
- Additionally, users can run only the authority-specific validations by selecting the authority-specific disclosure system. (`--disclosureSystem uksef`)

##### UKSEF validation initial progress + placeholders
- Configure UKSEF as first authority-specific validation set in ESEF
- Implement UKFRC6, UKFRC7 and UKFRC8 to demonstrate pattern
- Add integration test scripts that demonstrate ESEF + authority and authority-only uses

#### Steps to Test
CI

**review**:
@Arelle/arelle

### DIFF
--- a/arelle/Validate.py
+++ b/arelle/Validate.py
@@ -795,8 +795,9 @@ class Validate:
             if not isinstance(expected, list):
                 expected = [expected]  # type: ignore[list-item]
             for testErr in _errors:
+                testErrPrefix = None
                 if isinstance(testErr, str) and testErr.startswith(("ESEF.", "NL.NL-KVK")): # compared as list of strings to QName localname
-                    testErrSuffix = testErr.rpartition(".")[2]
+                    testErrPrefix, __, testErrSuffix = testErr.rpartition(".")
                     if not testErrSuffix.isdigit():
                         testErr = testErrSuffix
                 for _exp in _expectedList:
@@ -814,8 +815,11 @@ class Validate:
                                 # UKSEF conformance suite expects XML schema validation errors in format xbrl.core.xml.SchemaValidationError.*
                                 # Example: tests/FRC/FRC_08/index.xml:TC2_invalid
                                 (_exp.localName.startswith("xbrl.core.xml.SchemaValidationError.cvc") and errPrefix == "xmlSchema") or
-                                # Actual (str): *:match, Expected (QName): {*}*.match
-                                (errLocalName and errLocalName == _exp.localName.rpartition('.')[2])
+                                # ESEF.UK*: Actual (str): *:match, Expected (QName): {*}*.match
+                                (
+                                        testErrPrefix and testErrPrefix.startswith("ESEF.UK") and
+                                        errLocalName and errLocalName == _exp.localName.rpartition('.')[2]
+                                )
                         ):
                             _expMatched = True
                     elif type(testErr) is type(_exp):
--- a/arelle/Validate.py
+++ b/arelle/Validate.py
@@ -803,12 +803,15 @@ class Validate:
                     _expMatched = False
                     if isinstance(_exp,QName) and isinstance(testErr,str):
                         errPrefix, sep, errLocalName = testErr.rpartition(":")
-                        if ((not sep and errLocalName in commaSpaceSplitPattern.split(_exp.localName.strip())) or # ESEF has comma separated list of localnames of errors
-                            (_exp == qname(XbrlConst.errMsgPrefixNS.get(errPrefix) or
+                        if (
+                                # ESEF has comma separated list of localnames of errors
+                                (not sep and errLocalName in commaSpaceSplitPattern.split(_exp.localName.strip())) or
+                                (_exp == qname(XbrlConst.errMsgPrefixNS.get(errPrefix) or
                                            (errPrefix == _exp.prefix and _exp.namespaceURI),
                                            errLocalName)) or
-                            # XDT xml schema tests expected results
-                            (_exp.namespaceURI == XbrlConst.xdtSchemaErrorNS and errPrefix == "xmlSchema")):
+                                # XDT xml schema tests expected results
+                                (_exp.namespaceURI == XbrlConst.xdtSchemaErrorNS and errPrefix == "xmlSchema")
+                        ):
                             _expMatched = True
                     elif type(testErr) is type(_exp):
                         if isinstance(testErr,dict):
--- a/arelle/Validate.py
+++ b/arelle/Validate.py
@@ -810,7 +810,12 @@ class Validate:
                                            (errPrefix == _exp.prefix and _exp.namespaceURI),
                                            errLocalName)) or
                                 # XDT xml schema tests expected results
-                                (_exp.namespaceURI == XbrlConst.xdtSchemaErrorNS and errPrefix == "xmlSchema")
+                                (_exp.namespaceURI == XbrlConst.xdtSchemaErrorNS and errPrefix == "xmlSchema") or
+                                # UKSEF conformance suite expects XML schema validation errors in format xbrl.core.xml.SchemaValidationError.*
+                                # Example: tests/FRC/FRC_08/index.xml:TC2_invalid
+                                (_exp.localName.startswith("xbrl.core.xml.SchemaValidationError.cvc") and errPrefix == "xmlSchema") or
+                                # Actual (str): *:match, Expected (QName): {*}*.match
+                                (errLocalName and errLocalName == _exp.localName.rpartition('.')[2])
                         ):
                             _expMatched = True
                     elif type(testErr) is type(_exp):
--- a/arelle/plugin/validate/ESEF/Const.py
+++ b/arelle/plugin/validate/ESEF/Const.py
@@ -14,6 +14,12 @@ styleCssHiddenPattern = re.compile(r"(.*[^\w]|^)display\s*:\s*none([^\w].*|$)")
 datetimePattern = lexicalPatterns["XBRLI_DATEUNION"]
 docTypeXhtmlPattern = re.compile(r"^<!(?:DOCTYPE\s+)\s*html(?:PUBLIC\s+)?(?:.*-//W3C//DTD\s+(X?HTML)\s)?.*>$", re.IGNORECASE)
 
+
+AUTHORITY_UKFRC = "UKFRC"
+
+TARGET_UKFRS = "UKFRS"
+
+
 FOOTNOTE_LINK_CHILDREN = frozenset((
     XbrlConst.qnLinkLoc,
     XbrlConst.qnLinkFootnoteArc,
--- a/arelle/plugin/validate/ESEF/ESEFPluginData.py
+++ b/arelle/plugin/validate/ESEF/ESEFPluginData.py
@@ -4,28 +4,75 @@ See COPYRIGHT.md for copyright information.
 from __future__ import annotations
 
 from dataclasses import dataclass
+from typing import Any
 
 from arelle.Cntlr import Cntlr
+from arelle.ModelValue import qname
+from arelle.ModelXbrl import ModelXbrl
+from arelle.typing import TypeGetText
 from arelle.utils.PluginData import PluginData
+from .Util import AUTHORITY_CODES
+
+_: TypeGetText
 
 
 @dataclass
 class ESEFPluginData(PluginData):
-    esefAuthority: str | None = None
+    _esefAuthority: str | None = None
     esefInstanceValidated: bool = False
     nonEsefInstanceExcluded: bool = False
 
     @staticmethod
-    def get(cntlr: Cntlr, name: str) -> ESEFPluginData:
+    def get(cntlr: Cntlr, name: str, esefAuthority: str | None = None) -> ESEFPluginData:
         pluginData = cntlr.getPluginData(name)
         if pluginData is None:
             pluginData = ESEFPluginData(name)
+            pluginData._esefAuthority = esefAuthority
             cntlr.setPluginData(pluginData)
         elif not isinstance(pluginData, ESEFPluginData):
             raise RuntimeError(f"PluginData already set for {pluginData.name} with unexpected type {type(pluginData)}.")
         return pluginData
 
     def reset(self) -> None:
-        self.esefAuthority = None
+        self._esefAuthority = None
         self.esefInstanceValidated = False
         self.nonEsefInstanceExcluded = False
+
+
+    def getEsefAuthority(
+            self,
+            modelXbrl: ModelXbrl,
+            parameters: dict[Any, Any] | None,
+    ) -> str | None:
+        cntlr = modelXbrl.modelManager.cntlr
+        esefAuthority = self._esefAuthority
+        if not esefAuthority and cntlr.hasGui and cntlr.config is not None:
+            esefAuthority = cntlr.config.get("esefAuthority") or None
+        formulaAuthority = None
+        if parameters:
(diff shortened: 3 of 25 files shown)


derekgengenbacher-wf (dismissed):

arelle/plugin/validate/ESEF/ESEFPluginData.py
```diff
@@ -4,28 +4,75 @@
...
     esefInstanceValidated: bool = False
     nonEsefInstanceExcluded: bool = False
 
     @staticmethod
-    def get(cntlr: Cntlr, name: str) -> ESEFPluginData:
+    def get(cntlr: Cntlr, name: str, esefAuthority: str | None = None) -> ESEFPluginData:
         pluginData = cntlr.getPluginData(name)
         if pluginData is None:
             pluginData = ESEFPluginData(name)
+            pluginData._esefAuthority = esefAuthority
```
austinmatherne-wk: This introduces a side effect on a get named method. It's also silently ignored if `pluginData` already exists. Suggest creating a separate method for `setEsefAuthority` and keep this function as purely retrieval.

arelle/Validate.py
```diff
@@ -803,12 +803,20 @@ def determineTestStatus(
...
                                            errLocalName)) or
-                            # XDT xml schema tests expected results
-                            (_exp.namespaceURI == XbrlConst.xdtSchemaErrorNS and errPrefix == "xmlSchema")):
+                                # XDT xml schema tests expected results
+                                (_exp.namespaceURI == XbrlConst.xdtSchemaErrorNS and errPrefix == "xmlSchema") or
+                                # UKSEF conformance suite expects XML schema validation errors in format xbrl.core.xml.SchemaValidationError.*
+                                # Example: tests/FRC/FRC_08/index.xml:TC2_invalid
+                                (_exp.localName.startswith("xbrl.core.xml.SchemaValidationError.cvc") and errPrefix == "xmlSchema") or
+                                # Actual (str): *:match, Expected (QName): {*}*.match
+                                (errLocalName and errLocalName == _exp.localName.rpartition('.')[2])
```
austinmatherne-wk: Could this be handled a few lines above similar to the existing `ESEF` and `NL.NL-KVK` prefix normalization?

arelle/plugin/validate/ESEF/resources/config.xml
```diff
@@ -105,4 +105,14 @@
             contextElement="scenario"
     />
 
+    <DisclosureSystem
+            names="UKSEF-2025|uksef-2025|UKSEF|uksef"
+            description="UKSEF"
```
austinmatherne-wk: Do we expect users to select this or does this primarily exist for the conformance suite? I think UKSEF users are likely to select this without realizing it skips standard esef validation. The description/name needs to make that clear. Or maybe this shouldn't exist at all and we'd be better off adding another option `--authorityOnlyRules`?

We should also update [the ESEF docs](https://github.com/Arelle/Arelle/blob/1b389bd0001ca1c996c85e28bfcaa314a9482aae/docs/source/esef.md#L4).

arelle/plugin/validate/ESEF/PluginValidationDataExtension.py
```diff
@@ -3,8 +3,33 @@
...
+        return getContextsByEntityIdentifier(modelXbrl)
+
+    def isUkfrsTarget(self, modelXbrl: ModelXbrl) -> bool:
+        if not hasattr(modelXbrl, "ixdsTarget"):
+            return False
+        return cast(str | None, modelXbrl.ixdsTarget) == TARGET_UKFRS
+
+    def isUkfrcAuthority(self, val: ValidateXbrl) -> bool:
+        esefPluginData = ESEFPluginData.get(val.modelXbrl.modelManager.cntlr, self.name)
+        authority = esefPluginData.getEsefAuthority(val.modelXbrl, val.parameters)
```
austinmatherne-wk: Can this cause the errors raised within `getEsefAuthority` to be triggered many times (`invalidESEFAuthority` once for each UKFRC `isUkfrcAuthority` rule guard)? We should probably split authority retrieval from validation.

arelle/Validate.py
aaroncameron-wk: The purpose of the existing `ESEF` normalization is to strip everything but the suffix from the Arelle-fired `ESEF.*` code for comparison against the suffix-only code expected by the conformance suite (e.g. Arelle's `ESEF.2.1.3.segmentUsed` matches ESEF suite's `segmentUsed`).

This new condition relies on that same normalization, but extends it by also stripping the suite code down to the suffix (e.g. Arelle's `ESEF.UK.segmentUsed` matches UKSEF suite's `xbrl.efr.esef.esma.segmentUsed`)

arelle/plugin/validate/ESEF/resources/config.xml
aaroncameron-wk: I went with updating the name and description because I lean towards taking advantage of the existing/familiar disclosure system functionality rather than piping an additional input through the various interfaces, but I'm still open to the `--authorityOnlyRules` approach if this still doesn't feel user friendly.

arelle/plugin/validate/ESEF/rules/UKSEF/entity.py
```diff
@@ -0,0 +1,102 @@
...
+
+@validation(
+    hook=ValidationHook.XBRL_FINALLY,
+)
+def rule_ukfrc6(
+        pluginData: PluginValidationDataExtension,
+        val: ValidateXbrl,
+        *args: Any,
+        **kwargs: Any,
+) -> Iterable[Validation]:
```
volodymyrtur-wk, line 32: Return type does not cover `None` value.

arelle/plugin/validate/ESEF/rules/UKSEF/entity.py
```diff
@@ -0,0 +1,102 @@
...
+
+@validation(
+    hook=ValidationHook.XBRL_FINALLY,
+)
+def rule_ukfrc7(
+        pluginData: PluginValidationDataExtension,
+        val: ValidateXbrl,
+        *args: Any,
+        **kwargs: Any,
+) -> Iterable[Validation]:
```
volodymyrtur-wk, line 72: Same as above

arelle/plugin/validate/ESEF/Const.py
```diff
@@ -14,6 +14,12 @@
 datetimePattern = lexicalPatterns["XBRLI_DATEUNION"]
 docTypeXhtmlPattern = re.compile(r"^<!(?:DOCTYPE\s+)\s*html(?:PUBLIC\s+)?(?:.*-//W3C//DTD\s+(X?HTML)\s)?.*>$", re.IGNORECASE)
 
+
+AUTHORITY_UKFRC = "UKFRC"
+
+TARGET_UKFRS = "UKFRS"
+
+
```
volodymyrtur-wk, line 22: Between variables, there might be one blank row or fewer

arelle/plugin/validate/ESEF/PluginValidationDataExtension.py
```diff
@@ -3,8 +3,28 @@
...
+    def getContextIssues(self, modelXbrl: ModelXbrl) -> ContextIssues:
+        return getContextIssues(modelXbrl)
+
+    def getContextsByEntityIdentifier(self, modelXbrl: ModelXbrl) -> dict[tuple[str, str], list[ModelContext]]:
+        return getContextsByEntityIdentifier(modelXbrl)
+
+    def isUkfrsTarget(self, modelXbrl: ModelXbrl) -> bool:
+        if not hasattr(modelXbrl, "ixdsTarget"):
+            return False
+        return cast(str | None, modelXbrl.ixdsTarget) == TARGET_UKFRS
```
volodymyrtur-wk, line 28: I'm not sure, but if you add  the `else` block, it'll allow you to avoid using `cast`
```python
if not hasattr(modelXbrl, "ixdsTarget"):
    return False
else:
    return modelXbrl.ixdsTarget == TARGET_UKFRS

File: tests/integration_tests/scripts/tests/python_api_validate_uksef_only.py [hunk 9]
```diff
@@ -0,0 +1,79 @@
...
+    download_private=False,
+)
+package_paths = [str(a.full_local_path) for a in package_assets]
+
+print(f"Validating report: {target_path}")
+options = RuntimeOptions(
+    entrypointFile=str(report_zip_path),
+    formulaAction='none',
+    disclosureSystemName='uksef-only',
+    internetConnectivity='offline',
```
volodymyrtur-wk: To keep consistency, it would be nice to use one type of quote (double quotes) across the project

arelle/plugin/validate/ESEF/rules/UKSEF/entity.py
aaroncameron-wk: Since this is a generator function (uses `yield`) a naked `return` statement does not return `None`, it raises `StopIteration`, effectively "returning" an empty iterable in this case. ([docs](https://docs.python.org/3/reference/expressions.html#yield-expressions))

arelle/plugin/validate/ESEF/Const.py
aaroncameron-wk: Unofficially, we use an extra blank line to better organize/separate variables in files with many variables like this, though I don't feel strongly either way.

arelle/plugin/validate/ESEF/PluginValidationDataExtension.py
aaroncameron-wk: mypy complains about this:
```
error: Returning Any from function declared to return "bool"  [no-any-return]
```

tests/integration_tests/scripts/tests/python_api_validate_uksef_only.py
aaroncameron-wk: Completely agree! I overlooked this when copy/pasting.

derekgengenbacher-wf (dismissed):

arelle/plugin/validate/ESEF/rules/UKSEF/context.py
```diff
@@ -0,0 +1,54 @@
...
+        pluginData: PluginValidationDataExtension,
+        val: ValidateXbrl,
+        *args: Any,
+        **kwargs: Any,
+) -> Iterable[Validation]:
+    """
+    UKFRC8: xbrli:segment elements MUST be used in the contexts of UKFRS target FRC-tagged data.
+    xbrli:scenario elements MUST be used in the contexts of default target ESEF-tagged data.
+    """
+    if val.authority != AUTHORITY_UKFRC:
```
austinmatherne-wk, line 35: Stumbled across a subtle bug in the GUI during testing:

1. Select the standard ESEF disclosure system
2. Uncheck `Disclosure system checks`
3. Validate

```shell
[exception:AttributeError] Instance validation exception: 'ValidateXbrl' object has no attribute 'authority', instance: _IXDS - ../../IXDS 2138002P5RNKC5W2JZ46-2024-02-24-T01.xhtml 
  File "/Users/austinmatherne/git/open-source/Arelle/arelle/Validate.py", line 152, in validate
    self.instValidator.validate(self.modelXbrl, self.modelXbrl.modelManager.formulaOptions.typedParameters(self.modelXbrl.prefixedNamespaces))
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/austinmatherne/git/open-source/Arelle/arelle/ValidateXbrl.py", line 444, in validate
    pluginXbrlMethod(self)
    ~~~~~~~~~~~~~~~~^^^^^^
  File "/Users/austinmatherne/git/open-source/Arelle/arelle/plugin/validate/ESEF/__init__.py", line 371, in validateXbrlFinally
    return validationPlugin.validateXbrlFinally(val, *args, **kwargs)
           ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/austinmatherne/git/open-source/Arelle/arelle/utils/validate/ValidationPlugin.py", line 183, in validateXbrlFinally
    self._executeModelValidations(ValidationHook.XBRL_FINALLY, val, *args, **kwargs)
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/austinmatherne/git/open-source/Arelle/arelle/utils/validate/ValidationPlugin.py", line 263, in _executeModelValidations
    for val in validations:
               ^^^^^^^^^^^
  File "/Users/austinmatherne/git/open-source/Arelle/arelle/plugin/validate/ESEF/rules/UKSEF/context.py", line 35, in rule_ukfrc8
       ^^^^^^^^^^^^^
AttributeError: 'ValidateXbrl' object has no attribute 'authority'
```

arelle/plugin/validate/ESEF/rules/UKSEF/package.py
```diff
@@ -0,0 +1,206 @@
...
+    hook=ValidationHook.XBRL_FINALLY,
+)
+def rule_ukfrc11(
+        pluginData: PluginValidationDataExtension,
+        val: ValidateXbrl,
+        *args: Any,
+        **kwargs: Any,
+) -> Iterable[Validation] | None:
+    """
+    UKFRC11: Package naming convention validation.
```
austinmatherne-wk: Some of these rule docstrings are off. From the conformance suite:
```xml
	<number>FRC_11</number>
	<name>Including multiple iXBRL documents and multiple iXBRL document sets in report packages</name>
	<description>Subdirectories MUST NOT be used in the “reports” directory.</description>
	<automatable>true</automatable>
	<reference specification="FRC UKSEF 2025 Tagging Guide" id="UKFRC11" />
```

arelle/Validate.py
austinmatherne-wk: I'm a little suspicious of this because XII has a habit of using the same error suffix between specs and only differentiating based on the namespace prefix. Could we scope this to only apply to UKSEF based on the suite error code prefix?

volodymyrtur-wk (dismissed):

arelle/plugin/validate/ESEF/rules/UKSEF/context.py
aaroncameron-wk: I'll update the UKFRC validation rules to honor `validateDisclosureSystem`. Thoughts on whether that should be applied to all validation plugins [earlier on](https://github.com/Arelle/Arelle/blob/5bd6e1a7a85cbc48eed1fcdb1eb8c838418b2824/arelle/utils/validate/ValidationPlugin.py#L244)?

austinmatherne-wk: Yes, I think checking in both `_executeModelValidations` and `_executeCntlrValidations` makes sense

derekgengenbacher-wf (dismissed):

austinmatherne-wk (approved):

PR #2506, opened 2026-07-27
ESEF plug-in: make the report file names configurable per national regulator

gmongelli:
#### Reason for change

Current ESEF guidance 2.6.3 is too restrictive for the Danish case. The Danish national regulator only prescribes the list of valid file extensions and not the file name itself.

#### Description of change

Take the regular expression to validate the file names from the authority configuration file (`authority-validations.json`).

The parameters  `reportFileNamePattern` and `reportFileNameRegex` were already defined, but not used. This pull request modifies the code to actually use those parameters to perform the `ESEF.2.6.3.incorrectNamingConventionReportPackageReportFile` validation. 

#### Steps to Test

1. Find an ESEF annual report from Denmark that does not comply with the general ESEF file naming convention given in the ESEF Filing Rules, guidance 2.6.3.
2. Validate that report, for instance by using the Arelle GUI. Expected result: the warning `ESEF.2.6.3.incorrectNamingConventionReportPackageReportFile` shows up.
3. Set the authority to "DK" or "DBA" in the Arelle GUI ('Tools' > 'Formula' > 'Parameters').
4. Validate again the report in the GUI. Expected result: to warning `ESEF.2.6.3.incorrectNamingConventionReportPackageReportFile` no longer shows up!

##### Bonus

1. Manually modify the report package so that the file extension becomes `.htm` instead of `.xhtml` or `.html`.
2. Validate the modified report again (always in the Arelle GUI) while the authority is either 'DK' or 'DBA'. Expected result: an error `ESEF.2.6.1.incorrectFileExtension` show up in this case!

**review**:
@Arelle/arelle

### DIFF
--- a/arelle/plugin/validate/ESEF/Const.py
+++ b/arelle/plugin/validate/ESEF/Const.py
@@ -117,8 +117,8 @@ filenameRegexes = {
     "ref": r"(.{1,})-[0-9]{4}-[0-9]{2}-[0-9]{2}_ref[.]xml$",
 }
 
-reportBasenamePattern = "{base}-{date}-{version}-{lang}"
-reportBasenameRegex = re.compile(r"(.{1,})-[0-9]{4}-[0-9]{2}-[0-9]{2}-[1-9]-[a-zA-Z]{1,8}(-[a-zA-Z0-9]{1,8})*$")
+reportBasenamePattern = "{base}-{date}-{version}-{lang}.[x]htm[l]"
+reportBasenameRegex = re.compile(r"(.{1,})-[0-9]{4}-[0-9]{2}-[0-9]{2}-[1-9]-[a-zA-Z]{1,8}(-[a-zA-Z0-9]{1,8})*[.][xX]?[hH][tT][mM][lL]?$")
 
 mandatory: set[QName] = set()  # mandatory element qnames
 
--- a/arelle/plugin/validate/ESEF/Const.py
+++ b/arelle/plugin/validate/ESEF/Const.py
@@ -117,8 +117,8 @@ filenameRegexes = {
     "ref": r"(.{1,})-[0-9]{4}-[0-9]{2}-[0-9]{2}_ref[.]xml$",
 }
 
-reportDefaultFileNamePattern = "{base}-{date}-{version}-{lang}.[x]htm[l]"
-reportDefaultFileNameRegex = re.compile(r"(.{1,})-[0-9]{4}-[0-9]{2}-[0-9]{2}-[1-9]-[a-zA-Z]{1,8}(-[a-zA-Z0-9]{1,8})*[.]x?html?$")
+reportDefaultFileNamePattern = "{base}-{date}-{version}-{lang}.[x]html"
+reportDefaultFileNameRegex = re.compile(r"(.{1,})-[0-9]{4}-[0-9]{2}-[0-9]{2}-[1-9]-[a-zA-Z]{1,8}(-[a-zA-Z0-9]{1,8})*[.]x?html$")
 
 mandatory: set[QName] = set()  # mandatory element qnames
 
--- a/arelle/plugin/validate/ESEF/Const.py
+++ b/arelle/plugin/validate/ESEF/Const.py
@@ -117,8 +117,8 @@ filenameRegexes = {
     "ref": r"(.{1,})-[0-9]{4}-[0-9]{2}-[0-9]{2}_ref[.]xml$",
 }
 
-reportBasenamePattern = "{base}-{date}-{version}-{lang}.[x]htm[l]"
-reportBasenameRegex = re.compile(r"(.{1,})-[0-9]{4}-[0-9]{2}-[0-9]{2}-[1-9]-[a-zA-Z]{1,8}(-[a-zA-Z0-9]{1,8})*[.][xX]?[hH][tT][mM][lL]?$")
+reportDefaultFileNamePattern = "{base}-{date}-{version}-{lang}.[x]htm[l]"
+reportDefaultFileNameRegex = re.compile(r"(.{1,})-[0-9]{4}-[0-9]{2}-[0-9]{2}-[1-9]-[a-zA-Z]{1,8}(-[a-zA-Z0-9]{1,8})*[.]x?html?$")
 
 mandatory: set[QName] = set()  # mandatory element qnames
 
--- a/arelle/plugin/validate/ESEF/ESEF_Current/ValidateXbrlFinally.py
+++ b/arelle/plugin/validate/ESEF/ESEF_Current/ValidateXbrlFinally.py
@@ -98,6 +98,13 @@ def validateXbrlFinally(val: ValidateXbrl, *args: Any, **kwargs: Any) -> None:
     prefixedNamespaces = modelXbrl.prefixedNamespaces
 
     reportPackageMaxMB = val.authParam["reportPackageMaxMB"]
+    reportFileNamePattern = val.authParam["reportFileNamePattern"]
+    reportFileNameRegex = val.authParam["reportFileNameRegex"]
+    if reportFileNameRegex:
+        reportFileNameRegex = re.compile(reportFileNameRegex)
+    if not reportFileNamePattern:
+        reportFileNamePattern = reportBasenamePattern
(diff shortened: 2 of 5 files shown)


arelle/plugin/validate/ESEF/Const.py
```diff
@@ -117,8 +117,8 @@
     "ref": r"(.{1,})-[0-9]{4}-[0-9]{2}-[0-9]{2}_ref[.]xml$",
 }
 
-reportBasenamePattern = "{base}-{date}-{version}-{lang}"
-reportBasenameRegex = re.compile(r"(.{1,})-[0-9]{4}-[0-9]{2}-[0-9]{2}-[1-9]-[a-zA-Z]{1,8}(-[a-zA-Z0-9]{1,8})*$")
+reportBasenamePattern = "{base}-{date}-{version}-{lang}.[x]htm[l]"
+reportBasenameRegex = re.compile(r"(.{1,})-[0-9]{4}-[0-9]{2}-[0-9]{2}-[1-9]-[a-zA-Z]{1,8}(-[a-zA-Z0-9]{1,8})*[.][xX]?[hH][tT][mM][lL]?$")
```
brettkail-wk: I see `.htm` was added in the most recent ESEF version.  Do we need to make the validation sensitive to version?

Why should the extension be case insensitive?

arelle/plugin/validate/ESEF/ESEF_Current/ValidateXbrlFinally.py
```diff
@@ -98,6 +98,13 @@ def validateXbrlFinally(val: ValidateXbrl, *args: Any, **kwargs: Any) -> None:
     prefixedNamespaces = modelXbrl.prefixedNamespaces
 
     reportPackageMaxMB = val.authParam["reportPackageMaxMB"]
+    reportFileNamePattern = val.authParam["reportFileNamePattern"]
+    reportFileNameRegex = val.authParam["reportFileNameRegex"]
+    if reportFileNameRegex:
+        reportFileNameRegex = re.compile(reportFileNameRegex)
+    if not reportFileNamePattern:
+        reportFileNamePattern = reportBasenamePattern
+        reportFileNameRegex = reportBasenameRegex
```
brettkail-wk: The pre-existing globals `reportBasenameX` are somewhat confusingly named (given a path, I normally think of `/directory/basename`, but here I think it was `basename.extension`), but it seems even more confusing to mix conventions.  Should we rename the globals?

arelle/plugin/validate/ESEF/ESEF_Current/ValidateXbrlFinally.py
```diff
@@ -248,16 +255,25 @@ def checkFootnote(elt: ModelInlineFootnote | ModelResource, text: str) -> None:
...
                 _baseName, _baseExt = os.path.splitext(doc.basename)
-                if _baseExt not in (".xhtml",".html"):
-                    if val.consolidated:
-                        errorCode = "ESEF.2.6.1.incorrectFileExtension"
-                        reportType = _("Inline XBRL document included within a ESEF report package")
+                possibleExtensions: [str]
+                if val.consolidated:
+                    errorCode = "ESEF.2.6.1.incorrectFileExtension"
+                    reportType = _("Inline XBRL document included within a ESEF report package")
+                    if val.authority and val.authority in {"DK", "DBA"}:
```
brettkail-wk: This hardcoding seems unusual.  If we're going to hardcode here, should we just hardcode `reportFileNameRegex`/`reportFileNamePattern` rather than adding to `authority-validations.json`?  Or vice versa: should we add `consolidatedReportFileNameRegex`, `consolidatedReportFileNamePattern`, `consolidatedReportFileNameExtensions` and corresponding "unconsolidated"?

arelle/plugin/validate/ESEF/Const.py
gmongelli: Well, the Windows OS is case-insensitive.

That is why I chose a case-insensitive regular expression.

However, to be coherent with the case-sensitive checks for `ESEF.2.6.1.incohrentFileExtension`, I'll only keep the lowercase variant.

arelle/plugin/validate/ESEF/ESEF_Current/ValidateXbrlFinally.py
gmongelli: Yes, the naming is not coherent.

Taking https://en.wikipedia.org/wiki/Filename as a reference, a filename is composed of a 'base name' plus an extension.

Therefore, I would recommend renaming:
* `reportBasenamePattern` to `reportDefaultFileNamePattern`
* `reportBasenameRegex` to `reportDefaultFileNameRegex`

arelle/plugin/validate/ESEF/ESEF_Current/ValidateXbrlFinally.py
gmongelli: I would not hardcode `reportFileNameRegex`/ `reportFileNamePattern` as it is a well-defined configuration option in `authority-validations.json`.

OK for defining still more options, but this time for the extension (`[".xhtml",".html"]` vs. `[".xhtml",".html",".htm"]`).

The improvement will follow soon!

arelle/plugin/validate/ESEF/ESEF_Current/ValidateXbrlFinally.py
gmongelli: By executing the ESEF conformance tests, I noticed that there is an inline XBRL test case where the `.htm` extension is explicitly supposed to generate an `incorrectFileExtension` error. 

Because of that, I'll rollback the whole change as it becomes pointless.

arelle/plugin/validate/ESEF/ESEF_Current/ValidateXbrlFinally.py
gmongelli: > I would not hardcode `reportFileNameRegex`/ `reportFileNamePattern` as it is a well-defined configuration option in `authority-validations.json`.
> 
> OK for defining still more options, but this time for the extension (`[".xhtml",".html"]` vs. `[".xhtml",".html",".htm"]`).
> 
> The improvement will follow soon!

Sorry, the comment is misplaced.

I'll put it it in the right context.

This change is still meaningfull!

arelle/plugin/validate/ESEF/ESEF_Current/ValidateXbrlFinally.py
gmongelli: By executing the ESEF conformance tests, I noticed that there is an inline XBRL test case where the `.htm` extension is explicitly supposed to generate an `incorrectFileExtension` error.

Because of that, I have rolled back the whole change as it became pointless.

arelle/plugin/validate/ESEF/resources/authority-validations.json
```diff
@@ -250,6 +252,8 @@
         "name": "Denmark",
         "reportPackageMaxMB": "2000",
         "reportPackageMeasurement": "unzipped",
+        "reportFileNamePattern": "{base}-*.[x]html",
+        "reportFileNameRegex": "([^-]{1,}).*[.]x?html$",
```
austinmatherne-wk, line 256: Setting `reportFileNamePattern`/`reportFileNameRegex` for DK/DBA here activates a separate, currently dormant check: ESEF.3.1.5 on inline report documents. That rule is meant only for extension taxonomy files, not instance documents, but there's a pre-existing (incorrect) branch that also applies it to the inline XBRL report document if these are set:

https://github.com/Arelle/Arelle/blob/83b5401ba7e2cd78d78832bf3e1566bfc496584c/arelle/plugin/validate/ESEF/ESEF_Current/DTS.py#L59-L62

identical in ESEF_2021:
https://github.com/Arelle/Arelle/blob/83b5401ba7e2cd78d78832bf3e1566bfc496584c/arelle/plugin/validate/ESEF/ESEF_2021/DTS.py#L58-L61

Can you remove those two code blocks above from the DTS.py modules so 3.1.5 doesn't start firing for inline reports?

gmongelli: Done!
You were right!

austinmatherne-wk (approved):

PR #2507, opened 2026-07-28
Type Hints: arelle.ModelVersReport 

volodymyrtur-wk:
#### Description of change

#### Steps to Test
CI
**review**:
@Arelle/arelle

### DIFF
--- a/arelle/ModelVersReport.py
+++ b/arelle/ModelVersReport.py
@@ -1,17 +1,54 @@
-'''
+"""
 See COPYRIGHT.md for copyright information.
-'''
-import os, datetime
-import regex as re
-from collections import defaultdict
-from arelle import (XbrlConst, XbrlUtil, XmlUtil, UrlUtil, ModelXbrl, ModelDocument, ModelVersObject,
-                    Version, XmlValidate)
-from arelle.ModelDtsObject import ModelConcept
-from arelle.ModelObject import ModelObject
-from arelle.ModelValue import qname, QName
-from arelle.FileSource import FileNamedStringIO
+"""
+from __future__ import annotations
 
-def create(modelXbrlFromDTS, modelXbrlToDTS):
+import datetime
+import os
+from collections import defaultdict
+from collections.abc import Callable, Iterable
+from typing import Any
+
+import regex as re
+from lxml import etree
+
+from arelle import (
+    ModelDocument,
+    UrlUtil,
+    Version,
+    XbrlConst,
+    XbrlUtil,
+    XmlUtil,
+    XmlValidate,
+)
+from arelle.FileSource import FileNamedStringIO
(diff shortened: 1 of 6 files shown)


arelle/ModelVersReport.py
```diff
@@ -154,48 +181,48 @@ def versioningReportDiscover(self, rootElement):
...
-                                    DTSmodelXbrl = ModelXbrl.create(self.modelXbrl.modelManager,
+                                    DTSmodelXbrl = ModelXbrlCreate(self.modelXbrl.modelManager,
                                                  newDocumentType=ModelDocument.Type.DTSENTRIES,
                                                  url=self.uri[:-4] + "-" + ln + ".dts", isEntry=True)
                                     DTSdoc = DTSmodelXbrl.modelDocument
-                                    DTSdoc.inDTS = True
+                                    DTSdoc.inDTS = True  # type: ignore[union-attr]
                                     for schemaRefElt in schemaRefElts:
-                                        if schemaRefElt.get("{http://www.w3.org/1999/xlink}href") is not None:
+                                        if href := schemaRefElt.get("{http://www.w3.org/1999/xlink}href") is not None:
```
austinmatherne-wk: The walrus operator is binding `href` to the true/false result of (`val is not None` instead of `val`).
```suggestion
                                        if (href := schemaRefElt.get("{http://www.w3.org/1999/xlink}href")) is not None:
```

arelle/ModelVersReport.py
```diff
@@ -987,107 +1089,146 @@ def DRSdimsDiff(self, fromDTS, fromPriItemDRSrels, toDTS, toPriItemDRSrels):
...
             return qname(self.namespaceRenameToURI[toDTSqname.namespaceURI],
                          toDTSqname.localName)
         return toDTSqname
 
-    def createAction(self):
-        action = XmlUtil.addChild(self.reportElement, XbrlConst.ver, "ver:action", (("id","action{0:05}".format(self.actionNum) ),))
+    def createAction(self) -> ModelObject:
+        action = XmlUtil.addChild(self.reportElement, XbrlConst.ver, "ver:action", (("id", "action{0:05}".format(self.actionNum)),))  # type: ignore[arg-type]
         self.actionNum += 1
-        assignmentRef = XmlUtil.addChild(action, XbrlConst.ver, "ver:assignmentRef", (("ref","versioningTask"),) )
```
austinmatherne-wk, line 1011: `assignmentRef` was unused, but the function call was adding a node to `action`. This should be added back:

```python
XmlUtil.addChild(action, XbrlConst.ver, "ver:assignmentRef", (("ref","versioningTask"),) )
```

volodymyrtur-wk: I wanted to erase the variable name and accidentally deleted the whole line🙈

austinmatherne-wk (approved):

PR #2518, opened 2026-07-30
Type Hints: arelle.TkTableWrapper

volodymyrtur-wk:
#### Description of change

#### Steps to Test
CI
**review**:
@Arelle/arelle

### DIFF
--- a/arelle/TkTableWrapper.py
+++ b/arelle/TkTableWrapper.py
@@ -26,42 +26,49 @@
 This contains a wrapper class for the tktable widget as well a class for using
 tcl arrays that are, in some instances, required by tktable.
 """
+from __future__ import annotations
 
 __author__ = "Guilherme Polo <ggpolo@gmail.com>"
 
 __all__ = ["ArrayVar", "Table"]
 
 import os
-import collections.abc
 import platform
 import sys
-from pathlib import Path
-
 import tkinter
+from tkinter import Variable, Widget, messagebox, Event, Menu, Tk
+from pathlib import Path
+from typing import TYPE_CHECKING, Any
+
 
 from arelle.SystemInfo import PlatformOS
+from arelle.typing import TypeGetText
+
+if TYPE_CHECKING:
+    from collections.abc import Iterable
+    from tkinter import Misc
+
+_: TypeGetText
 
 
-
-def _setup_master(master):
+def _setup_master(master: Misc | None) -> Misc:
     if master is None:
-        if tkinter._support_default_root:
-            master = tkinter._default_root or tkinter.Tk()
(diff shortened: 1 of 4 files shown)


arelle/TkTableWrapper.py
```diff
@@ -514,45 +507,45 @@ def window_configure(self, index, option=None, **kwargs):
...
             for key, val in kwargs.items():
-                args += ('-%s' % key, val)
+                args += ("-%s" % key, val)
 
-            self.tk.call(self._w, 'window', 'configure', index, *args)
+            self.tk.call(self._w, "window", "configure", index, *args)
 
-    def window_delete(self, *indexes):
-        self.tk.call(self._w, 'window', 'delete', *indexes)
+    def window_delete(self, *indexes: Iterable[str | int]) -> None:
```
austinmatherne-wk: For a `*args` parameter the annotation types each individual argument, not the collected tuple:
```suggestion
    def window_delete(self, *indexes: str | int) -> None:
```

arelle/TkTableWrapper.py
```diff
@@ -514,45 +507,45 @@ def window_configure(self, index, option=None, **kwargs):
...
+    def xview_moveto(self, fraction: float) -> None:
         """Adjusts the view in the window so that fraction of the total width
         of the table text is off-screen to the left. The fraction parameter
         must be a fraction between 0 and 1."""
-        self.tk.call(self._w, 'xview', 'moveto', fraction)
+        self.tk.call(self._w, "xview", "moveto", fraction)
 
 
-    def xview_scroll(self, *L):
+    def xview_scroll(self, *L: Iterable[str | int]) -> None:
```
austinmatherne-wk:
```suggestion
    def xview_scroll(self, *L: str | int) -> None:
```

arelle/TkTableWrapper.py
```diff
@@ -580,17 +572,18 @@ def yview(self, index=None):
...
-    def yview_moveto(self, fraction):
+    def yview_moveto(self, fraction: float) -> None:
         """Adjusts the view in the window so that the element given by
         fraction appears at the top of the window. The fraction parameter
         must be a fraction between 0 and 1."""
-        self.tk.call(self._w, 'yview', 'moveto', fraction)
+        self.tk.call(self._w, "yview", "moveto", fraction)
 
-    def yview_scroll(self, *L):
+    def yview_scroll(self, *L: Iterable[str | int]) -> None:
```
austinmatherne-wk:
```suggestion
    def yview_scroll(self, *L: str | int) -> None:
```

austinmatherne-wk (approved):

PR #2519, opened 2026-07-30
Edgar ignore node modules

freddym56:
#### Reason for change
Add python monkey dependency for EDGAR plugin. Ignore node_modules directory when discovering entrypoints.

#### Description of change
Upcoming EDGAR release will have a dependency on pythonmonkey. Ignoring node_modules directory when finding entrypoints.

#### Steps to Test

**review**:
@Arelle/arelle

### DIFF
--- a/requirements-plugins.txt
+++ b/requirements-plugins.txt
@@ -5,6 +5,7 @@ ixbrl-viewer==1.5.0
 holidays==0.101
 matplotlib==3.10.9
 pytz==2026.3.post1
+pythonmonkey==1.3.2
 
 # EDGAR and XULE plugins
 aniso8601==10.0.1
--- a/LICENSE.md
+++ b/LICENSE.md
@@ -56,6 +56,9 @@ Pillow: Copyright © 2010-2022 by Alex Clark and contributors
 PyParsing: Copyright © 2003-2022 Paul McGuire.
 [License](https://github.com/pyparsing/pyparsing/blob/master/LICENSE)
 
+PythonMonkey: Copyright (c) 2023-2024 Distributive Corp.
+[License](https://github.com/Distributive-Network/PythonMonkey/blob/main/LICENSE)
+
 rdflib: Copyright (c) 2002-2021, RDFLib Team. All rights reserved.
 [License](https://opensource.org/licenses/BSD-3-Clause)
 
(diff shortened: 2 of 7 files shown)


requirements-plugins.txt
```diff
@@ -5,6 +5,7 @@ ixbrl-viewer==1.5.0
 holidays==0.101
 matplotlib==3.10.9
 pytz==2026.3.post1
+pythonmonkey==1.3.2
```
austinmatherne-wk, line 8: Hi @freddym56. Can you provide more details on why and how this will be used? Specifically, is EDGAR going to evaluate JS inline, or `require()` a module, and if so, where will it be loaded from? I'd like to understand the packaging requirements so I can make sure this works across all the platforms and environments we support.

At the very least it will require configuring in `distro.py` for the frozen builds, a quick local test showed `import pythonmonkey` segfaults in a cx_Freeze build unless `asyncio`, `pminit` and `aiohttp` are bundled explicitly, along with pythonmonkey's own `node_modules`. `pminit` also shells out to `npm i` at install time, so the Docker build will need Node added. Happy to handle this.

Also, just flagging this is a chunky dependency. My rough estimate from a quick build test is a 15-20% increase in size over the current platform build artifacts.

freddym56: Hi @austinmatherne-wk, the EDGAR/render plugin will be making a call to pythonmonkey.eval. The JS code that we are executing will live under EDGAR/view/rfileviewer/dist. I did also notice our build size increased by about 25-30%.

freddym56: @austinmatherne-wk we've explored quickjs which is a more lightweight library. However, this library is not actively maintained which would then introduce problems for newer python versions. There is a fork (quickjs-ng) which claims to be active but i haven't seen updates in months. It also currently has a bug that is breaking the functionality we are trying to implement. Is it feasible for you to do your build with Pythonmonkey?

austinmatherne-wk: Hi @freddym56. Thanks for investigating alternatives. I appreciate the effort. I'll create a test branch with builds to test packaging in our various distributions and let you know within a day or two if it's working. But do note that my tests will be limited to a simple js eval from a simple test plugin. It's possible that the EDGAR usage of pythonmonkey depends on something my own testing won't cover. I'll report back with my findings soon.

austinmatherne-wk (approved):

derekgengenbacher-wf (approved):

derekgengenbacher-wf:

PR #2522, opened 2026-08-03
Type Hints: arelle.ViewFileRenderedGrid

volodymyrtur-wk:
#### Description of change

#### Steps to Test
CI
**review**:
@Arelle/arelle

### DIFF
--- a/arelle/ViewFileRenderedGrid.py
+++ b/arelle/ViewFileRenderedGrid.py
@@ -325,7 +325,7 @@ class ViewRenderedGrid(ViewFile.View):
                     lytMdlYGroups = getattr(lytMdlTable.lytMdlAxisHeaders("y"), "lytMdlGroups", [])
                     # build y row headers
                     numYrows = lytMdlTable.numBodyCells("y")
-                    yRowHdrs: list[list[dict[str, str | None]]] = [[] for _ in range(numYrows)]  # list of lists of row header elements for each row
+                    yRowHdrs: list[list[dict[str, str | int | None]]] = [[] for _ in range(numYrows)]  # list of lists of row header elements for each row
                     for lytMdlYGrp in lytMdlYGroups:
                         for lytMdlYHdr in lytMdlYGrp.lytMdlHeaders:
                             yRow = 0
--- a/arelle/ViewFileRenderedGrid.py
+++ b/arelle/ViewFileRenderedGrid.py
@@ -128,29 +128,30 @@ class ViewRenderedGrid(ViewFile.View):
 
     def viewHTML(self, lytMdlTblMdl: LytMdlTableModel) -> None:
         for lytMdlTableSet in lytMdlTblMdl.lytMdlTableSets:
-            self.tblElt.append(etree.Comment(f"TableSet linkbase file: {lytMdlTableSet.srcFile}, line {lytMdlTableSet.srcLine}"))  # type: ignore[union-attr]
-            self.tblElt.append(etree.Comment(f"TableSet linkrole: {lytMdlTableSet.srcLinkrole}"))  # type: ignore[union-attr]
-            tblSetHdr = etree.SubElement(self.tblElt, "{http://www.w3.org/1999/xhtml}tr")  # type: ignore[type-var]
-            etree.SubElement(tblSetHdr, "{http://www.w3.org/1999/xhtml}td").text = lytMdlTableSet.label  # type: ignore[type-var, union-attr]
+            assert self.tblElt is not None
+            self.tblElt.append(etree.Comment(f"TableSet linkbase file: {lytMdlTableSet.srcFile}, line {lytMdlTableSet.srcLine}"))
+            self.tblElt.append(etree.Comment(f"TableSet linkrole: {lytMdlTableSet.srcLinkrole}"))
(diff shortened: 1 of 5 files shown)


arelle/ViewFileRenderedGrid.py
```diff
@@ -186,72 +218,73 @@ def viewHTML(self, lytMdlTblMdl):
...
+                            if lytMdlCell.isOpenAspectEntrySurrogate:  # type: ignore[union-attr]
+                                continue
+                            justify = "left"
+                            for f, v, justify in lytMdlCell.facts:  # type: ignore[union-attr]
+                                break
+                            etree.SubElement(rowElt, "{http://www.w3.org/1999/xhtml}td",  # type: ignore[type-var]
+                                                      attrib={"class": "cell",
+                                                              "style": f"text-align:{justify};width:8em;"}
+                                             ).text = "\n".join(v for f, v, justify in lytMdlCell.facts)  # type: ignore[union-attr]
                             yRowNum += 1
```
austinmatherne-wk: Indention on this line didn't move with the removal of `if True:`
```suggestion
                        yRowNum += 1
```

arelle/ViewFileRenderedGrid.py
```diff
@@ -263,62 +296,61 @@ def viewXLSX(self, lytMdlTblMdl):
...
                             if all(lytMdlCell.isOpenAspectEntrySurrogate for lytMdlCell in lytMdlYHdr.lytMdlCells):
-                                continue # skip header with only open aspect entry surrogate
+                                continue  # skip header with only open aspect entry surrogate
                             for lytMdlYCell in lytMdlYHdr.lytMdlCells:
                                 for iLabel in range(lytMdlYHdr.maxNumLabels):
                                     if lytMdlYCell.isOpenAspectEntrySurrogate:
-                                        continue # strip all open aspect entry surrogates from layout model file
-                                    rowHdrElt = {"align": "left"}
+                                        continue  # strip all open aspect entry surrogates from layout model file
+                                    rowHdrElt: dict[str, str | None] = {"align": "left"}
```
austinmatherne-wk: Below an int value (`lytMdlYCell.span`) is assigned. `yRowHdrs` type should be updated, as well: 
```suggestion
                                    rowHdrElt: dict[str, str | int | None] = {"align": "left"}
```

arelle/ViewFileRenderedGrid.py
```diff
@@ -84,81 +106,91 @@ def __init__(self, modelXbrl, outfile, lang, cssExtras):
...
             self.viewXLSX(lytMdlTblMdl)
 
-    def viewHTML(self, lytMdlTblMdl):
+    def viewHTML(self, lytMdlTblMdl: LytMdlTableModel) -> None:
         for lytMdlTableSet in lytMdlTblMdl.lytMdlTableSets:
-            self.tblElt.append(etree.Comment(f"TableSet linkbase file: {lytMdlTableSet.srcFile}, line {lytMdlTableSet.srcLine}"))
-            self.tblElt.append(etree.Comment(f"TableSet linkrole: {lytMdlTableSet.srcLinkrole}"))
-            tblSetHdr = etree.SubElement(self.tblElt, "{http://www.w3.org/1999/xhtml}tr")
-            etree.SubElement(tblSetHdr, "{http://www.w3.org/1999/xhtml}td").text = lytMdlTableSet.label
+            self.tblElt.append(etree.Comment(f"TableSet linkbase file: {lytMdlTableSet.srcFile}, line {lytMdlTableSet.srcLine}"))  # type: ignore[union-attr]
```
austinmatherne-wk: `assert self.tblElt is not None` would remove a bunch of type ignore comments?

volodymyrtur-wk: Yes, it will. Usually, I try to avoid such constructions because, in some cases, the error must appear, and we'll catch it in higher layers

arelle/ViewFileRenderedGrid.py
```diff
@@ -263,62 +297,61 @@ def viewXLSX(self, lytMdlTblMdl):
...
                     continue
                 lytMdlZBodyCell = lytMdlTable.lytMdlBodyChildren[0] # examples only show one z cell despite number of tables
-                for lytMdlYBodyCell in lytMdlZBodyCell.lytMdlBodyChildren:
+                for lytMdlYBodyCell in lytMdlZBodyCell.lytMdlBodyChildren:  # type: ignore[union-attr]
                     lytMdlXGroups = getattr(lytMdlTable.lytMdlAxisHeaders("x"), "lytMdlGroups", [])
                     lytMdlYGroups = getattr(lytMdlTable.lytMdlAxisHeaders("y"), "lytMdlGroups", [])
                     # build y row headers
                     numYrows = lytMdlTable.numBodyCells("y")
-                    yRowHdrs = [[] for i in range(numYrows)] # list of list of row header elements for each row
+                    yRowHdrs: list[list[dict[str, str | None]]] = [[] for _ in range(numYrows)]  # list of lists of row header elements for each row
```
austinmatherne-wk: Looks like this still needs to be updated to accommodate the correct `rowHdrElt` type:
```suggestion
                    yRowHdrs: list[list[dict[str, str | int | None]]] = [[] for _ in range(numYrows)]  # list of lists of row header elements for each row
```

derekgengenbacher-wf (approved):

austinmatherne-wk (approved):

PR #2525, opened 2026-08-04
Migrate from flake8 to ruff

aaroncameron-wk:
#### Reason for change

`flake8` has fallen behind on active maintenance velocity and requires multiple plugins (`flake8-noqa`, etc.) plus a separate `setup.cfg` (since it doesn't fully support `pyproject.toml`) to get the behavior we want. `ruff` is a drop-in replacement that's significantly faster and configurable entirely from `pyproject.toml`.

#### Description of change

- Replaced `flake8`/`flake8-noqa` with `ruff` in `requirements-dev.txt`.
- Updated the `lint` tox environment to run `ruff check arelle tests` instead of `flake8 arelle tests`.
- Added `[tool.ruff]`/`[tool.ruff.lint]` to `pyproject.toml`, porting over the same `exclude` paths and `ignore` codes that were previously configured in `setup.cfg`'s `[flake8]` section (`select = ["E", "W", "F"]` mirrors flake8's default pycodestyle + Pyflakes checks).
- Removed `setup.cfg`, which only existed to hold the `[flake8]` config.

This migration is configuration-only and does not change any linting behavior or touch any source files: `ruff check arelle tests` reports zero violations, matching the current clean `flake8 arelle tests` baseline.

Note: flake8's ignored codes `E121, E123, E124, E125, E126, E127, E128, E129, E131, E704, W503, W504` were omitted from the ruff `ignore` list because ruff doesn't implement these checks at all (they're excluded from pycodestyle's own default select list too), so this has no effect on enforced behavior.

#### Steps to Test

- CI

**review**:
@Arelle/arelle


pyproject.toml
```diff
@@ -116,6 +116,164 @@ markers = [
     "slow: marks tests as slow (deselect with '-m \"not slow\"')",
 ]
 
+[tool.ruff]
```
austinmatherne-wk, line 119: arbitrary line:

I think docker/git ignores need to be updated to exclude `.ruff_cache/`

pyproject.toml
```diff
@@ -116,6 +116,164 @@ markers = [
...
+target-version = "py310"
```
austinmatherne-wk: This [should be removed](https://docs.astral.sh/ruff/settings/#target-version)

austinmatherne-wk (approved):

derekgengenbacher-wf (approved):

PR #2535, opened 2026-08-07
Enforce double quote usage with ruff

aaroncameron-wk:
#### Description of change
- Update ruff configuration to enforce double quote usage
- Update double quote usage across codebase with `ruff check --fix`
- Add update commit hash to `.git-blame-ignore-revs`

#### Steps to Test
CI

**review**:
@Arelle/arelle


pyproject.toml
```diff
@@ -146,9 +146,10 @@ extend-exclude = [
 
 [tool.ruff.lint]
 # Mirrors the previous flake8 configuration (pycodestyle E/W, Pyflakes F) so
-# that migrating off flake8 does not change what gets flagged.
+# that migrating off flake8 does not change what gets flagged. "Q"
+# (flake8-quotes) is additionally enabled to enforce double quotes.
 preview = true
-select = ["E", "W", "F"]
+select = ["E", "W", "F", "Q"]
```
austinmatherne-wk, line 152: Thanks for doing this. If we're going to fail ci on quotes we should add documentation explaining how contributors can run ruff (this was a preexisting issue for both flake8 and mypy, but a larger problem now if we introduce quote standardization).

austinmatherne-wk (dismissed):

austinmatherne-wk (approved):

PR #2539, opened 2026-08-11
Type Hints: arelle.CntlrQuickBooks 

volodymyrtur-wk:
#### Reason for change

#### Steps to Test
CI
**review**:
@Arelle/arelle

### DIFF
--- a/arelle/CntlrQuickBooks.py
+++ b/arelle/CntlrQuickBooks.py
@@ -5,23 +5,36 @@ See COPYRIGHT.md for copyright information.
 """
 from __future__ import annotations
 
+import datetime
+import io
+import uuid
+from typing import TYPE_CHECKING, BinaryIO
+from urllib.parse import SplitResult
+
 from lxml import etree
-import uuid, io, datetime
+
 from arelle import XmlUtil
 
-clientVersion = None
-userName = None
-sessions = {}  # use when interactive session started by Quickbooks side (not used now)
-qbRequests = []  # used by rest API or GUI requests for QB data
-qbRequestStatus = {}
-xbrlInstances = {}
-cntlr = None
+if TYPE_CHECKING:
+    from arelle.Cntlr import Cntlr
+    from arelle.typing import TypeGetText
+
+    _: TypeGetText
+
+clientVersion: str | None = None
+userName: str | None = None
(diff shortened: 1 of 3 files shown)


derekgengenbacher-wf (dismissed):

arelle/CntlrQuickBooks.py
```diff
@@ -37,66 +50,65 @@
...
-            print ("request {0}".format(requestName))
-            response = None
+            requestName = request.tag.partition("}")[2]  # type: ignore[arg-type,union-attr]
+            response: str | list[str] | None = None
             if request.tag == "{http://developer.intuit.com/}serverVersion":
                 response = "Arelle 1.0"
             elif request.tag == "{http://developer.intuit.com/}clientVersion":
                 global clientVersion
-                clientVersion = request.find("{http://developer.intuit.com/}strVersion").text
+                strVersionElt = request.find("{http://developer.intuit.com/}strVersion").text  # type: ignore[union-attr]
```
austinmatherne-wk: Both this new var `strVersionElt` and the original global var `clientVersion` are never used after assignment. Remove the dead code please.

arelle/CntlrQuickBooks.py
```diff
@@ -37,66 +50,65 @@
...
+    if cntlr is None:
+        cntlr = _cntlr
     soapDocument = etree.parse(soapFile)
     soapBody = soapDocument.find("{http://schemas.xmlsoap.org/soap/envelope/}Body")
     if soapBody is None:
         return ""
     else:
         for request in soapBody.iterchildren():
-            requestName = request.tag.partition("}")[2]
-            print ("request {0}".format(requestName))
```
austinmatherne-wk, line 89: Was this print diagnostic intentionally removed?

volodymyrtur-wk: the mypy had a complaint on that line, and instead of adding a comment, I decided to remove it. I can return it if needed.

austinmatherne-wk: I'm not familiar with this module (I suspect it doesn't work), but I'd default to retaining diagnostics unless we actually run the code and know it isn't useful

austinmatherne-wk (approved):

derekgengenbacher-wf (approved):

derekgengenbacher-wf:

PR #2550, opened 2026-08-16
Fixed current_year/prior_year detection issue.

allenbross-wf:
#### Description of change

Updated EDINET issue around detection of current year and prior year DEI facts.

#### Steps to Test

CI

**review**:
@Arelle/arelle

### DIFF
--- a/arelle/plugin/validate/EDINET/ContextRequirement.py
+++ b/arelle/plugin/validate/EDINET/ContextRequirement.py
@@ -10,8 +10,11 @@ from typing import Literal
 @dataclass(frozen=True)
 class ContextRequirement:
     """
-    If {elementExists} is set, and {elementDoesNotExist} is not set,
-    any context with ID starting with {contextId} must have {element} value
+    If {elementExists} is set, that DEI value must be present.
+    If {elementDoesNotExist} is set, that DEI value must be absent.
+    If {contextIdExists} is set, some context ID in the same instance must start with that prefix.
+    If {contextIdDoesNotExist} is set, no context ID in the same instance may start with that prefix.
+    Any context with ID starting with {contextId} must have {element} value
     matching the {elementMatch} DEI value (adjusted by {dayAdjustment} days).
     """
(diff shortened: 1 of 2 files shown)


aaroncameron-wk:
[Validation Guidelines](https://disclosure2dl.edinet-fsa.go.jp/guide/static/disclosure/download/ESE140305.pdf) 3.4.2 has notes that essentially explain that these warnings might trigger in valid scenarios, but they can be ignored. I'm wondering if the scenario these changes are addressing is one of those scenarios. In which case, I think there's a larger discussion to be had around whether we want to match true EDINET results vs. differing from true EDINET for improved user experience.

arelle/plugin/validate/EDINET/ContextRequirement.py
```diff
@@ -20,35 +23,45 @@ class ContextRequirement:
...
     elementMatch: str
     dayAdjustment: int = 0  # days to adjust by (e.g. -1 for "day before")
+    contextIdExists: str | None = None
+    contextIdDoesNotExist: str | None = None
 
 
 # See section 3-4-2 in the Validation Guidelines.
+# Prior1Year/Prior2Year dates shift forward when next-FY interim DEI is present,
+# but only if the instance has no CurrentYear contexts (typical of registration
+# statements that disclose next-FY interim results).
```
aaroncameron-wk, line 33: Doesn't have to be with this PR, but would be good to at least have a valid testcase that represents "registration statements that disclose next-FY interim results" since we do not appear to cover it currently.

aaroncameron-wk (approved):
Separate discussion revealed that this does in fact put us closer into alignment with what live EDINET yields in this scenario.

arelle/plugin/validate/EDINET/ContextRequirement.py
allenbross-wf: I will write a follow up ticket so this can be released for consumption/testing.

PR #2557, opened 2026-08-19
Refine and update UKSEF1-2 validation and schema checks

volodymyrtur-wk:
#### Reason for change

Improve validation processes for UKSEF1-2, including error code clarity, schema reference accuracy, and alignment with correct taxonomy and conformance rules.

#### Description of change

- Renamed error codes for better identification.
- Adjusted schemaRef checks and updated validation for schemaRefs and taxonomy versions.
- Verified report conformance within the context of UKSEF validation.
- Updated FRC test configurations to reflect these changes.

#### Steps to Test
CI
**review**:
@Arelle/arelle


arelle/plugin/validate/ESEF/rules/UKSEF/taxonomy.py
```diff
@@ -46,6 +59,60 @@ def rule_ukfrc1(
...
+        ix_ns_tag = getattr(ixds_html_root_elt.modelDocument, "ixNStag", "")
+        for references_elt in ixds_html_root_elt.iterdescendants(tag=f"{ix_ns_tag}references"):
+            if references_elt.get("target") != TARGET_UKFRS:
+                continue
+
+            is_valid_target = True
+            for schema_ref in references_elt.iterdescendants(tag=_LINK_SCHEMA_REF):
+                href = schema_ref.get(_XLINK_HREF, "").strip()
+                if _UKSEF_ENTRY_POINT_PATTERN.match(href):
+                    found_hrefs.append(href)
```
aaroncameron-wk: I think we can take advantage of some parsing that Arelle has already done at this point. For example:
```
schemaRefElts = [
    ref
    for ref in model_xbrl.modelDocument.referencesDocument.values()
    if ref.referringModelObject.localName == 'schemaRef'
]
```
This is already scoped to target, so no additional target check is necessary.

arelle/plugin/validate/ESEF/rules/UKSEF/taxonomy.py
```diff
@@ -46,6 +59,60 @@ def rule_ukfrc1(
     """
     if val.authority != AUTHORITY_UKFRC:
         return None
+
+    model_xbrl = val.modelXbrl
+    if pluginData.isUkfrsTarget(model_xbrl):
+        return None
```
aaroncameron-wk: Each rule will be called once _per-target_, and since this rule pertains to the UKSEF (UKFRS) target, I think you want the inverse in this condition.

arelle/plugin/validate/ESEF/rules/UKSEF/taxonomy.py
```diff
@@ -46,6 +59,60 @@ def rule_ukfrc1(
...
+
+    ixds_html_elements = getattr(model_xbrl, "ixdsHtmlElements", None)
+    if not ixds_html_elements:
+        return None
```
aaroncameron-wk: Once you have a `yield` statement in the function, you can simplify the return type to only `Iterable[Validation]` and these return statements can be shortened to simply `return`

arelle/plugin/validate/ESEF/rules/UKSEF/taxonomy.py
```diff
@@ -46,6 +59,60 @@ def rule_ukfrc1(
     """
     if val.authority != AUTHORITY_UKFRC:
         return None
+
+    model_xbrl = val.modelXbrl
```
aaroncameron-wk: Variable casing is inconsistent in Arelle as a whole, but we do try and stay consistent within a given area. The ESEF plugin, for example, has already established `camelCase` for local variable names, and that's also what I used in the other UKSEF validation module, `entity.py`.

arelle/plugin/validate/ESEF/rules/UKSEF/taxonomy.py
```diff
@@ -46,6 +59,60 @@ def rule_ukfrc1(
...
+                    found_hrefs.append(href)
+
+    if not is_valid_target:
+        yield Validation.error(
+            codes="ESEF.UKFRC1.incorrectTarget",
+            msg=_(
+                "UKSEF reports MUST have a \"UKFRS\" targeted ix:references element. "
+                "No matching ix:references element was found in the report."
+            ),
+        )
```
aaroncameron-wk: I think this error should fire if `not pluginData.isUkfrsTarget(model_xbrl)` and `len(found_hrefs) > 0`, which indicates that a UKSEF entry point is assigned the default (incorrect) target.

arelle/plugin/validate/ESEF/rules/UKSEF/taxonomy.py
```diff
@@ -59,8 +126,67 @@ def rule_ukfrc2(
...
+            # Try to resolve via referenced document uris (href may differ from normalized uri).
+            for uri, doc in url_docs.items():
+                if uri.endswith(href) or href.endswith(uri):
+                    extension_doc = doc
+                    break
+        if extension_doc is None:
+            continue
+        for referenced_doc, doc_ref in extension_doc.referencesDocument.items():
+            if "import" in doc_ref.referenceTypes:
+                imported_urls.add(referenced_doc.uri)
```
aaroncameron-wk: I think this could be simplified to something like:
```
imported_urls: set[str] = set()
for doc in model_xbrl.urlDocs.values():
    if not isExtensionDoc(doc):
        continue
    for referenced_doc, doc_ref in doc.referencesDocument.items():
        if "import" in doc_ref.referenceTypes:
            imported_urls.add(referenced_doc.uri)
```
Look into how other plugins use `from arelle.utils.validate.Common import isExtensionUri` for ideas on how to identify extension documents.

aaroncameron-wk:
Good first pass!

It's functionally close but I think there's some ways to re-use existing work to simplify these implementations.

arelle/plugin/validate/ESEF/rules/UKSEF/taxonomy.py
```diff
@@ -8,19 +8,33 @@
...
+    r"^http[s]?://www\.esma\.europa\.eu/taxonomy/(\d{4})-\d{2}-\d{2}/esef_"
+)
+_MIN_ESEF_YEAR = 2022
+_LINK_SCHEMA_REF = f"{{{XbrlConst.link}}}schemaRef"
+_XLINK_HREF = f"{{{XbrlConst.xlink}}}href"
+
 
 @validation(
-    hook=ValidationHook.XBRL_FINALLY,
+    hook=ValidationHook.FINALLY,
```
aaroncameron-wk, line 39: Might be good to add a comment explaining why we are straying from the default here (to access `ixdsReferences`)

arelle/plugin/validate/ESEF/rules/UKSEF/taxonomy.py
```diff
@@ -46,6 +60,49 @@ def rule_ukfrc1(
...
+                'UKSEF reports MUST have a "UKFRS" targeted ix:references element. '
+                'No matching ix:references element was found in the report.'
+                ),
+            )
+
+    modelXbrl = val.modelXbrl
+    if not pluginData.isUkfrsTarget(modelXbrl):
+        return None
+
+    if targetIxReference := val.ixdsReferences.get(TARGET_UKFRS, []):
```
aaroncameron-wk: Nit: `targetIxReferences`

arelle/plugin/validate/ESEF/rules/UKSEF/taxonomy.py
```diff
@@ -46,6 +60,49 @@ def rule_ukfrc1(
...
+            codes="ESEF.UKFRC1.incorrectTarget",
+            msg=_(
+                'UKSEF reports MUST have a "UKFRS" targeted ix:references element. '
+                'No matching ix:references element was found in the report.'
+                ),
+            )
+
+    modelXbrl = val.modelXbrl
+    if not pluginData.isUkfrsTarget(modelXbrl):
+        return None
```
aaroncameron-wk: Just `return`, return type can be `-> Iterable[Validation]`.

arelle/plugin/validate/ESEF/rules/UKSEF/taxonomy.py
```diff
@@ -46,6 +60,49 @@ def rule_ukfrc1(
...
+                if _UKSEF_ENTRY_POINT_PATTERN.match(href):
+                    foundHrefs.append(href)
+
+        if len(foundHrefs) > 1:
+            yield Validation.error(
+                codes="ESEF.UKFRC1.multipleEntryPoints",
+                msg=_(
+                    'UKSEF reports MUST have a single schemaRef in a "UKFRS" targeted ix:references element. '
+                    'Multiple matching schemaRefs were found in the report.'
+                    ),
```
aaroncameron-wk, line 92: Validation errors can contain "references" to the elements that caused the error (via `modelObject=...`), and we try to provide those whenever possible. In this case, you could track `uksefSchemaRefs` instead of `foundHrefs`, and then provide `modelObject=uksefSchemaRefs` int the validation error.

arelle/plugin/validate/ESEF/rules/UKSEF/taxonomy.py
volodymyrtur-wk: mypy is throwing an error if I leave it blank

aaroncameron-wk: You need to change all three of the return statements and the return type.

tests/integration_tests/validation/conformance_suite_configurations/uksef.py
aaroncameron-wk: Test is failing because the testcase uses 2021 ESEF packages, so we'll need to add that here.

aaroncameron-wk (approved):

arelle/plugin/validate/ESEF/rules/UKSEF/taxonomy.py
```diff
@@ -57,10 +116,40 @@ def rule_ukfrc2(
...
+            if "import" in docRef.referenceTypes:
+                importedUrls.add(referencedDoc.uri)
+
+    esefYears: list[int] = []
+    for url in importedUrls:
+        match = _ESEF_TAXONOMY_URL_PATTERN.match(url)
+        if match:
+            esefYears.append(int(match.group(1)))
+
+    if esefYears and max(esefYears) < _MIN_ESEF_YEAR:
```
austinmatherne-wk, line 145: I would think multiple esef years will raise some other error, but should this be `min(esefYears)` not max? Also, should it raise if `esefYears` is empty? @aaroncameron-wk do you know?

arelle/plugin/validate/ESEF/rules/UKSEF/taxonomy.py
```diff
@@ -45,8 +61,51 @@ def rule_ukfrc1(
     The reference must be in the report, NOT in the extension taxonomy.
     """
     if val.authority != AUTHORITY_UKFRC:
-        return None
-    return None
+        return
+
+    if val.ixdsReferences and TARGET_UKFRS not in val.ixdsReferences:
```
austinmatherne-wk, line 66: I think this will throw an attribute error on `ixdsReferences` if a user selects this plugin+disclosure system, but then (incorrectly) loads a non-ixbrl report. We should handle that gracefully with getattr and a fallback None or empty dict

arelle/plugin/validate/ESEF/rules/UKSEF/taxonomy.py
```diff
@@ -45,8 +61,51 @@ def rule_ukfrc1(
...
+                href = schemaRef.get(_XLINK_HREF, "").strip()
+                if _UKSEF_ENTRY_POINT_PATTERN.match(href):
+                    uksefSchemaRefs.append(schemaRef)
+
+        if len(uksefSchemaRefs) > 1:
+            yield Validation.error(
+                codes="ESEF.UKFRC1.multipleEntryPoints",
+                msg=_(
+                    'UKSEF reports MUST have a single schemaRef in a "UKFRS" targeted ix:references element. '
+                    'Multiple matching schemaRefs were found in the report.'
```
austinmatherne-wk, line 91: The `found in the report` part of this error message suggests it should be raised on a uksef schema ref in a non-UKFRS target? @aaroncameron-wk should `pluginData.isUkfrsTarget` set the expected number of uksef schema refs (1 for UKFRS 0 for all others) and let the rule run against all targets?

aaroncameron-wk: `incorrectTarget` is expected to fire whenever any UKSEF schema ref exists in a non-UKFRS targeted `references` block, so the scenario you describe would at least be covered by that.

FRC_05/TC2_valid is an example of a testcase that has two UKSEF entry points, one within a UKFRS target and the other in a non-UKFRS target. The testcase has `ignore-error` for `incorrectTarget`, but no expectation of `multipleEntryPoints`, which suggests the existing implementation aligns with the conformance suite.

arelle/plugin/validate/ESEF/rules/UKSEF/taxonomy.py
aaroncameron-wk: Agreed on `min(esefYears)`.

I'm not seeing anything that suggests that `incorrectEsefTaxonomyVersionUsed` should fire if there are no/multiple entry points.

The no ESEF entry point scenario is part of FRC_05 (`noESEFData`).
The multiple ESEF entry point scenario does not appear to be covered.

PR #2560, opened 2026-08-20
Preprocessing function for conformance suite configs

aaroncameron-wk:
#### Reason for change
Some conformance suites references non-existent files, but the intent is clear. We can directly manipulate the file so we can still run the intended test.

#### Description of change
Add an optional `preprocessing_func` to conformance suite configurations.

#### Steps to Test
CI

**review**:
@Arelle/arelle

### DIFF
--- a/tests/integration_tests/validation/conformance_suite_configurations/uksef.py
+++ b/tests/integration_tests/validation/conformance_suite_configurations/uksef.py
@@ -6,6 +6,22 @@ from tests.integration_tests.validation.conformance_suite_config import (
 )
 
 
+def _preprocessing_func(config: ConformanceSuiteConfig) -> None:
+    with open(
+            config.entry_point_root /
+            "tests/FRC/FRC_09/index.xml",
+            "r+"
+    ) as f:
+        content = f.read()
+        # Test case references TC2_valid.zip, but actual file in suite has .xbri extension.
+        content = content.replace("TC2_valid.zip", "TC2_valid.xbri")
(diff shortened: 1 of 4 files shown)


tests/integration_tests/validation/conformance_suite_configurations/uksef.py
```diff
@@ -6,6 +6,22 @@
...
+    with open(
+            config.entry_point_root /
+            "tests/FRC/FRC_09/index.xml",
+            "r+"
+    ) as f:
+        content = f.read()
+        # Test case references TC2_valid.zip, but actual file in suite has .xbri extension.
+        content = content.replace("TC2_valid.zip", "TC2_valid.xbri")
+        # Test case references TC3_valid.zip, but actual file in suite has .xbri extension.
+        content = content.replace("TC3_valid.zip", "TC3_valid.xbri")
```
austinmatherne-wk, line 19: This is completely fine, but raising an exception if the expected string doesn't exist would be nice to help keep unneeded config out of the suite when future versions are released.

austinmatherne-wk (approved):

tests/integration_tests/validation/conformance_suite_configurations/uksef.py
aaroncameron-wk: The main reason I didn't do that is because for subsequent runs of the suite, the changes have already been applied, so no changes would actually be made. We could work around this by forcing the source zip to be re-extracted each run.

derekgengenbacher-wf (approved):

derekgengenbacher-wf:

PR #2579, opened 2026-09-08
Update UKSEF validation rules and enhance target attribute checks

volodymyrtur-wk:
#### Reason for change

The changes improve the validation logic for UKSEF, UKFRS, and ESEF reporting to ensure enhanced reliability and accuracy, while removing obsolete test case exclusions.

#### Description of change

- Removed resolved FRC test case exclusions from `expected_failure_ids`.
- Enhanced `target` attribute checks across UKFRC validations (FRC3-FRC5) and added validation for missing or invalid ESEF/UKFRS targets.
- Updated expected testcase errors for UKFRS and ESEF scenarios.
- Incorporated logic for disallowed elements and invalid `target` values in UKFRC3 validation.
- Updated `ixdsReferences` handling to include `None` in the type definition and remove unnecessary type ignores.
- Fixed logic to validate minimum ESEF year in UKFRS2 validation.

#### Steps to Test
CI

**review**:
@Arelle/arelle


allenbross-wf (dismissed):

derekgengenbacher-wf (dismissed):

arelle/plugin/validate/ESEF/rules/UKSEF/entity.py
```diff
@@ -36,9 +36,11 @@ def rule_ukfrc6(
     """
     if val.authority != AUTHORITY_UKFRC:
         return
+
     modelXbrl = val.modelXbrl
-    if pluginData.isUkfrsTarget(modelXbrl):
+    if getattr(modelXbrl, "ixdsTarget", "") is not None:
```
aaroncameron-wk: Suggest `isEsefTarget` helper similar to `isUkfrsTarget`, I think that would allow the code to be clearer that an absent `ixdsTarget` property is treated as a non-ESEF target and thus skips the validation in this case.

arelle/plugin/validate/ESEF/rules/UKSEF/target.py
```diff
@@ -8,68 +8,214 @@
...
+        local names.
+    """
+    return tuple(f"{{{ns}}}{ln}" for ln in local_names)
+
 
 @validation(
     hook=ValidationHook.XBRL_FINALLY,
 )
-def rule_ukfrc3(
+def ruleUkfrc3(
```
aaroncameron-wk: Inadvertent name change here, I'm guessing.

arelle/plugin/validate/ESEF/rules/UKSEF/target.py
```diff
@@ -8,68 +8,214 @@
...
     if val.authority != AUTHORITY_UKFRC:
-        return None
-    return None
+        return
+
+    # Check if the Inline XBRL document has a target attribute set to "UKFRS" (case-insensitive).
+    # If not, skip the validation for this rule.
+    ixdsTarget = getattr(val.modelXbrl, "ixdsTarget", None)
+    if ixdsTarget is None or isinstance(ixdsTarget, str) and ixdsTarget.upper() != TARGET_UKFRS:
+        return
```
aaroncameron-wk: Is there a reason we aren't using the `isUkfrsTarget` helper here?

arelle/plugin/validate/ESEF/rules/UKSEF/target.py
```diff
@@ -8,68 +8,214 @@
...
+            yield Validation.error(
+                codes="ESEF.UKFRC4.targetAttributeUsedForESEFContents",
+                msg=_(
+                    "ESEF tagged data MUST be in the default (unnamed) target XBRL document and "
+                    "MUST NOT carry a target attribute. Found ESEF concept element(s) with target attribute."
+                ),
+                modelObject=ESEFTargets,
+            )
+
+    return
```
aaroncameron-wk: These trailing `return` statements are not necessary

arelle/plugin/validate/ESEF/rules/UKSEF/target.py
```diff
@@ -8,68 +8,214 @@
...
-        return None
-    return None
+        return
+
+    if not pluginData.isUkfrsTarget(val.modelXbrl):
+        return
+
+    ixdsReferences = getattr(val, "ixdsReferences", None)
+    if ixdsReferences:
+        ESEFTargets = ixdsReferences.get(None, [])
```
austinmatherne-wk, line 105: Could we check the ESEF facts themselves for a target attribute? Having a default/unnamed `ix:references` container doesn’t prevent ESEF facts from being assigned to `UKFRS`

arelle/plugin/validate/ESEF/rules/UKSEF/target.py
```diff
@@ -8,68 +8,214 @@
...
+        ns = getattr(ixdsHtmlRootElt.modelDocument, "ixNS", ixbrl11)
+        allowedTags = _ix_tags_for_namespace(ns, _TARGET_ALLOWED_LOCAL_NAMES)
+        disallowedTags = _ix_tags_for_namespace(ns, _TARGET_DISALLOWED_LOCAL_NAMES)
+
+        # Also cover the ix 1.0 namespace to be safe when both are present.
+        otherNs = ixbrl if ns == ixbrl11 else ixbrl11
+        allowedTags = allowedTags + _ix_tags_for_namespace(otherNs, _TARGET_ALLOWED_LOCAL_NAMES)
+        disallowedTags = disallowedTags + _ix_tags_for_namespace(otherNs, _TARGET_DISALLOWED_LOCAL_NAMES)
+
+        for elt in ixdsHtmlRootElt.iter():
```
austinmatherne-wk: nit: it's more efficient to let lxml filter elements in the Cython/native code level:
```suggestion
        for elt in ixdsHtmlRootElt.iter(*allowedTags, *disallowedTags):
```

arelle/plugin/validate/ESEF/rules/UKSEF/target.py
```diff
@@ -8,68 +8,214 @@
...
+
+        if (pluginData.isUkfrsTarget(val.modelXbrl)
+                and (foundTargets := ixdsReferences.get(TARGET_UKFRS, []))
+                and len(foundTargets) > 1):
+
+            yield Validation.error(
+                codes="ESEF.UKFRC5.multipleEntryPoints",
+                msg=_(
+                    'UKSEF reports MUST have a single targeted element with the "UKFRS" target. '
+                    'Multiple matching ix:references elements were found in the report.'
```
austinmatherne-wk, line 201: The UKFRC5 docstring above describes both the reference containers and the schemas they contain. The current implementation checks the reference targets, but doesn’t inspect the schemas within them.

It seems like the validation flow should be:

1. Validate that exactly two `ix:references` containers exist.
2. Validate that one has `target="UKFRS"` and references a supported UKSEF entry point schema.
3. Validate that the other omits `target`, and does not reference a UKSEF entry point schema.

arelle/plugin/validate/ESEF/rules/UKSEF/target.py
volodymyrtur-wk: this's a case-insensitive check unlike `isUkfrsTarget`, but I removed it because it is not needed anymore

arelle/plugin/validate/ESEF/rules/UKSEF/target.py
volodymyrtur-wk: There are strict rules for defining facts:
1. The report must contain 2 `ix:references` containers.
2. One container must include a `target` attribute with a UKFRS value, which refers to UKFRS facts.
3. The second container is the default and does not include the `target` attribute, which refers to ESEF facts.

I'll rewrite this rule because we found that the description of the ticket doesn't correspond to the real rule signature

arelle/plugin/validate/ESEF/rules/UKSEF/target.py
volodymyrtur-wk: In this rule, we don't need to check references to a supported UKSEF entry point schema; ukfrc1 does that. A few other rules check whether the value in the target attribute is wrong. Here, we check only the correct structure of `ix:references` elements

aaroncameron-wk: We've had some discussion about whether or not it makes more sense to align the rules to the "UKFRC#" rules (i.e. `rule_ukfrc5`), or instead to implement them based on error code (i.e. `rule_multipleEntryPoints`). My initial scaffolding of the plugin assumed the former, but I'm leaning more towards the latter as we learn more about the ruleset. @volodymyrtur-wk It may be worth experimenting with implementing `rule_incorrectTarget` and `rule_multipleEntryPoints` to see if it might simplify these implementations.

arelle/plugin/validate/ESEF/rules/UKSEF/target.py
```diff
@@ -8,68 +8,197 @@
...
     """
-    if val.authority != AUTHORITY_UKFRC:
-        return None
-    return None
+    if val.authority != AUTHORITY_UKFRC or pluginData.isEsefTarget(val.modelXbrl):
+        return
+
+    ixdsReferences = getattr(val, "ixdsReferences", None)
+
+    if ixdsReferences and not ixdsReferences.get(TARGET_UKFRS, []):
```
austinmatherne-wk: Is the `UKFRS` target presence check intentional?

I would expect both of these filings to raise `incorrectTarget`:
Filing 1 targets: `(default)`, `UKFRS`, `OTHER`
Filing 2 targets: `(default)`, `OTHER`

As written, only filing 1 does.

arelle/plugin/validate/ESEF/rules/UKSEF/target.py
```diff
@@ -8,68 +8,197 @@
...
+        for target, referencesElts in ixdsReferences.items():
+            if target is not None and target.upper() == TARGET_UKFRS:
```
austinmatherne-wk: The error message below says `The target attribute on Inline XBRL elements in a UKSEF report MUST be "UKFRS" (case-sensitive)`, which suggests:

```suggestion
            if target is not None and target != TARGET_UKFRS:
```

arelle/plugin/validate/ESEF/rules/UKSEF/target.py
```diff
@@ -8,68 +8,257 @@
...
+) -> Iterable[Validation]:
+    """
+    Validate all UKSEF multiple-entry-point conditions in one place.
+
+    UKFRC1 limits the matching UKSEF schemaRefs in a single UKFRS target to one.
+    UKFRC5 limits the UKFRS-targeted ix:references elements to one.
+    """
+    if (val.authority != AUTHORITY_UKFRC
+            or not pluginData.isUkfrsTarget(val.modelXbrl)
+            or not (ixdsReferences := getattr(val, "ixdsReferences", None))):
```
aaroncameron-wk: Is `getattr` strictly necessary here?

arelle/plugin/validate/ESEF/rules/UKSEF/target.py
```diff
@@ -8,68 +8,257 @@
...
+        yield Validation.error(
+            codes="ESEF.UKFRC5.multipleEntryPoints",
+            msg=_(
+                'UKSEF reports MUST have a single targeted element with the "UKFRS" target. '
+                "Multiple matching ix:references elements were found in the report."
+            ),
+            modelObject=foundTargets,
+        )
+
+    if len(foundTargets) == 1:
```
aaroncameron-wk: I think this can just be `else:`. It can at least be `elif len(foundTargets) == 1:`

arelle/plugin/validate/ESEF/rules/UKSEF/target.py
```diff
@@ -8,68 +8,257 @@
...
 
 _: TypeGetText
 
+# Inline XBRL elements on which the "target" attribute may legitimately appear.
+_TARGET_ALLOWED_LOCAL_NAMES = ("nonFraction", "nonNumeric", "footnote", "references")
+# Inline XBRL elements on which the "target" attribute MUST NOT appear.
+_TARGET_DISALLOWED_LOCAL_NAMES = ("resources", "continuation", "exclude")
+
+
+def _ix_tags_for_namespace(ns: str, local_names: Iterable[str]) -> tuple[str, ...]:
```
aaroncameron-wk: Nit: `_ixTagsForNamespace`

arelle/plugin/validate/ESEF/rules/UKSEF/target.py
```diff
@@ -8,68 +8,257 @@
...
+        for elt in ixdsHtmlRootElt.iter(*allowedTags, *disallowedTags):
+            tag = getattr(elt, "tag", None)
+            if not isinstance(tag, str) or "target" not in elt.attrib:
+                continue
+
+            targetValue = elt.get("target") or ""
+            if tag in allowedTags and targetValue == TARGET_UKFRS:
+                continue
+
+            if tag in allowedTags:
```
aaroncameron-wk:
```suggestion
            if tag in disallowedTags:
```

tests/integration_tests/validation/conformance_suite_configurations/uksef.py
```diff
@@ -76,14 +76,58 @@ def _preprocessing_func(config: ConformanceSuiteConfig) -> None:
...
+            # and we can't separate schemas iso17442 and ENTITY_IDENTIFIER_SCHEME_CRN
+            "invalidIdentifier": 1,
+            # same explanation as invalidIdentifier error above
+            "multipleIdentifiers": 1,
+            "segmentUsed": 1,
+        },
+        "FRC_03/index.xml:TC3_invalid": {
+            # rule_incorrectTarget covers the UKFRC1 and UKFRC3 incorrectTarget conditions.
+            # UKFRC1 and UKFRC5 have the same conditions for the test case, but have different checks and fire different errors
+            "noUKFRSData": 2,
```
aaroncameron-wk, line 106: Just a note:

This fires twice because `rule_noUKFRSData` fires once for each target. We can't switch to a hook like `COMPLETE` which would cause the rule to run once across the entire filing because we need access to `ValidateXbrl.ixdsReferences`, and `ValidateXbrl` is scoped to a target level (even though the value itself includes *all* reference elements across targets). A refactor of how we can efficiently inspect references elements is probably necessary, but this is fine for this PR.

tests/integration_tests/validation/conformance_suite_configurations/uksef.py
```diff
@@ -76,14 +76,58 @@ def _preprocessing_func(config: ConformanceSuiteConfig) -> None:
...
+        "FRC_04/index.xml:TC2_invalid": {
+            "targetAttributeUsedForESEFContents": 1,
+            # Data in this test case is invalid for the rule UKFRC5
+            "noESEFData": 2,
+        },
+        "FRC_05/index.xml:TC2_valid": {
+            "targetAttributeUsedForESEFContents": 1,
+        },
+        "FRC_05/index.xml:TC3_valid": {
+            "targetAttributeUsedForESEFContents": 1,
```
aaroncameron-wk: It isn't obvious to me why this code should be firing on these testcases. Could be related to my other comment on `rule_targetAttributeUsedForESEFContents`? If not, a comment as an explanation seems appropriate.

arelle/plugin/validate/ESEF/PluginValidationDataExtension.py
```diff
@@ -23,6 +28,35 @@ def getContextsByEntityIdentifier(self, modelXbrl: ModelXbrl) -> dict[tuple[str,
...
-            return False
-        return cast(str | None, modelXbrl.ixdsTarget) == TARGET_UKFRS
+        """Check if the target document is a UKFRS target."""
+        ixdsTarget: str | None = getattr(modelXbrl, "ixdsTarget", "")
+        return ixdsTarget == TARGET_UKFRS
+
+    def isEsefTarget(self, modelXbrl: ModelXbrl) -> bool:
+        """Check if the target document is an ESEF target."""
+        ixdsTarget: str | None = getattr(modelXbrl, "ixdsTarget", "")
+        return ixdsTarget is None
```
aaroncameron-wk, line 39: I prefer the previous implementation because it still holds a strong reference to `ModelXbrl.ixdsTarget` which allows our IDE to find it. It also makes it very clear that the absence of the attribute is a special case, rather than it being sort of implicit by defaulting to `""` with `getattr`.

derekgengenbacher-wf (approved):

derekgengenbacher-wf:

PR #2587, opened 2026-09-13
Fix qname nullable return type

austinmatherne-wk:
#### Reason for change

`qname` overloads incorrectly promised a `QName` for calls that can return `None`. It also includes runtime parsing and error handling logic that's unnecessary for known values.

#### Description of change

* add simple `QName.fromParts` factory
* migrate explicit construction
* reuse existing constants
* correct `qname` overloads and update affected callers to handle nullable results

#### Steps to Test

* CI

**review**:
@Arelle/arelle


volodymyrtur-wk (approved):

derekgengenbacher-wf (approved):

aaroncameron-wk (approved):

arelle/ModelFormulaObject.py
```diff
@@ -34,7 +34,7 @@
 
 class FormulaOptions:
     def __init__(self, savedValues: dict[str, Any] | None = None) -> None:
-        self.parameterValues: dict[QName, Any] = {} # index is QName, value is typed value
+        self.parameterValues: dict[QName | None, Any] = {}
```
brettkail-wk, line 37: Out of curiosity, what does a `None` key mean (for all the `dict[QName,` -> `dict[QName | None,` in the PR)?  Would it be better to avoid inserting `None`?

arelle/ViewWinRenderedGrid.py
```diff
@@ -801,7 +801,9 @@ def updateInstanceFromFactPrototypes(self) -> None:
                                 # new context
                                 if concept.isNumeric:
                                     if concept.isMonetary:
-                                        unitMeasure = qname(XbrlConst.iso4217, self.newFactItemOptions.monetaryUnit)
+                                        monetaryUnitMeasure = qname(XbrlConst.iso4217, self.newFactItemOptions.monetaryUnit)
+                                        assert monetaryUnitMeasure is not None, "A fixed namespace and string name always produce a QName"
```
brettkail-wk, line 805: Out of curiosity, why not use `fromParts` here?  If `qname` is still preferable for some reason, could these `assert` be removed if we added more non-`None` typing overloads?

brettkail-wk (approved):

derekgengenbacher-wf:

arelle/ViewWinRenderedGrid.py
austinmatherne-wk: The docstring on FactItemOptions said it was a prefix:localName string, so I stuck with `qname` for parsing, but the docstring was wrong. Updated in #2590

arelle/ModelFormulaObject.py
austinmatherne-wk: Invalid options passed from the CLI. I added a check to avoid inserting them in #2590

PR #2588, opened 2026-09-14
Use `ixdsFootnotesById` for `uniqueFootnoteId` validation

aaroncameron-wk:
#### Reason for change
The implementation for `uniqueFootnoteId` seems to expect that `ixdsFootnotes` is a mapping of `footnote ID -> list of footnote elements`. However, `ixdsFootnotes` is actually a mapping of `footnote ID -> last validated footnote element`.

This means that when the "length" of the expected list is checked, it is actually checking the number of child elements directly beneath the footnote element, meaning a footnote like this:
Screenshot 2026-09-14 at 10 32 25 AM
would trigger the logic for firing `uniqueFootnoteId` since `len(objs) > 1`.

#### Description of change
Correct this by tracking a separate `ixdsFootnotesById` that *does* collect a list of footnote elements associated with that footnote ID, which allows the `uniqueFootnoteId` implementation to fire as expected.

#### Steps to Test
CI

**review**:
@Arelle/arelle


derekgengenbacher-wf (approved):

arelle/ValidateXbrlDTS.py
```diff
@@ -966,6 +967,9 @@ def linkbaseTopElts() -> Iterable[ModelObject]:
                                 _("Inline XBRL footnote %(footnoteID)s is missing an xml:lang attribute"),
                                 modelObject=elt, footnoteID=id)
                     if elt.namespaceURI == XbrlConst.ixbrl:
+                        if elt.footnoteID not in val.ixdsFootnotesById:
+                            val.ixdsFootnotesById[elt.footnoteID] = []
```
brettkail-wk, line 971: I see `ixdsReferences` uses `defaultdict(list)`.  Out of curiosity, when is that preferred?

brettkail-wk (approved):

derekgengenbacher-wf:

PR #2589, opened 2026-09-14
WebCache: retry downloads failing with transient HTTP or connection errors

hermfischer-xb:
#### Reason for change

This arose in bulk validation of EDGAR filings from sec.gov monthly XBRL RSS feeds (one feed month is about 19,000 filings). In a single sequential process, well within SEC's fair-access rate, 112 of the first 3,856 filings of a month received HTTP 503 Service Unavailable for their filing zip. The ones that validated did so only because the zip is requested again while the filing opens; some never downloaded and were not validated.

Now, within the existing retry count (RETRIEVAL_RETRY_COUNT):
- HTTP 429, 500, 502, 503 and 504, and a URLError whose reason is a timeout or a dropped connection, are retried; other failures, such as 404, still fail at once.
- Each retry waits: the server's Retry-After header (seconds or an HTTP date) when given, else a delay doubling from one second (1, 2, 4, 8).
- The total wait for one download is limited to TRANSIENT_RETRY_WAIT_BUDGET_SECONDS (20). A download holds the cache file lock, and another process waiting on that lock gives up after FILE_LOCK_TIMEOUT (30), so a long Retry-After is shortened to the time left.
- Each retry is logged as webCache:retryingOperation, as other retries are.
- A failure while rechecking an already cached file still keeps the cached file without waiting.

tests/unit_tests/arelle/test_webcache.py covers retry then success, no retry for 404, doubling delays until the retries are used, Retry-After, the wait budget, a dropped connection, cache recheck, and the classification of transient errors.

#### Description of change

WebCache._downloadFile retries a truncated transfer and any unclassified exception, but its except clause for HTTPError and URLError, needed for the 401/407 authentication handling, catches every HTTP status first. So a 503 was logged once as webCache:retrievalError and treated as permanent, like a 404, and connection failures raised as URLError were not retried either.

#### Steps to Test

Run an SEC RSS monthly, was getting 503's intermittently before fix.

**review**:
@Arelle/arelle


arelle/WebCache.py
```diff
@@ -874,6 +883,18 @@ def _downloadFile(
                 if retrievingDueToRecheckInterval:
                     self.internetRecheckFailedRecovery(url, err, timeNowStr)
                     return True
+                if (retryCount > 1 and WebCache._isTransientRetrievalError(err) and
+                        transientRetryWaitSeconds < TRANSIENT_RETRY_WAIT_BUDGET_SECONDS):
+                    delay = min(WebCache._transientRetryDelay(err, initialRetryCount - retryCount),
```
austinmatherne-wk: `Retry-After` is potentially shortened below the server's requested delay here. If only five seconds remain in the local budget and the provider says 15 seconds, Arelle would sleep five seconds and send another request before the server's requested delay. If the `Retry-After` delay exceeds the remaining budget I think we should fail without retrying, otherwise we risk triggering server penalties.

arelle/WebCache.py
```diff
@@ -52,6 +52,13 @@
 FILE_LOCK_TIMEOUT = 30
 INF = float("inf")
 RETRIEVAL_RETRY_COUNT = 5
+# HTTP status codes of server conditions that a repeated request may not meet: too many requests, and server
+# errors such as 503 Service Unavailable, which a busy server returns during sustained retrieval of many files.
+TRANSIENT_HTTP_STATUS_CODES = frozenset({429, 500, 502, 503, 504})
```
austinmatherne-wk: I did a quick check of this list [against other tools](https://curl.se/docs/manpage.html#--retry) to see which codes are commonly considered transient. `408` [looks like](https://www.rfc-editor.org/rfc/rfc9110.html#section-15.5.9) another good candidate for inclusion here.

arelle/WebCache.py
hermfischer-xb: Retry-After no longer clipped. If the server's requested delay doesn't fit the remaining budget, the download fails instead of sleeping less and retrying early. Two tests cover it: a second 15 s wait that ends the retries, and a Retry-After longer than the whole budget that isn't retried at all.

arelle/WebCache.py
hermfischer-xb: 408 added to the transient status codes, with the comment updated.

hermfischer-xb:
Both of these are addressed in f7174ea9b:

- Retry-After is no longer shortened to fit the remaining budget. When the requested
  delay doesn't fit, the download fails without retrying, so a request is never repeated
  before the delay the server asked for. Tests cover a wait that ends the retries and a
  Retry-After longer than the whole budget, which is not retried at all.
- 408 added to TRANSIENT_HTTP_STATUS_CODES.

Some field data behind this change, from bulk validation of EDGAR monthly RSS feeds.

Validating the August 2026 feed (19,037 filings, one process, no fix) logged 199
webCache:retrievalError entries across 150 filings. None carried "Forbidden" or "Too Many
Requests" text; they were transient server errors, and each cost that filing its validation.

Downloading a month of filing zips in parallel showed where they come from. With 10 workers
under a shared 8 requests/s cap, per-minute totals were:

  minute 1: 325 files, 5.4/s, 503s 8
  minute 2: 726 files, 6.7/s, 503s 13
  minute 3: 1036 files, 5.2/s, 503s 16
  minute 4: 1227 files, 3.2/s, 503s 25

No 429 and no 403 at any point, and all 25 requests that drew a 503 succeeded when repeated
singly. Throughput decayed as retries took a growing share of the rate budget, so we settled
at 5 workers and 4 requests/s. That is the condition this branch targets: a busy server
shedding load during sustained retrieval, where the retrieval succeeds shortly afterwards.

Truncated bodies are already covered upstream: _downloadFile catches
(ContentTooShortError, IncompleteRead) and retries, so this branch only adds the transient
HTTP and connection path beside it.

austinmatherne-wk (approved):

PR #2593, opened 2026-09-15
RSS feed validation: collect garbage after each item

hermfischer-xb:
#### Reason for change

Validating a month-long EDGAR XBRL RSS feed (xbrlrss-2026-08, 19,037 filings in one process) grew resident memory by about 165-200 kB per filing, reaching 4.3 GB.  Closed filing models, their documents and lxml trees are left in reference cycles, and the cycle collector's automatic passes, triggered by allocation counts, let several closed filings stay alive between passes.

#### Description of change

Measured over the first 300 filings of that feed: without a per-item collection, 6 to 15 closed ModelXbrl and 68 to 142 ModelDocument objects were alive at checkpoints and resident memory reached 1,517 MB; with gc.collect() after each item, live model counts stayed constant (only the feed and shared documents), gc-tracked objects stayed flat near 98,000, resident memory was 1,085 MB, and nothing was uncollectable.  Wall time was unchanged (130 s against 132 s).

#### Steps to Test

Memory profiling on SEC monthly RSS feeds.

**review**:
@Arelle/arelle


arelle/Validate.py
```diff
@@ -251,6 +252,11 @@ def validateRssFeed(self) -> None:
                 except Exception as err:
                     pass
             del modelXbrl  # completely dereference
+            # A closed filing's model objects and lxml trees remain in reference cycles, which only the cycle
+            # collector frees.  Its automatic passes are triggered by allocation counts, not memory, so over a
+            # long feed several closed filings stay alive between passes and resident memory grows with the
+            # largest of them.  Collecting once per item keeps memory to about one filing at a time.
+            gc.collect()
```
austinmatherne-wk: [`ModelManager.close()`](https://github.com/Arelle/Arelle/blob/f72fd5d8a4cfcff08620f760f5ba8d72ac0aa74f/arelle/ModelManager.py#L240-L257) already handles closing a model and collecting garbage. Could `_closeRssItemModelXbrl()` use that existing close path? That would also prevent early continue paths from skipping collection.

hermfischer-xb: Thanks, good suggestion. _closeRssItemModelXbrl now closes the item through ModelManager.close(modelXbrl), so collection also happens for items skipped by doNotProcessRSSitem and after a validation exception. The item's modelXbrl is passed explicitly, since the model manager's current modelXbrl is the feed. Tests cover the normal, skipped and exception paths.

austinmatherne-wk (approved):

PR #2594, opened 2026-09-16
Accept full XML Schema year range in table constraints

austinmatherne-wk:
#### Reason for change

Table constraints validation went through the model's XML validation, which limits years to the datetime range. Valid XML Schema values such as `-12024-01-01` or `12024` were rejected.

#### Description of change

Parses `xs:date`, `xs:dateTime`, `xs:gYear` and `xs:gYearMonth` with any valid XML Schema year for table constraints. Table constraints now gives correct results when validating a report without loading it into a model (`--validateTableConstraintsSkipLoading`). Arelle will still raise an error on reports with these dates if it is attempted to load them into the core Arelle model.

#### Steps to Test

* CI

**review**:
@Arelle/arelle


arelle/oim/_tc/xs_dates.py
```diff
@@ -0,0 +1,151 @@
...
+def _sign(value: Decimal) -> int:
+    if value < 0:
+        return -1
+    if value > 0:
+        return 1
+    return 0
+
+
+def is_leap_year(year: int) -> bool:
+    return year % 4 == 0 and (year % 100 != 0 or year % 400 == 0)
```
volodymyrtur-wk: To check if the year is a leap year, you can use the standard package `calendar` and the function `isleap(year)`

arelle/oim/_tc/xs_dates.py
```diff
@@ -0,0 +1,151 @@
...
+
+
+def days_in_month(year: int, month: int) -> int:
+    if month == 2:
+        return 29 if is_leap_year(year) else 28
+    return 30 if month in (4, 6, 9, 11) else 31
```
volodymyrtur-wk: You can simplify this using the package `calendar`:
```python
from calendar import monthrange

def days_in_month(year: int, month: int) -> int:
    return monthrange(year, month)[1]
```

arelle/oim/_tc/xs_dates.py
```diff
@@ -0,0 +1,151 @@
...
+
+    This is Howard Hinnant's days_from_civil algorithm.
+    """
+    if month <= 2:
+        year -= 1
+    era = year // 400
+    year_of_era = year - era * 400
+    day_of_year = (153 * (month + (-3 if month > 2 else 9)) + 2) // 5 + day - 1
+    day_of_era = year_of_era * 365 + year_of_era // 4 - year_of_era // 100 + day_of_year
+    return era * 146097 + day_of_era - 719468
```
volodymyrtur-wk, line 80: I want to suggest a shorter approach using `datetime` package:
```python
from datetime import date

def days_since_epoch(year: int, month: int, day: int) -> int:
    """Days from 1970-01-01 to the given Gregorian date, for any integer year.

    This is Howard Hinnant's days_from_civil algorithm.
    """
    return (date(year, month, day) - date(1970, 1, 1)).days
```

austinmatherne-wk: This module specifically exists because the Python datatime module [doesn't support](https://docs.python.org/3/library/datetime.html#datetime.MINYEAR) the full XML Schema date range. Using `datetime` here would raise `ValueError` on valid dates in the table constraints conformance suite.

arelle/oim/_tc/xs_dates.py
austinmatherne-wk: Thanks. I was too focused on not being able to use Python's `datatime` that I didn't consider that calendar can still be used.

volodymyrtur-wk (approved):

derekgengenbacher-wf (approved):

derekgengenbacher-wf:

PR #2596, opened 2026-09-16
Type hint audit ModelObject

austinmatherne-wk:
#### Reason for change

The base class declarations on `ModelObject` promised non-None values that every subclass property can return as `None`.

#### Description of change

* Declare `ModelObject` fields as optional.
* Skip inline footnotes without an id when building the IXDS footnote maps. Schema validation already reports the missing attribute, and the old `None` key could produce a misleading duplicate id error.
* Fall back to the standard reference role in the concept property view instead of raising `TypeError` on a reference without `xlink:role`.
* Drop formula rules whose QName has an unbound prefix instead of storing them under a `None` key that evaluation already ignored. DTS validation already raises errors on these.

#### Steps to Test

* CI

**review**:
@Arelle/arelle


derekgengenbacher-wf (approved):

derekgengenbacher-wf:

PR #2600, opened 2026-09-17
Implement base64Binary pattern matcher for improved validation

### LINKED ISSUE #2595: MemoryError validating large base64Binary values (nested-quantifier lexical pattern)
**Summary**

`arelle.XmlValidate.lexicalPatterns["base64Binary"]` is a nested-quantifier regular expression. Matching it against a multi-megabyte base64 value raises `MemoryError` (or effectively hangs while consuming all available RAM) during instance discovery (`instanceDiscover` → `xmlValidate` → `XmlValidate.py:768`), so affected filings cannot be loaded at all — `validate=False` does not avoid the call path.

**Where**

`arelle/XmlValidate.py`, `lexicalPatterns` (line ~143):

```python
"base64Binary": re_compile(
    r"((([A-Za-z0-9+/]\s?){4})*(([A-Za-z0-9+/]\s?){3}[A-Za-z0-9+/]|([A-Za-z0-9+/]\s?){2}[AEIMQUYcgkosw048]\s?=|[A-Za-z0-9+/]\s?[AQgw]\s?=\s?=))?$"
),
```

The pattern contains a nested quantifier `(([A-Za-z0-9+/]\s?){4})*` — a repetition of a repetition — over an unbounded-length input. With the `regex` module (used via `re_compile`), matching a large input appears to allocate unbounded memory.

**Observed**

* arelle-release 2.44.0 and 2.45.0 (pattern identical in both).
* Real-world trigger: CNMV IPP (semi-annual financial report) XBRL instances embed the source PDF in `xbrli:base64BinaryItemType` facts (~8 MB of base64 text per fact). Loading such an instance raises `MemoryError` on a 16 GB machine.
* The lexical check is only meaningful for *validating* the base64 alphabet/length modulo-4 structure — it does not need a byte-exact grammar over the whole payload.

**Minimal reproducer**

```python
import base64
from arelle.XmlValidate import lexicalPatterns

payload = base64.b64encode(b"\xab" * (9 * 1024 * 1024)).decode("ascii")  # ~12 MB
lexicalPatterns["base64Binary"].match(payload)   # MemoryError (or eats all RAM)
```

Run in a child process — the match can consume all available memory before raising.

**Suggested direction**

Replace the full-input regex match with a linear check, e.g. a bounded scan for the base64 alphabet plus validation of the (optional) `=`/`==` padding tail, or skip the lexical match for values above a size threshold. Happy to provide the exact fact elements / filing samples privately if useful.

**Context**

Found while parsing official CNMV IPP filings with `internetConnectivity="offline"`; mitigated downstream with a version/fingerprint-gated equivalent linear validator.

volodymyrtur-wk:
#### Reason for change
The issue https://github.com/Arelle/Arelle/issues/2595

#### Description of change

Improved XSD base64Binary validation. Instead of relying on a loose regex, it adds a custom matcher that properly enforces XSD lexical rules: ignores whitespace, validates padding/length constraints, and accepts only legal base64 characters. This makes validation more accurate for large or edge-case values.

#### Steps to Test
CI

**review**:  
@Arelle/arelle


arelle/XmlValidate.py
```diff
@@ -140,14 +140,52 @@ def isXValid(self) -> bool:
...
+class _Base64BinaryPattern:
+    """Linear matcher for the lexical space of xsd:base64Binary."""
+
+    def match(self, value: str) -> Match[str] | None:
+        dataLength = 0
+        paddingLength = 0
+        lastDataChar = ""
+
+        for char in value:
+            if char.isspace():
```
austinmatherne-wk: I suspect we need a literal check against [XML permitted whitespace chars](https://www.w3.org/TR/xml/#NT-S). Does `isspace()` match, or does it permit or exclude additional whitespace chars?

volodymyrtur-wk: This method checks whether all characters in the string are whitespace. In our case, we check each character of the string and ignore all whitespace, including tabs and newlines

austinmatherne-wk: Right. The issue is that "whitespace" can include a variety of different characters (tabs, new lines, etc.) The XML spec strictly defines which of these whitespace characters are permitted. If `isspace` doesn't use the same strict set of characters (I don't see why it would), then it will potentially skip "whitespace" that XML forbids.

volodymyrtur-wk: Validation occurs on these decoded string values. Here's an example of how it works with whitespaces:
```python
from lxml import etree


xml_content = b"""<root>
    <data>   &#x20; &#x09; &#x0A; &#x0D;   </data>
</root>"""

root = etree.fromstring(xml_content)
element_text = root.find("data").text
assert all(char.isspace() for char in element_text)
```

austinmatherne-wk: My concern is that `isspace()` also accepts characters outside XML’s whitespace definition. For example:

```python
root = etree.fromstring(b"<data>&#xA0;AAAA</data>")
text = root.text

assert text[0] == "\u00a0"
assert text[0].isspace()
assert text[0] not in " \t\r\n"
```

volodymyrtur-wk: Do we need to keep Nbsp?

austinmatherne-wk: We should reject it. The four permitted XML whitespace chars map to `" \t\r\n"`

austinmatherne-wk (dismissed):

arelle/XmlValidate.py
```diff
@@ -140,21 +140,59 @@ def isXValid(self) -> bool:
...
+            isValid = dataLength % 4 == 0
+        elif paddingLength == 1:
+            isValid = dataLength % 4 == 3 and lastDataChar in "AEIMQUYcgkosw048"
+        else:
+            isValid = dataLength % 4 == 2 and lastDataChar in "AQgw"
+
+        return _base64BinaryMatchPattern.match(value) if isValid else None
+
+
+_base64BinaryMatchPattern = re_compile(r"[\s\S]*")
```
brettkail-wk: Out of curiosity, do we need to return a real `re.Match` object, or could we just `return isValid`?

If we need a real object for some reason, is it worth adding a comment explaining that `[\s\S]*` is a simple pattern that matches the entire string just to get an `re.Match` object?

tests/unit_tests/arelle/test_xmlvalidate.py
```diff
@@ -1488,6 +1489,20 @@ def test_fractionDigits_facet_integer_types(self, base_xsd_type: str, value: str
...
+        ("AAAAAAAA", VALID),
+        ("A A A A", VALID),
+        ("AA=A", INVALID),
+        ("====", INVALID),
+        ("\u00a0B", INVALID),
+        ])
+    def test_large_base64_binary_is_validated_linearly(self, value: str, expected_result: int):
+        value *= (1024 * 1024)
+
+        # assert lexicalPatterns["base64Binary"].match(value) is not None
```
brettkail-wk: Remove the commented out code and unused `import`?

arelle/XmlValidate.py
volodymyrtur-wk: All other entries in `lexicalPatterns` are standard compiled `re.Pattern[str]` objects whose match() methods return `Match[str] | None`.  `_Base64BinaryPattern` acts as a drop-in duck-typed replacement for `re.Pattern[str]`. We can't return a boolean value because `if match is None` checks yield the wrong result.

brettkail-wk: Sure, then `return _SENTINEL_MATCH if isValid else None` (`re.match("", "")`)?  Alternatively, could we define some protocol like `XsdMatch` with the subset of methods actually needed so that `_Base64BinaryPattern` can return a simply object implementing the protocol?

Anyway, it just seems weird to have to run a full RE just to satisfy `Pattern` if call-sites ultimately just care about a truthy value, but if that's really the best approach, I wonder if `".*", DOTALL` is preferable to `r"[\s\S]*"`.

volodymyrtur-wk: Good point. Using a precomputed sentinel match object like `_SENTINEL_MATCH` is indeed much better than executing a regular expression on the payload on every call, which we don't need.
Since `_SENTINEL_MATCH` already satisfies `Match[str]` out of the box with zero boilerplate, defining a custom `Protocol` adds architectural ceremony without performance or safety benefits.

derekgengenbacher-wf (approved):

austinmatherne-wk (approved):

arelle/XmlValidate.py
brettkail-wk: To be sure, the disadvantage of a sentinel is that all the properties will be wrong (start, end, etc.).  Perhaps that's not important for any callers using it for base64(?), but it might be worth mentioning in the comment.  The advantage of a protocol object is it can be implemented with minimal state (e.g., the length of the str), but the disdavantage is the complexity.

brettkail-wk (approved):

derekgengenbacher-wf:

PR #2603, opened 2026-09-18
EDINET.EC8023W: Fix IndexError when a fact's parent is an inline fact

hermfischer-xb:
#### Reason for change

`rule_EC8023W` raises `IndexError: list index out of range` on ordinary EDINET markup, aborting EDINET validation for the affected instance.

[`arelle/plugin/validate/EDINET/rules/upload.py:1241`](https://github.com/Arelle/Arelle/blob/master/arelle/plugin/validate/EDINET/rules/upload.py#L1241):

```python
elif (parent := fact.getparent()) is not None:
    if fact == list(parent)[0] and parent.text:
```

The rule needs the parent's first **XML** child, but `ModelInlineFact` overrides `__iter__` ([`ModelInstanceObject.py:875`](https://github.com/Arelle/Arelle/blob/master/arelle/ModelInstanceObject.py#L875)) to yield the fact's **tuple** children instead:

```python
def __iter__(self) -> Iterator[ModelObject]:
    if self.localName == "fraction":
        ...
    for tupleFact in self.modelTupleFacts:
        yield tupleFact
```

So whenever a `nonFraction` fact's parent is itself an inline fact, `list(parent)` does not return the XML children. With no tuple children it is empty and the subscript raises.

The triggering shape is an `ix:nonFraction` nested directly in an `ix:nonNumeric` text block — ordinary EDINET markup for a count embedded in a note:

```html
<ix:nonNumeric name="jpcrp_cor:NumberOfUnconsolidatedSubsidiariesAnd...TextBlock"
               contextRef="CurrentYearDuration" escape="true">(1）持分法適用の関連会社数　<ix:nonFraction
    name="jpcrp_cor:NumberOfAssociatesAccountedForUsingEquityMethod"
    contextRef="CurrentYearInstant" unitRef="pure" xsi:nil="true" />－社</ix:nonNumeric>
```

The exception escapes the rule generator, so it aborts **every remaining EDINET `XBRL_FINALLY` rule** for that instance, not just EC8023W. Arelle logs it and exits 0, which makes the lost coverage easy to miss. On the reproducer filing below that is 26 suppressed messages, including EC2002W entirely:

| | codes reported | messages |
|---|---|---|
| before | 10 | 151 |
| after | 11 (`+EDINET.EC2002W`) | 177 |

Seen in 49 of 1,749 EDINET filing packages (≈2.8%) from a single June 2026 week. A wider sweep of the same harness's stored results — 5,525 validations spanning the June 2025 and June 2026 annual-report peaks — turns up 102 filings failing with an identical signature (`IndexError: list index out of range` on `instance: _IXDS`), 29 in the 2025 cohort and 73 in 2026.

#### Description of change

Use `iterchildren()`, which is the XML-level iterator and is not overridden:

```python
if fact == next(parent.iterchildren(), None) and parent.text:
```

This also fixes a latent false positive. Where the parent *is* an inline fact with tuple children, the old expression compared the fact against a tuple fact, so the `parent.text` check was skipped and EC8023W was reported even when the required `△` was present. A plain bounds check (`if parentChildren and fact == parentChildren[0]`) stops the crash but keeps that false positive.

Adds a second variation to the existing `EC8023W` testcase — `invalid02`, derived from `valid05.zip` by relocating two negative facts to be the first child of a `nonNumeric` text block, one preceded by `△` in the text block's text (no warning expected) and one not (warning expected). The fixture is built so it distinguishes the fix from a plain bounds check:

| | result |
|---|---|
| master | `IndexError`, rule aborted |
| bounds check only | 2 × EC8023W (one spurious, for the `△` case) |
| this PR | 1 × EC8023W |

`tests/resources/conformance_suites_timing/edinet.json` is left alone, since the timing file is refreshed in its own commits.

#### Steps to Test

Run the EDINET conformance suite:

```
python -m pytest tests/integration_tests/validation/test_conformance_suites.py \
    --name edinet --shard 0,1,2,3
```

`258 passed`. Reverting the one-line change in `upload.py` while keeping the new testcase gives `1 failed, 257 passed`, the single failure being `EC8023W/index.xml:invalid02`.

To see the original crash directly, EDINET filing S100YGNY (琉球セメント株式会社, EDINET code E01191, submitted 2026-06-22, period end 2026-03-31, 406 KB) reproduces it offline against a primed cache:

```
arelleCmdLine.py -f S100YGNY.zip \
    --plugins 'validate/EDINET|inlineXbrlDocumentSet' \
    --disclosureSystem EDINET-2026 -v --internetRecheck never
```

```
[exception:IndexError] Instance validation exception: list index out of range, instance: _IXDS
  ...
  File "arelle/plugin/validate/EDINET/rules/upload.py", line 1241, in rule_EC8023W
    if fact == list(parent)[0] and parent.text:
               ~~~~~~~~~~~~^^^
IndexError: list index out of range
```

I can attach the package if that is useful.

**review**:
@Arelle/arelle

---
🤖 Generated with [Claude Code](https://claude.com/claude-code)


arelle/plugin/validate/EDINET/rules/upload.py
```diff
@@ -1237,8 +1237,10 @@ def rule_EC8023W(
                 if strippedText:
                     precedingChar = strippedText[-1]
         # If nothing found, check the parent element if this is the first child.
+        # Note: iterchildren() is used instead of indexing the parent directly because
+        # ModelInlineFact overrides __iter__ to iterate tuple facts, not XML children.
         elif (parent := fact.getparent()) is not None:
-            if fact == list(parent)[0] and parent.text:
+            if fact == next(parent.iterchildren(), None) and parent.text:
```
austinmatherne-wk, line 1243: This is a no-op, isn't it? This `elif` branch is only reached if `precedingSibling`/`fact.getprevious()` is `None`. The fact has to be first child for the code to reach this line?

```python
elif (parent := fact.getparent()) is not None and parent.text:
    strippedText = parent.text.strip()
    if strippedText:
        precedingChar = strippedText[-1]
```

PR #2605, opened 2026-09-20
Load and validate perf improvements 

austinmatherne-wk:
#### Reason for change

Miscellaneous performance improvements from profiling. Reduces average load+validation (not including EDGAR/ESEF plugins) time by 5-7%.

#### Description of change

* Cache built in attribute schema types
* Cache decimal type checks within fact validation
* Collapse schema whitespace in one pass
* Avoid full QName parsing for plain HTML attribute names

#### Steps to Test

* CI 

**review**:
@Arelle/arelle


volodymyrtur-wk (dismissed):

arelle/ValidateXbrl.py
```diff
@@ -745,6 +746,10 @@ def checkDuplicateFacts(self, facts: list[ModelFact], duplicateType: ValidateDup
             )
 
     def checkFacts(self, facts: list[ModelFact], inTuple: dict[Any, Any] | None = None) -> None:  # do in document order
+        @cache
```
brettkail-wk: I assume this cache is shared across `ModelXbrl`.  Will the `@cache` grow without bound if repeated new instances of `ModelXbrl` are used?  Will the `@cache` cause memory leaks by pinning the `ModelConcept` (and all referenced objects) in memory?  Should this property be cached on the `ModelConcept` itself?

austinmatherne-wk: No, the cache was local to the surrounding `checkFacts` function and only used within the facts loop below. It was collected when `checkFacts` exited. I went ahead and implemented a cache on `instanceOfType` instead

aaroncameron-wk (approved):

arelle/ValidateXbrl.py
brettkail-wk: Oh interesting, I hadn't imagined Python would recreate the cache on each call.  In that case, the change seemed fine as-is.

I guess a dict cache is slightly slower than a bool cache for this particular use-case, but perhaps the dict will speed up other call-sites.  I had imagined you would add `isDtrNoDecimalsItem` alongside `isEnumeration`/`isEnumeration2Item`; I guess those caches are somewhat superseded?

PR #2606, opened 2026-09-20
Reduce base memory usage

austinmatherne-wk:
#### Reason for change

Loading a Solvency II entry point peaked at 1.95 GB before validation and 2.9 GB after.

* pyparsing generates a display name for every XPath grammar level, and because each operator level references the level below twice those names double per level to about 1 GB in total.
* The rest is duplicate strings. lxml returns a new string for every tag, attribute name, and attribute value, and validation kept every copy.

| Filing | Run | Memory before | Memory after | Time before | Time after |
| --- | --- | --- | --- | --- | --- |
| Solvency II | load | 1952 MB | 920 MB | 18.8 s | 15.1 s |
| Solvency II | load and validate | 2935 MB | 1637 MB | 103 s | 103 s |
| Tesco 2026 ESEF | load | 460 MB | 421 MB | 7.4 s | 6.6 s |
| Tesco 2026 ESEF | load and validate | 1839 MB | 794 MB | 42.2 s | 38.8 s |
| Workiva 2025 10-K | load | 187 MB | 178 MB | 2.6 s | 2.7 s |
| Workiva 2025 10-K | load and validate | 285 MB | 243 MB | 7.7 s | 7.3 s |

The 10-K only used the EDGAR transform plugin. No ESEF or EDGAR validation plugins were enabled.

#### Description of change

* Name each XPath grammar level so pyparsing stops generating exponential names.
* Intern tag parts, attribute names and values, element text, and QName parts so each distinct string is stored once.
* Keep one AnyURI object per distinct value on the model.

#### Steps to Test

* CI

**review**:
@Arelle/arelle


volodymyrtur-wk (approved):

arelle/ModelObject.py
```diff
@@ -173,12 +175,9 @@ def slottedAttributesNames(self) -> set[Any]:
...
 
     def setNamespaceLocalName(self) -> None:
-        tag = self.tag
-        ns, sep, self._localName = tag.rpartition("}")
-        if sep:
-            self._namespaceURI: str | None = ns[1:]
-        else:
-            self._namespaceURI = None
+        ns, sep, localName = self.tag.rpartition("}")
+        self._localName = sys.intern(localName)
```
brettkail-wk, line 179: Out of curiosity, is `sys.intern` preferred over a more localized cache (e.g., on `ModelXbrl`)?  I might have guessed that storing all strings in a global intern dict would be less preferable.

austinmatherne-wk: There's sufficient overlap in strings between models that a global cache does provide meaningfully lower memory usage when multiple models are opened at the same time (such as in the GUI). Both when they only share base XBRL schemas, but in particular when they share jurisdiction taxonomies.

aaroncameron-wk (approved):

